PR #13, opened 2025-11-24
Blank session cleanup

alexispurslane:
## Summary
Automatically clean up blank sessions when creating new ones to prevent accumulation of empty sessions in the UI.

## Details
- When a new session is created, the system now scans for and removes previously identified ("known") blank sessions
- "Known" blank sessions are those where message content has already been loaded and confirmed empty (typically sessions created or visited during the current instance)
- This approach provides a good balance: it handles most real-world cases of keeping clutter low efficiently while avoiding requiring a full history scan or doing huge destructive edits on the entire history of someone's session

## Benefits
- Keeps the session list clean by removing empty sessions users don't need
- Fast operation that doesn't impact performance
- Non-destructive: only removes sessions we've verified are blank through normal usage
- Avoids complex history manipulation while still solving the common case


shantur:
That's amazing feature. Thanks a lot for building it.
Would you be able to rebase your changes on the current dev?

PR #39, opened 2025-12-08
feat(ui/qol): add copy button to message items

crouch-era:
Add a Copy button that allows users to copy raw message contents (text and reasoning parts) to clipboard. The button appears on all messages - alongside Revert/Fork for user messages, and standalone for assistant messages. Shows "Copied!" feedback for 2 seconds after clicking.

Tested on electron and tauri on dev builds, both work as expected. 

image

### DIFF
--- a/packages/ui/src/components/message-item.tsx
+++ b/packages/ui/src/components/message-item.tsx
@@ -16,7 +16,7 @@ interface MessageItemProps {
   showAgentMeta?: boolean
   onContentRendered?: () => void
  }
- 
+
  export default function MessageItem(props: MessageItemProps) {
   const [copied, setCopied] = createSignal(false)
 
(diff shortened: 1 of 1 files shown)


shantur:
Thanks @crouch-era 
We can add copy buttons to thinking and text blocks
Thinking blocks are plain text so will copy plain text
Text blocks will copy the raw text in this case.  I don't think there is any benefit copying the rendered markdown.

shantur:
Hey @crouch-era ,
Yeah that should be ok, don't think anyone would be copying thinking blocks.
Thanks, I will review and merge

shantur:
Hi @crouch-era ,

I am happy with the changes, but would like to make the merge a fast-forward.
I will rebase your commits if you don't mind. I am happy if you would like to do it in your branch

PR #69, opened 2026-01-15
Fix: Question Tool UI Interaction Issues

bizzkoot:
# Fix: Question Tool UI Interaction Issues

## Summary

Resolves critical UI interaction issues with the question tool that prevented users from selecting options and typing custom answers.

## Problem

When agents used the `question` tool, the UI rendered but users could not interact with it:
1. ❌ Radio buttons and checkboxes were not selectable
2. ❌ Custom answer input lost focus after each character typed
3. ❌ Submit/Dismiss buttons were disabled

### 1. State Lookup Key Mismatch
In `instance-store.ts`, questions were stored with a fallback key (`partId ?? request.id`), but `getQuestionState()` only looked up by `partId`, causing lookups to always return `null`.

### 2. Non-Reactive Active Variable  
In `tool-call.tsx`, the `active` variable was computed once as a closure variable instead of a reactive accessor, preventing UI updates when the active state changed.

### 3. Focus Loss During Re-renders
The `<For>` loop re-rendered on state updates, causing the custom input to lose focus after each keystroke.

### `packages/ui/src/stores/message-v2/instance-store.ts`
- Enhanced `getQuestionState()` to perform fallback search by `tool.callID` when direct lookup fails
- Matches question's `tool.callID` against the tool part's callID to resolve the correct question entry

### `packages/ui/src/components/tool-call.tsx`
- Converted `active` variable to reactive accessor function `() => Boolean(request && isQuestionActive())`
- Updated all 13 usages of `active` to `active()` throughout `renderQuestionBlock()`
- Added `ref` callback to custom answer input that restores focus after component re-renders
- Added `data-question-index` attribute to track which input should maintain focus

## Testing

✅ Manually tested with `npm run dev`:
- Radio button options are now selectable
- Checkbox options work correctly for multiple-choice questions
- Custom answer input maintains focus while typing
- Submit/Dismiss buttons respond to clicks
- Keyboard shortcuts (Enter/Escape) function properly

## Related Issues

This fixes the question tool integration issues when consolidating question handling into the unified permission notification system.

### BEFORE
Question_tool_before

### AFTER

[Uploading Question_tool_after.mp4…
](https://github.com/user-attachments/assets/643edaec-e131-4c79-adb1-84d3f8a7376a)

## Checklist

- [x] Code compiles without errors
- [x] Manually tested all interaction scenarios
- [x] No breaking changes to existing functionality
- [x] Commit messages follow conventional commits format

### DIFF
--- a/packages/ui/src/components/tool-call.tsx
+++ b/packages/ui/src/components/tool-call.tsx
@@ -1174,6 +1174,13 @@ export default function ToolCall(props: ToolCallProps) {
                       <Show when={active()}>
                         <div class="mt-2 flex items-center gap-2">
                           <input
+                            ref={(el) => {
+                              const activeEl = document.activeElement
+                              if (activeEl?.getAttribute?.('data-question-index') === String(i())) {
+                                queueMicrotask(() => el?.focus())
+                              }
+                            }}
+                            data-question-index={i()}
                             class="flex-1 rounded-md border border-base/50 bg-surface px-2 py-1 text-sm"
                             type="text"
                             placeholder="Type your own answer"


bizzkoot:
To give clearer contexts,

I'm using macOS (26.1)
Opencode version 1.1.23 (installed using brew)

PR #103, opened 2026-01-28
feat(ui): implement smart heuristic session naming

bizzkoot:
# feat(ui): Smart heuristic session naming

## Summary
Replaces the default timestamp-based session naming (e.g., "New session - 2024-03-20...") with a smart, client-side heuristic that extracts semantic meaning from the user's first prompt.

## Problem
- The default "New session - [Timestamp]" name provides no context when browsing history.
- Using an LLM/Agent to generate titles (the "Title Agent") consumes premium tokens for every new chat, which is costly and adds latency.

## Solution
This PR implements a **Zero-Cost, Frontend-Only Heuristic** that runs locally in the browser when the first message is sent.

### Logic (MVP)
The `generateHeuristicTitle` function applies the following logic:
1.  **File Context Priority:** If the user mentions a file (e.g., `@src/auth.ts` or `main.rs`), the session is named `Context: auth.ts`.
2.  **Noise Filtering:** Strips common conversational stop words (*hi, please, can, you, help, ...*).
3.  **Privacy Guard:** Ignores "high entropy" strings (words > 25 chars) to prevent accidentally naming a session after a pasted API key.
4.  **Smart Truncation:**
    - Fills a "budget" of ~30 characters.
    - Capitalizes the first significant word.
    - Appends `...` only if meaningful keywords were truncated.

## Examples

| User Input | Generated Title |
| :--- | :--- |
| "Can you check `@src/auth.ts`?" | **Context: auth.ts** |
| "Fix the login bug please" | **Fix Login Bug** |
| "I'm getting a 500 error on the gateway" | **500 Error Gateway** |
| "My key is sk-ant-12345..." | **Key Is** (Long string ignored) |
| "for the question wizard, we implement a simple markdown preview..." | **Question Wizard Implement...** |

## Implementation Details
- Modified `packages/ui/src/stores/session-actions.ts`.
- Added `generateHeuristicTitle` utility.
- Updated `sendMessage` to check if `session.messageIds.length === 0` and trigger a non-blocking `renameSession` call.

## Cost Impact
- **Zero tokens.** No LLM calls are made for naming.

### DIFF
--- a/packages/ui/src/stores/session-actions.ts
+++ b/packages/ui/src/stores/session-actions.ts
@@ -2,7 +2,7 @@ import { resolvePastedPlaceholders } from "../lib/prompt-placeholders"
 import { instances } from "./instances"
 
 import { addRecentModelPreference, getModelThinkingSelection, setAgentModelPreference } from "./preferences"
-import { providers, sessions, withSession } from "./session-state"
+import { providers, sessions, setSessions, withSession } from "./session-state"
 import { getDefaultModel, isModelValid } from "./session-models"
 import { updateSessionInfo } from "./message-v2/session-info"
 import { messageStoreBus } from "./message-v2/bus"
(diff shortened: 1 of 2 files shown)


shantur:
I wonder even if we do this Opencode would still try to name the session.
How does Opencode behave with this?

shantur:
Sorry, I am failing to understand the need for this.

For me OpenCode always names the session using the small_model without any issues, CodeNomad updates the name when OC names the session. 

Why do we need to implement this?
Is there any case when OpenCode doesn't name the session?

PR #121, opened 2026-02-06
feat(ui): add PWA support with vite-plugin-pwa

jderehag:
- Add vite.config.pwa.ts extending the base config with VitePWA plugin
- Generate PWA icons at build time from source logo via @vite-pwa/assets-generator
- Add web app manifest with name, theme color, display overrides
- Add Workbox runtime caching: NetworkFirst for API, CacheFirst for assets
- Set navigateFallback to null to preserve server-side auth redirects
- Server build uses build:pwa for PWA-enabled output; Electron/Tauri use the base build without PWA


shantur:
Thanks for PR, I like it

shantur:
> My second PR will be a voice-agent on top of CodeNomad (just a microphone button or something, connected to elevenlabs voice agent, eventually we could add locally hosted STT models as well).
> That way we would get something like happy/hapivoice mode app, so STT<->TTS specifically for codenomad, but without the need for an external app (and possible relays).

I plan to integrate it with something similar to https://github.com/shantur/jarvis-mcp 
It is purely Browser based and doesn't depend on anything else, it works well.

Please do discuss it with me technical implementation if you plan to do it, we may come up with a good solution.
Open another issue for Voice input support that we can use to discuss.

Thanks for this PR, its merged now.

shantur:
Hey @jderehag,

I also added the https support in the server so you won't need to have a proxy to frontend with https

PR #210, opened 2026-03-08
fix(tauri): restore Windows desktop startup

pascalandr:
## Summary
- fix Windows packaged startup by resolving the bundled server entry from `resources/server/dist/*` next to the desktop executable
- strip Windows extended-length path prefixes (`\?\`) from canonicalized paths before spawning Node
- add the native `@tauri-apps/cli-win32-x64-msvc` dependency so `tauri build` works reliably on Windows

## Verification
- built the desktop app on Windows with `npm run build:tauri`
- launched the generated executable and confirmed the bundled CLI resolves successfully
- verified the app reaches `cli ready on http://127.0.0.1:...` instead of hanging on the splash screen

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3305,6 +3305,22 @@
         "node": ">= 10"
       }
     },
+    "node_modules/@tauri-apps/cli-win32-x64-msvc": {
+      "version": "2.9.4",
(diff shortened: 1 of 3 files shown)


shantur:
@pascalandr - Thanks for the PR, I will review this.

PR #214, opened 2026-03-14
fix(tauri): isolate desktop auth cookies per app

pascalandr:
## Summary
- Generate a unique auth cookie name per Tauri launch instead of reusing a shared host-scoped cookie.
- Thread the cookie name through the desktop bootstrap path and server auth manager.
- Fix multi-instance stalls when separate Tauri windows/apps share the same `127.0.0.1` auth state.

### DIFF
--- a/packages/electron-app/electron/main/main.ts
+++ b/packages/electron-app/electron/main/main.ts
@@ -327,7 +327,6 @@ function finalizeCliSwap(url: string) {
   mainWindow.loadURL(url).catch((error) => console.error("[cli] failed to load CLI view:", error))
 }
 
-const SESSION_COOKIE_NAME = "codenomad_session"
 let bootstrapExchangeInFlight = false
 
 function extractCookieValue(setCookieHeader: string | string[] | undefined, name: string): string | null {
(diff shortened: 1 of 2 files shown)


shantur:
@CodeNomadBot

- Validate the issue if its present
- Validate if the fix is correct
- Check if fix brings in any regressions for other platforms MacOS and Linux?
- Any other suggestions for the PR

Be the gatekeeper and check if this PR fixes what it says, and does that keeping the coding standards

pascalandr:
Thanks, this was a valid catch.

I pushed a focused follow-up fix on the PR branch (`8001333`).

What changed:
- removed the stray `emit_perf_startup(...)` call from `packages/tauri-app/src-tauri/src/cli_manager.rs`
- this fixes the missing-helper compile failure on this branch
- this also removes the unnecessary `json!({ "cookieName": auth_cookie_name })` path, so the `Arc<String>` serialization concern no longer applies here

The actual auth-cookie isolation logic is unchanged:
- the cookie name is still generated per desktop launch
- it is still passed to the CLI/server path via `auth_cookie_name.as_str()`

I rebuilt the updated PR branch with `npm run build:tauri`, and the build now passes.

@CodeNomadBot
Please perform the checks again with updated information

PR #215, opened 2026-03-14
perf(ui): lazy-load markdown and defer diff rendering

pascalandr:
## Summary
- lazy-load the markdown and diff render paths so they stop inflating initial UI startup work
- move shared text rendering helpers out of the markdown path and keep diff rendering on the deferred path
- defer the Monaco secondary viewers so the markdown and diff path no longer keeps that work in the main bundle

## Follow-ups
- related fork follow-up: Pagecran/CodeNomad#1
- that follow-up is now independent on dev and only keeps the remaining right panel, picker, and tool-call secondary chunking work

## Testing


pascalandr:
This PR reduces startup and initial-render work in the UI by moving markdown rendering and diff/highlighting code off the initial bundle path.

Before this change, the app paid the markdown + diff viewer cost eagerly even when the user had not opened content that needed those renderers yet. The result was a heavier initial UI load and slower perceived startup. This PR lazy-loads those paths and slims the diff highlighter dependency so we only pay that cost when the user actually opens markdown or diff-heavy content.

shantur:
@CodeNomadBot

- Perform PR checks again

pascalandr:
I reshuffled the split a bit.

I moved the Monaco-specific secondary viewer chunking into this base branch, since it directly completes the lazy markdown and diff story here. The remaining right panel, picker, and tool-call chunking work stays in Pagecran/CodeNomad#1, which now targets dev directly instead of stacking on top of this branch.

shantur:
@CodeNomadBot

Can you list the languages supported pre and post PR changes

shantur:
@CodeNomadBot

Can you list the languages supported pre and post PR changes

PR #222, opened 2026-03-16
feat(messaging): implement web worker for message processing and optimize list rendering

pixellos:
Introduces a web worker to offload message processing and optimizes virtual list rendering to improve UI responsiveness and performance for large message histories.

### DIFF
--- a/packages/ui/src/components/message-item.tsx
+++ b/packages/ui/src/components/message-item.tsx
@@ -383,6 +383,7 @@ export default function MessageItem(props: MessageItemProps) {
   return (
     <div
       class={containerClass()}
+      style={{ "content-visibility": "auto" }}
       data-view="message-item"
       data-instance-id={props.instanceId}
       data-session-id={props.sessionId}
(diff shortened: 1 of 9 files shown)


shantur:
@pixellos  - Do checkout https://github.com/NeuralNomadsAI/CodeNomad/pull/230

shantur:
> I believe the usage of `virtua` library instead of custom virtualization can be more optimized, do You have any prefferences to have custom one? The trace from switching between rather small size conversation, but on custom list was laggy af [Trace-20260319T220919.json.gz](https://github.com/user-attachments/files/26126728/Trace-20260319T220919.json.gz)

Totally depends how the virtua library works. I haven't used it myself but we want

1. Scroll bars to be stable height once am item is renderened 
2. Auto scroll and bottom follow supports
3. Fast rendering

I know the bug that caused custom list being laggy as in a case virtual item virtualisation was stopped.

Maybe open an issue / PR to discuss and we can see

PR #226, opened 2026-03-17
fix(tauri): restore desktop menu controls and fullscreen shortcut

pascalandr:
## Summary
- restore the missing desktop View and Window menu controls
- use native reload and window actions where supported instead of brittle webview-only behavior
- restore the working fullscreen keyboard shortcut while keeping the zoom menu labels aligned with the intended desktop behavior

## Testing
- cargo check --manifest-path packages/tauri-app/src-tauri/Cargo.toml


pascalandr:
Updated this to target dev and rebuilt the branch on top of it.

The diff is now limited to the desktop menu, fullscreen, and zoom work, without the unrelated auth, Electron, or server changes that were showing up before.

PR #238, opened 2026-03-19
perf(ui): defer locale and overlay bundles

pascalandr:
## Summary
- defer locale and overlay loading work away from the first critical render path
- seed locale state from the bootstrap preload so the first render can use the preloaded language immediately
- keep bootstrap cache and locale fallback behavior consistent on subsequent launches

## Testing


pascalandr:
Done in b7ed232: removed the separate ui-bootstrap cache and reused the existing storage cache path.

PR #239, opened 2026-03-19
perf(ui): split right panel and secondary viewer chunks

pascalandr:
## Summary
- split the right panel, picker, and tool call secondary viewers into smaller deferred chunks
- release hidden right-panel file buffers and stop tracking static tool-call scrollers when they are not needed
- keep this branch focused on the remaining secondary viewer chunking work now that the Monaco-specific chunking moved into PR 215

## Testing


pascalandr:
I reshuffled this branch a bit.

The Monaco-specific chunking now lives in PR 215, since it directly completes the lazy markdown and diff story there. This PR now only keeps the remaining right panel, picker, and tool-call secondary chunking work on top of dev.

pascalandr:
conflicts resolved

PR #240, opened 2026-03-20
fix(tauri): force Windows process tree shutdown

pascalandr:
## Summary
- force the Windows CLI process tree shutdown path during normal app close
- avoid leaving child server processes alive when the direct wrapper process exits first
- keep the change limited to the Windows shutdown path in cli_manager

## Testing
- cargo check --manifest-path packages/tauri-app/src-tauri/Cargo.toml


pascalandr:
this is broken again, the final fix wasn't robust enough, it doesn't close processes anymore at all ... so this is the third time I have to work on this....

PR #247, opened 2026-03-25
PR: Extend delete overlays and clean companion parts during tool deletion

### LINKED ISSUE #246: Bug / UX Issue: Incomplete deletion when removing tool calls from the timeline
# Bug / UX Issue: Incomplete deletion when removing tool calls from the timeline

## Problem
When users select specific tool calls from the timeline for bulk deletion (instead of selecting an entire message), the deletion process only removes the tool call parts. The associated "reasoning" block (thinking tokens) and "text" parts generated based on that tool output are left behind. 

Additionally, the red pre-deletion overlay only highlights the tool call cards. It does not visually indicate that any other associated components (like the message header, reasoning, or text) are part of the context being managed.

## Impact
- **Orphaned Content:** Deleting a tool call leaves behind an incomplete message structure—an assistant response with "thinking" and "text" but no underlying tool call output.
- **Misleading Visuals:** The UI fails to clearly communicate what parts of the message belong to the selected tool calls, making the delete action feel unpredictable and preventing users from cleaning up their conversation history accurately.

## Expected Behavior
1. **Aggressive Companion-Part Deletion:** When tool parts of a message are selected for deletion, the deletion action should also automatically remove the corresponding `step-finish`, `reasoning`, and `text` parts associated with that message. The chat history should be cleanly purged of all content tied to those tools.
2. **Extended Delete Overlays:** When a user selects tool calls for deletion, the red destructive overlay should extend to the message headers, reasoning cards (thinking tokens), and text parts to accurately reflect the scope of what will be removed.

VooDisss:
# PR: Extend delete overlays and clean companion parts during tool deletion
Fixes: #246 
## Purpose
This PR addresses incomplete deletions when users bulk-delete specific tool calls from the timeline. Previously, deleting a tool part left behind orphaned reasoning logs (thinking tokens) and generated text parts, and the destructive hover overlay only highlighted the tool cards themselves.

This update ensures that deleting tool parts cleans up the entire associated response context (`step-finish`, `reasoning`, and `text` parts) from that message. It also extends the red delete overlay to message headers, reasoning cards, and text parts to accurately reflect the scope of deletion to the user.

### `packages/ui/src/components/message-block.tsx`
- Threaded the `selectedToolPartKeys` prop through `MessageContentItem`, `MessageBlock`, and `ReasoningCard`.
- Updated `ReasoningCard`'s `isSelectedForDeletion` logic to check both `selectedMessageIds` (whole message) and `selectedToolPartKeys` (tool-part selection via a `messageId:` prefix match).
- Moved the block-level selection highlight out of `isDeleteMessageHovered` to be handled at the part level via `data-delete-part-hover` attributes avoiding double overlays.
- Added `data-delete-part-hover` to the `ReasoningCard`'s root `div`.

### `packages/ui/src/components/message-item.tsx`
- Added the `selectedToolPartKeys` prop to `MessageItemProps`.
- Expanded `isSelectedForDeletion` using the same prefix-match pattern to detect tool-part selections.
- The `<header>` element now includes a `delete-hover-scope` class and `data-delete-part-hover` attribute to trigger the overlay at the header level.
- Each `.message-part-shell` rendering text content now includes `data-part-type` and `data-delete-part-hover` for per-part overlay highlighting.

### `packages/ui/src/components/message-section.tsx`
- Updated the `deleteSelectedMessages` bulk action. It now additionally collects and deletes `step-finish`, `reasoning`, and `text` parts when a message has its tool parts selected for deletion. This ensures no orphaned assistant content remains.
- Simplified the deletion guard condition from `selectedToolPartIds.size === 0 || stepFinishPartIds.length === 0` down to `selectedToolPartIds.size === 0`, ensuring reasoning/text parts evaluate correctly even if `step-finish` parts are absent.

### `packages/ui/src/styles/messaging/delete-overlays.css`
- Widened the part-level overlay inset from `-2px` to `-4px` for visual consistency with the message-level overlay.
- Added a new rule for `.message-reasoning-card[data-delete-part-hover]` using `inset: 0` for flush overlay rendering within reasoning card boundaries.


pascalandr:
Hi @VooDisss, is this PR still intended to address #246, or should we re-evaluate the current timeline deletion behavior from scratch?

VooDisss:
@pascalandr hi, yes

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr (changes requested):
Gatekeeper review of latest head `e99103f7` after the conflict resolution and step-scoped cleanup changes.

The step grouping itself is substantially safer than the previous implementation: unrelated steps are preserved, shared reasoning for partially selected parallel tools is retained, assistant text is intentionally preserved, and the overlays are derived from the same companion set as deletion. The focused helper tests, UI typecheck, and production build pass.

There is still one merge blocker in the destructive workflow.

## Blocking finding

**P1: a companion deletion failure after tool deletion leaves an unrecoverable partial state and can turn Retry into whole-message deletion.**

`packages/ui/src/components/message-section.tsx:666-677` deletes all selected tools first, then sends independent requests for their `reasoning` and `step-finish` companions. Every successful request immediately mutates the local store at `packages/ui/src/stores/session-actions.ts:434-446`.

If a companion request fails:

1. the tool has already been deleted;
2. the catch at `message-section.tsx:679-685` leaves delete mode and its timeline selection active;
3. companions are recomputed only from tool parts that still exist, so the orphaned reasoning/finish parts are no longer recoverable through this action;
4. timeline segment IDs are positional (`packages/ui/src/components/message-timeline.tsx:199-225`);
5. after the tool disappears, its old ID such as `messageId:0` can be reassigned to the preserved assistant text;
6. stale-selection pruning checks only whether that ID still exists (`message-section.tsx:1171-1177`), not whether it still identifies the same part/type;
7. the retained selection is then interpreted as a non-tool segment and promotes the message into `selectedForDeletion`.

A user pressing Delete again after the error can therefore delete the whole message, including the assistant text this PR explicitly preserves.

This is not fixed merely by catching the error. At minimum, make partial failure recoverable: process companions before their tools, keep selection keyed by stable message/part IDs rather than positional segment IDs during the operation, and clear/reconcile stale selection on failure. A backend batch/transaction would be preferable if available.

## Missing regression coverage

`packages/ui/src/components/tool-deletion-companions.test.ts` validates companion collection, but not the destructive orchestration. Add coverage where a tool deletion succeeds and a companion deletion fails, then assert that:

- preserved text is never promoted to whole-message deletion;
- the remaining operation can be retried safely;
- overlays and token estimates reconcile with the surviving parts.

## Validation

- Companion helper tests: passed, 4/4.
- Conflict-specific diff check: passed.
- PR is conflict-free and GitHub reports it mergeable.
- The GitHub build matrix is still partially pending at review time.

## Verdict

Request changes. The normal path is improved and the feature remains worthwhile, but a single failed companion request currently creates a destructive retry path that can remove preserved content or the entire message.

pascalandr (dismissed):
Gatekeeper re-review of head `042731ba`.

The partial-failure blocker from my previous review is resolved.

- The operation snapshots full-message, companion, and tool targets before mutation.
- Positional timeline selection is cleared before the first request, so a rebuilt segment ID cannot redirect a retry to preserved text or whole-message deletion.
- `reasoning` and `step-finish` companions are deleted before their tools. A companion failure therefore leaves the selected tools intact.
- Any failure requires a fresh selection, producing a new plan from stable message/part IDs.
- Full-message deletions remain newest-first, and part targets belonging to fully selected messages are excluded.

Regression coverage now verifies:

- selection is cleared before mutation;
- companions execute before tools;
- a companion failure prevents tool deletion;
- a message deletion failure does not continue into part deletion.

Validation:

- focused tests: passed, 7/7;
- UI typecheck: passed;
- UI production build: passed;
- `git diff --check`: passed;
- independent re-review found no remaining blocker;
- GitHub reports the PR conflict-free and mergeable; the refreshed CI run is still pending.

Approved from the gatekeeper perspective, subject to the required CI checks completing successfully.

pascalandr (approved):
Thanks @VooDisss, sorry for my mixup while reviewing, you PR as you said make perfect sens and is super usefull !!!

PR #257, opened 2026-03-29
fix(tauri): sync native app version with package releases

pascalandr:
## Summary
- sync the Tauri native version metadata from `packages/tauri-app/package.json` so release builds pick up workspace version bumps like `0.13.1`
- update the checked-in Tauri `Cargo.toml` and `tauri.conf.json` versions from `0.12.3` to `0.13.1`
- document the prebuild sync behavior in `BUILD.md`

## Testing
- `node packages/tauri-app/scripts/sync-tauri-version.js`

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -252,7 +252,7 @@ A: Approximately 100-150 MB (includes Electron runtime).
 A: Not required, but recommended for public distribution to avoid security warnings.
 
 **Q: How do I update the version?**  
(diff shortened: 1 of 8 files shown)


shantur:
Thanks @pascalandr 

So I can understand, can you please explain what's the issue that's being solved here.

shantur:
@CodeNomadBot 

Suggest a simple patch that does the following

- As part of bumpVersion script, it should update all npm package versions for all packages
- packages/tauri-app package.json and Tauri app versions to be updated as part the bumpVersion script updating all the versions for tauri-app

PR #258, opened 2026-03-29
fix(tauri): stop stale UI assets from shadowing desktop builds

pascalandr:
## Summary
- prefer the bundled desktop UI over the downloaded cache when both report the same version, so rebuilt installers do not keep serving stale frontend assets
- rebuild the server workspace during the Tauri prebuild step on every desktop package build, matching Electron's correctness boundary for fresh UI/server assets
- add a regression test covering the equal-version bundled-vs-downloaded UI selection path

## Why
- local desktop rebuilds should reflect the latest server and UI code without requiring users to manually clear cached assets
- packaged updates should keep favoring the freshly bundled frontend when the cached copy is not actually newer

## Testing
- node --import tsx --test packages/server/src/ui/__tests__/remote-ui.test.ts
- npm run build:tauri

### DIFF
--- a/packages/tauri-app/scripts/prebuild.js
+++ b/packages/tauri-app/scripts/prebuild.js
@@ -53,52 +53,10 @@ async function ensureMonacoAssets() {
   })
 }
 
-function latestModifiedTime(targetPath) {
-  if (!fs.existsSync(targetPath)) {
-    return 0
-  }
-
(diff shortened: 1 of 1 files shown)


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

Also check how this differs from packages/electron-app build process. Electron-app works correctly

PR #262, opened 2026-03-31
fix(server): show sane remote URLs for 0.0.0.0 binds

### LINKED ISSUE #261: Server startup logs should advertise sane LAN URLs when bound to 0.0.0.0
## Summary

When CodeNomad is started with `--host 0.0.0.0`, the server is reachable on multiple local interface addresses, but the startup log currently advertises only a single "Remote Connection URL". On Windows machines with WSL, Hyper-V, Docker, or other virtual adapters, the printed address often ends up being a virtual or less useful private IP instead of the address a user is most likely to open from another device on the LAN.

This makes remote access appear broken or confusing even when the server is actually reachable from a normal LAN address.

## Problem

Example behavior:

- Command:
  - `codenomad --host 0.0.0.0 ...`
- Startup output may show:
  - `Remote Connection URL : https://172.19.x.x:9898`

But the server may also be reachable at a more useful address such as:

- `https://192.168.1.128:9898`

The problem is not that the server is unavailable. The problem is that the startup output does not accurately communicate the reachable addresses a user is likely to need.

This is especially confusing on Windows hosts with multiple adapters, including:

- WSL / Hyper-V virtual Ethernet
- Docker virtual networks
- additional NAT / bridged interfaces

## Expected behavior

When the server binds to `0.0.0.0`:

1. The startup log should prefer a user-meaningful remote address for the primary "Remote Connection URL".
2. The startup log should also show additional reachable remote URLs instead of hiding them.
3. Clearly low-value addresses should not be prioritized for user-facing output.
4. The behavior should remain stable across machines with multiple private-network adapters.

## Preferred solution

A good solution would be:

- keep network-address discovery in the server/network-address resolver layer
- distinguish between:
  - all reachable addresses
  - addresses suitable for user-facing advertisement
- choose the primary remote URL from the advertisable set rather than blindly taking the first external address
- print additional reachable URLs in the startup log when multiple useful addresses are available
- avoid hard-coding one RFC1918 subnet family (for example `192.168.*`) as universally better than all others
- deprioritize or suppress clearly low-value addresses such as link-local `169.254.*` in terminal output

## Notes

This issue is about **address selection and presentation**, not about changing listener binding behavior, TLS mode, or authentication behavior.

## Acceptance criteria

- Starting CodeNomad with `--host 0.0.0.0` no longer advertises only a single potentially misleading private IP.
- The startup log includes additional useful reachable URLs when multiple interfaces are available.
- Link-local addresses are not chosen as the primary advertised remote URL.
- Address-ordering behavior is covered by targeted tests so interface enumeration does not silently regress the primary URL again.

VooDisss:
Closes #261

## Summary

- improve startup remote URL selection when the server binds to `0.0.0.0`
- print additional reachable remote URLs instead of advertising only the first external address
- add targeted tests for address ordering and advertisability behavior

## Problem

When CodeNomad was started with `--host 0.0.0.0`, the CLI chose the first external IPv4 address it discovered and displayed only that one as the remote URL.

On Windows machines with WSL, Hyper-V, Docker, or other virtual adapters, that often surfaced a virtual `172.x.x.x` address even though a more useful LAN address such as `192.168.x.x` was also reachable and usable from other devices.

That made remote access look broken or confusing even though the server itself was accessible.

## What changed

- reuse the resolved network-address list for both:
  - primary remote URL selection
  - startup logging of additional reachable URLs
- choose the primary remote URL from the **advertisable** external addresses instead of any external address
- print `Other Accessible URLs` when multiple useful remote URLs are available
- avoid hard-coding a preference like `192.168 > 10 > 172`
- suppress link-local `169.254.*` addresses from user-facing advertised URLs
- add tests covering:
  - stable ordering across RFC1918 address ranges
  - link-local addresses being non-advertisable
  - link-local-first discovery not stealing the primary LAN URL

## Why this approach

This keeps address derivation in the network-address resolver layer and limits `index.ts` to startup wiring and presentation.

It also fixes the misleading terminal output without redesigning binding behavior, TLS behavior, or the server API contract.

## Validation

- `npx tsx --test '.\\src\\server\\__tests__\\network-addresses.test.ts'`

## Notes

- this change is intentionally focused on selection and presentation of reachable addresses
- it does not attempt a broader virtual-adapter classification policy beyond suppressing clearly low-value link-local addresses in user-facing output

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -21,7 +21,7 @@ import { launchInBrowser } from "./launcher"
 import { resolveUi } from "./ui/remote-ui"
 import { AuthManager, BOOTSTRAP_TOKEN_STDOUT_PREFIX, DEFAULT_AUTH_COOKIE_NAME, DEFAULT_AUTH_USERNAME } from "./auth/manager"
 import { resolveHttpsOptions } from "./server/tls"
-import { isAdvertisableRemoteAddress, resolveNetworkAddresses } from "./server/network-addresses"
+import { resolveNetworkAddresses, resolveRemoteAddresses } from "./server/network-addresses"
 import { startDevReleaseMonitor } from "./releases/dev-release-monitor"
 import { SpeechService } from "./speech/service"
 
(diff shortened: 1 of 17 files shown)


shantur:
@CodeNomadBot

Can you do your thing again

PR #274, opened 2026-04-02
perf(ui): fix O(n²) reactive subscriptions in timeline effects (HUGE SPEED IMPROVEMENT)

pascalandr:
## Summary

- Wraps store-proxied array iteration in `untrack()` in two `createEffect` blocks and one `createMemo` in `message-section.tsx` to prevent SolidJS from creating O(n) per-element reactive subscriptions on every run
- Replaces `ids.includes()` with `Set.has()` for O(1) cleanup lookups in the part-count tracking effect

## Problem

Two `createEffect` blocks in `message-section.tsx` iterate the `messageIds()` store proxy array inside a tracked reactive context. This causes SolidJS to create **O(n) per-element subscriptions** on every run. When any element changes, all n subscriptions fire, re-running the entire effect — resulting in **O(n²) total work**.

Additionally, the cleanup loop in the part-count tracking effect uses `ids.includes(trackedId)` which is O(n) per tracked ID, compounding to O(n²).

For long-running sessions with large message history (e.g. 7569 messages), this caused **~4.8 seconds of input latency** when sending a new prompt.

## Fix

1. **Timeline sync effect (~line 738):** Wrap entire body in `untrack()`, replace `ids.slice()` with `[...ids]` to snapshot without proxy tracking
2. **Part-count tracking effect (~line 891):** Wrap iteration in `untrack()`, replace `ids.includes()` with `new Set(ids).has()` for O(1) lookups
3. **`lastAssistantIndex` memo:** Read message records via `untrack()` to avoid O(n) subscriptions on part-level updates

## Result

On a 7569-message session: prompt input latency reduced from **~4.8s to ~42ms** (114x improvement).

### DIFF
--- a/packages/ui/src/components/message-section.tsx
+++ b/packages/ui/src/components/message-section.tsx
@@ -129,6 +129,8 @@ export default function MessageSection(props: MessageSectionProps) {
     return map
   })
 
+  const lastAssistantMessageId = createMemo(() => store().getLastAssistantMessageId(props.sessionId))
+
   const lastCompactionIndex = createMemo(() => {
     // Depend on a single session revision signal (not every message/part read)
     // to keep reactive overhead small.
@@ -315,20 +317,9 @@ export default function MessageSection(props: MessageSectionProps) {
   }
 
   const lastAssistantIndex = createMemo(() => {
-    // Subscribe to messageIds (for length/order changes) and sessionRevision
-    // (for role changes after compaction or hydration), but read individual
-    // message records inside untrack() to avoid O(n) subscriptions that would
-    // fire on every part-level update in any message.
-    const ids = messageIds()
-    sessionRevision()
-    const resolvedStore = untrack(store)
-    for (let index = ids.length - 1; index >= 0; index--) {
-      const record = untrack(() => resolvedStore.getMessage(ids[index]))
-      if (record?.role === "assistant") {
-        return index
-      }
-    }
-    return -1
+    const messageId = lastAssistantMessageId()
+    if (!messageId) return -1
+    return messageIndexById().get(messageId) ?? -1
   })
  
   const [timelineSegments, setTimelineSegments] = createSignal<TimelineSegment[]>([])
--- a/packages/ui/src/components/message-section.tsx
+++ b/packages/ui/src/components/message-section.tsx
@@ -315,10 +315,15 @@ export default function MessageSection(props: MessageSectionProps) {
   }
 
   const lastAssistantIndex = createMemo(() => {
+    // Subscribe to messageIds (for length/order changes) and sessionRevision
+    // (for role changes after compaction or hydration), but read individual
+    // message records inside untrack() to avoid O(n) subscriptions that would
+    // fire on every part-level update in any message.
     const ids = messageIds()
-    const resolvedStore = store()
+    sessionRevision()
+    const resolvedStore = untrack(store)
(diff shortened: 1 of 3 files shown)


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

pascalandr:
I want to correct one detail:

**`applyPartDeltaV2` does NOT bump `sessionRevision` on streaming deltas.** The streaming hot path (`session-events.ts:177-185` → `bridge.ts:109-123`) explicitly calls `applyPartDeltaV2` with `bumpSessionRevision: false`. So the original PR was as described — `lastAssistantIndex` was only re-evaluated on actual structural mutations (message insert/remove/replace), not on every streamed token.

That said, the core suggestion was correct: coupling `lastAssistantIndex` to `sessionRevision` at all is needlessly coarse, since most mutations that bump the revision don't change which message is the last assistant one.

The new commit (`5c905f3`) addresses this fully by doing exactly what you suggested — **deriving the last assistant message ID incrementally in the store itself**:

- Added a `lastAssistantMessageIds` cache (`Map<sessionId, messageId>`) to the store state
- `recomputeLastAssistantMessageId()` is called on all 7 mutation paths that can change message ordering/roles: `addOrUpdateSession`, `hydrateMessages`, `upsertMessage`, `removeMessage`, `replaceMessageId`, `pruneMessagesAfterRevert`, `clearSession`
- `message-section.tsx` now reads this cached value via `getLastAssistantMessageId(sessionId)` — a simple O(1) Map lookup with no dependency on `sessionRevision` or the messages array
- The backward O(n) scan is completely eliminated from the render path

Typecheck and build both pass.

shantur:
@CodeNomadBot

Do your thing again

shantur:
@CodeNomadBot

Are there any unnecessary white space changes in the PR ?

pascalandr:
@shantur
thinking about it, commit 2 is correct and marginally better in reactive precision, but the complexity-to-benefit ratio is unfavorable for its stated purpose. The store cache makes more sense if other consumers will use getLastAssistantMessageId in the future. For this single memo, it's over-engineered.
In my oppinion we should revert commit 2, just drop sessionRevision() from commit 1's memo for nearly the same precision with zero store changes.

shantur:
@pascalandr - TBH, I haven't tried this PR and reviewed changes myself yet.
I would say, it depends on how frequently the lastAssistantMessageID is needed and TBH, thinking about it, there shouldn't be any calculations needed to find the lastAssistantMessageID but it should be saved in store when an assistant message is received.

What do you think

pascalandr:
@shantur That's exactly what commit 2 does. I was leaning toward reverting it since the complexity-to-benefit ratio felt off for a single consumer, but I agree your point of view.

shantur:
@CodeNomadBot

Can you validate my claim in last comments

shantur:
@CodeNomadBot

What patch would you suggest on the PR to move the lastAssistantMessageId in store and save if from messages are received and store mutations happen?

shantur:
@pascalandr - What do you think about the patch, if you agree, would you be able to apply it on your PR so I can review and merge it?

pascalandr:
@shantur that's already what commit 5c905f3 does. I pushed it earlier, so the PR match the patch shape described above.

shantur:
Thanks @pascalandr

PR #279, opened 2026-04-02
feat(sidecars): add proxied sidecar tabs

shantur:
## Summary
- add SideCar support across the server and UI, including proxied tabs, picker/settings flows, and websocket-aware proxying
- unify top-level tab handling so workspace instances and SideCars share the same tab model and navigation flows
- limit SideCars to port-based services only, removing server-managed process control from the final API and UI


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

do your thing again

shantur:
Thanks, this is a fair observation about the browser security model.

For this product, sidecars are an explicitly trusted integration surface, so same-origin embedding is intentional rather than an isolation boundary. In other words, they are closer to trusted app extensions than to untrusted third-party content.

The earlier proxy hardening was meant to prevent accidental forwarding of CodeNomad auth headers into sidecar backend requests, not to sandbox sidecars from the parent UI. With that trust model clarified, I also restored cookie behavior for sidecars in `9638db1` so normal cookie-backed embedded webapps continue to work.

So I agree with the technical description, but not with treating it as a blocker for this PR under the intended trusted-sidecar model.

PR #289, opened 2026-04-04
fix(desktop): support self-signed remote HTTPS windows

pascalandr:
## Summary
- route remote desktop windows through a local HTTPS proxy so WebView2 accepts secure cookies when the target server uses a self-signed certificate
- generate and trust a local proxy certificate on Windows, then terminate TLS locally while forwarding requests to the remote CodeNomad server
- install the rustls crypto provider explicitly so the HTTPS proxy starts reliably at runtime on Windows

## Testing
- cargo check
- npx @tauri-apps/cli build
- manually verified remote window connection works against a self-signed HTTPS CodeNomad server

### DIFF
--- a/packages/tauri-app/Cargo.lock
+++ b/packages/tauri-app/Cargo.lock
@@ -48,6 +48,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
+name = "arc-swap"
+version = "1.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "async-broadcast"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -212,6 +221,105 @@ name = "autocfg"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
+
+[[package]]
+name = "aws-lc-rs"
+version = "1.16.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a054912289d18629dc78375ba2c3726a3afe3ff71b4edba9dedfca0e3446d1fc"
+dependencies = [
+ "aws-lc-sys",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-lc-sys"
+version = "0.39.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "83a25cf98105baa966497416dbd42565ce3a8cf8dbfd59803ec9ad46f3126399"
+dependencies = [
+ "cc",
+ "cmake",
+ "dunce",
+ "fs_extra",
+]
+
+[[package]]
+name = "axum"
+version = "0.7.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
+dependencies = [
+ "async-trait",
+ "axum-core",
+ "bytes",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
+ "hyper",
+ "hyper-util",
+ "itoa",
+ "matchit",
+ "memchr",
+ "mime",
+ "percent-encoding",
+ "pin-project-lite",
+ "rustversion",
+ "serde",
+ "serde_json",
+ "serde_path_to_error",
+ "serde_urlencoded",
+ "sync_wrapper",
+ "tokio",
+ "tower",
+ "tower-layer",
+ "tower-service",
+ "tracing",
+]
+
+[[package]]
+name = "axum-core"
+version = "0.4.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09f2bd6146b97ae3359fa0cc6d6b376d9539582c7b4220f041a33ec24c226199"
+dependencies = [
+ "async-trait",
+ "bytes",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
+ "mime",
+ "pin-project-lite",
+ "rustversion",
+ "sync_wrapper",
+ "tower-layer",
+ "tower-service",
+ "tracing",
+]
+
+[[package]]
+name = "axum-server"
+version = "0.7.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c1ab4a3ec9ea8a657c72d99a03a824af695bd0fb5ec639ccbd9cd3543b41a5f9"
+dependencies = [
+ "arc-swap",
+ "bytes",
+ "fs-err",
+ "http",
+ "http-body",
+ "hyper",
+ "hyper-util",
+ "pin-project-lite",
+ "rustls",
+ "rustls-pemfile",
+ "rustls-pki-types",
+ "tokio",
+ "tokio-rustls",
+ "tower-service",
+]
 
 [[package]]
 name = "base64"
@@ -408,6 +516,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7a0dd1ca384932ff3641c8718a02769f1698e7563dc6974ffd03346116310423"
 dependencies = [
  "find-msvc-tools",
+ "jobserver",
+ "libc",
  "shlex",
 ]
 
@@ -445,6 +555,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
+
+[[package]]
 name = "chrono"
 version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


shantur:
@pascalandr - Thanks, that's a clever trick to work around Tauri self signed.

A few things, 
- can we restrict this server to only accept connections from the Tauri process? 
- Maybe not use a network port but use some other way?
- Maybe handle self signed in tauri with a custom protocol handler instead of starting a whole server?

shantur:
@CodeNomadBot

What are our options here, do you have some ideas

pascalandr:
Hardened in ba9cd53: loopback-only + ephemeral port, one-time random bootstrap token, then HttpOnly; Secure; SameSite=Strict local cookie, and 403 for unauthenticated requests. I also looked at a custom protocol, but it breaks the normal HTTPS/cookie/SSE behavior we need here, so I kept the proxy and tightened it instead.

shantur:
@pascalandr - 

Few things

1. The CodeNomad server already generates a self signed certificate when the server starts in ~/.config/codenomad so if needed that can be used instead of managing 2 certificates
2. If we are automatically trusting the locally generated certificate, can't we just get the remote server certificate and trust that rather than trusting local cert?

pascalandr:
1. Reusing the existing local CodeNomad cert is a good follow-up and would let us avoid managing a second local cert.
2. I don’t want to auto-trust the remote server cert in the OS store: that broadens trust to that remote host outside the app and can persist across server/cert changes. Trusting a loopback cert we generate and only use for the local proxy is narrower and easier to reason about.

shantur:
1. We are indirectly trusting the remote server anyway, so what's the harm directly accepting the certificate?
2. Also the local CodeNomad server can be modified to do this job instead of creating a separate server just for proxy, it already implements proxying in it for using sidecars.

pascalandr:
1. The difference is scope: trusting the remote cert in the OS store would trust that host outside CodeNomad too. Here we only trust a loopback cert that the app owns.
2. Agreed on reusing the existing local CodeNomad server/proxy path. I treated that as a follow-up refactor and kept this PR focused on getting self-signed remote windows working first.

shantur:
> 1. The difference is scope: trusting the remote cert in the OS store would trust that host outside CodeNomad too. Here we only trust a loopback cert that the app owns.

Is it possible to un-trust the remote cert ? We could make it work like
- Connect to an untrusted server
- User is asked for confirmation to allow trusting
- Trust certificate
- Connect to server
- When remote window is closed
- Untrust the certificate

pascalandr:
Possible in theory, but I think it’s the riskier path: we’d need to pin the exact remote cert/fingerprint, trust it in the OS store, then reliably remove exactly that trust entry on close without affecting other uses of the same cert/CA. That’s broader and more fragile than trusting a local loopback cert that CodeNomad fully owns.

shantur:
> other uses of the same cert/CA.

There aren't going to be other users for that cert, if the cert is already there we won't need to trust it ourself. Isn't it?

shantur:
@CodeNomadBot

Can you please verify my suggestion regarding trusting untrusting cert

- Does Tauri need trusted cert to work in all 3 platforms
- Is it possible to request Trusting a cert on all 3 platforms macos, linux and windows?
- Is it possible to untrust a cert on 3 platforms
- What is your take on the idea?

Keep your response brief but informative

shantur:
@CodeNomadBot

Will we have the same issues with Trusting CodeNomad owned certificate on these platforms?
Research if there is any library / Tauri helper for this that simplifies the process.

shantur:
@pascalandr - It seems this approach of installing a certificate and creating a proxy in CodeNomad for working around a self signed SSL might not work very well on all the platforms, as explained above.
I am leaning towards it being handled by OS either by
1. Trusting the remote certificate
2. Setting up a proxy outside CodeNomad and use that to connect. I can quickly whip up a nginx config to help

pascalandr:
@shantur I think that’s the real tradeoff: there may simply not be a good cross-platform in-app solution here.
This PR still has value as a pragmatic Windows fix now ; And we can also try to move toward an OS/external-proxy approach. I’d be interested in the nginx config to see if that broader direction is actually viable.

shantur:
Below config should work

```
# /etc/nginx/conf.d/reverse-proxy.conf

upstream remote-codenomad-server {
    server <remote-codenomad-server-ip>:9898;
    keepalive 16;
}

# Local HTTPS listener
server {
    listen 8443 ssl;
    server_name localhost;

    ssl_certificate     /etc/nginx/ssl/local.crt;
    ssl_certificate_key /etc/nginx/ssl/local.key;

    location / {
        proxy_pass https://remote-codenomad-server;

        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;

        proxy_ssl_server_name on;
        proxy_ssl_name <remote-codenomad-server-ip>;
        proxy_ssl_verify off;

        proxy_connect_timeout 60s;
        proxy_send_timeout 60s;
        proxy_read_timeout 60s;
    }
}
```

pascalandr:
Thanks, this helps. I agree this could be a cross-platform path, but it looks like a much bigger project, not a small follow-up. So the immediate question is whether we keep this PR as a pragmatic Windows fix for now ?

shantur:
For a cross platform project, I think having a feature just to work on one won't be a good idea.

pascalandr:
Just to clarify, are you opposed to this being the Windows implementation specifically, or only to merging it without a clear path for macOS/Linux as well? I agree the feature itself should be cross-platform, even if the implementation may differ by OS.

shantur:
Yes that's correct. Thanks for understanding

pascalandr:
I think the most realistic cross-platform direction is to keep the same overall design on all 3 OSes, but use OS-specific trust-store integration.

## Common design

- Do not trust the remote server cert in the OS store.
- Instead, generate a CodeNomad-owned local cert for `localhost` / `127.0.0.1`.
- Start a local HTTPS loopback proxy.
- Open the remote window on that local HTTPS origin.
- Forward traffic from the proxy to the remote CodeNomad server.

That keeps normal browser semantics for:
- secure cookies
- `credentials: include`
- SSE / EventSource
- standard HTTPS origin behavior

So the feature can be cross-platform even if the implementation differs by OS.

## Windows

This is the approach in this PR and it is the only one actually validated so far.

- generate a persistent local cert
- trust it in `CurrentUser\Root`
- run a loopback HTTPS proxy on an ephemeral port
- keep the proxy single-target / single-session / window-scoped

## macOS

I think the same architecture should work.

- generate the same persistent local cert
- add it to the login keychain, not the system keychain
- mark it trusted for SSL
- run the same loopback HTTPS proxy

Most likely implementation path is via `security` / Keychain integration.

## Linux

Same proxy architecture in theory, but this is the hardest platform because trust stores are not uniform.

Likely approach:
- generate the same persistent local cert
- try OS-specific trust-store installation depending on environment
- examples: `trust anchor`, `update-ca-certificates`, `update-ca-trust`
- fall back to a guided/manual flow if we cannot reliably install trust automatically

So I think Linux should be treated as best-effort first, then tightened based on real distro testing.

## Practical conclusion

I do think the feature can be cross-platform, but probably not with one identical implementation.
The shared product model can stay the same; the trust-store plumbing will need to be OS-specific.
What do you think ?

shantur:
> Same proxy architecture in theory, but this is the hardest platform because trust stores are not uniform.
>  think Linux should be treated as best-effort first, then tightened based on real distro testing.

I don't want to inherit something that is so specific to platforms and it's not a native CodeNomad feature that's missing here. Please use a proxy, accept remote codenomad certs or run remote codenomad with proper free certs from Let's Encrypt

pascalandr:
That can only work in environments where the user controls DNS / reverse proxy / certificates, but in many work setups that is decided by IT/network admins, so it is not really a user-available solution. That is exactly why I explored a client-side workaround here.

Just to clarify, I did try hard on the other paths on Windows :

- accepting the remote self-signed cert directly in Tauri/WebView2: I could not get that to work at all
- reusing the existing local CodeNomad server/proxy path: I tried it, but it did not isolate the remote session correctly.
- using a Tauri custom protocol: I explored it, but it was a poor fit for normal HTTPS/cookie/SSE behavior
- The dedicated local HTTPS proxy was the ONLY approach that worked end-to-end.`

shantur:
Or you can use Electron version, it works fine, it has capability to allow self signed certificate

pascalandr:
Well it's a lot slower on my side, especialy in this case.
________________________________
De : Shantur Rathore ***@***.***>
Envoyé : dimanche 5 avril 2026 00:57
À : NeuralNomadsAI/CodeNomad ***@***.***>
Cc : Pascal André ***@***.***>; Mention ***@***.***>
Objet : Re: [NeuralNomadsAI/CodeNomad] fix(desktop): support self-signed remote HTTPS windows (PR #289)

[https://avatars.githubusercontent.com/u/586441?s=20&v=4]shantur left a comment (NeuralNomadsAI/CodeNomad#289)<https://github.com/NeuralNomadsAI/CodeNomad/pull/289#issuecomment-4187882951>

Or you can use Electron version, it works fine, it has capability to allow self signed certificate

shantur:
> Well it's a lot slower on my side, especialy in this case.

@pascalandr - Sorry to hear that. I can surely help you get it working with nginx if you are available on Discord.

pascalandr:
@Shantur, I think there may be a cleaner path now that SideCar support exists: keep the Windows/self-signed workaround outside CodeNomad in an external helper, and only use CodeNomad SideCars as the integration point. i'll try that.

shantur:
> @shantur, I think there may be a cleaner path now that SideCar support exists: keep the Windows/self-signed workaround outside CodeNomad in an external helper, and only use CodeNomad SideCars as the integration point. i'll try that.

Brilliant. If it doesn't work with Sidecar, the remote server can be worked with the Windows/Self-signed workaround. That would be a better experience IMO as it will give you access to keyboard shortcuts

shantur:
@pascalandr - Closing this as we have another way forward.

pascalandr:
@shantur  I tried the SideCar approach, but for some reason it pointed back to the local CodeNomad session instead of a properly isolated remote one. Hotkeys worked fine thought; droping it for now.
I also had to revert the proxy hardening to make the Windows self-signed remote window work again. I’m keeping that version in my fork for now, and I’ll try a similar approach on a Linux VM to see if this can become a more credible multi-platform path.

shantur:
Thanks

PR #307, opened 2026-04-09
feat(plugin): add session title rename tool

### LINKED ISSUE #306: Add plugin tool for renaming the active session title
CodeNomad should support renaming the active OpenCode session title through the plugin tool bridge so the assistant can update a conversation title when the user explicitly asks for it.

## Problem

Today, session renaming exists in the UI, but not through the plugin tool layer used by the assistant during a conversation.

That creates a gap between what users can do manually in the interface and what they can ask the assistant to do conversationally. A user can ask to rename the session, but there is no dedicated supported path for updating the actual session title through the assistant.

## Goal

Add a dedicated plugin tool that renames the current session title.

## Scope

In scope:

- Add a plugin tool dedicated to session title renaming
- Register the tool in the CodeNomad OpenCode plugin entry point
- Add a server-side path that updates the current session title
- Preserve the existing session targeting behavior so the correct session is renamed

## Requirements

1. The assistant must be able to rename the current session through a tool call.
2. The tool must fail clearly when there is no active session context.
3. Empty or whitespace-only titles must be rejected.
4. The server bridge must forward the rename to the correct workspace instance with the correct HTTP method and auth.
5. The implementation must reuse the existing session/worktree semantics instead of introducing a second title store.

## Proposed Implementation

1. Add a `rename_session` tool under `packages/opencode-config/plugin/lib/session.ts`.
2. Register that tool from `packages/opencode-config/plugin/codenomad.ts`.
3. Add a plugin endpoint under `packages/server/src/server/routes/plugin.ts` for session title updates.
4. Forward the request to the OpenCode instance so the session title is updated through the existing session system.

## Verification

Suggested verification steps:

1. Start CodeNomad with the updated plugin and server code.
2. Ask the assistant to rename the current session.
3. Confirm the title changes in the session list and active session UI.
4. Confirm invalid titles are rejected.
5. Confirm the rename still applies when the session is associated with a non-root worktree context.

VooDisss:
Fixes #306

## Summary

This change adds a dedicated plugin tool for renaming the active session title and wires it through the existing CodeNomad plugin bridge to the correct OpenCode workspace instance.

The goal is to let the assistant perform the same session title rename operation that already exists in the UI, without introducing a separate metadata store or bundling in the still-unsettled description feature.
image

## Why

Users can already rename sessions manually in the UI, but there was no supported end-to-end conversational path for the assistant to do the same thing.

That mismatch meant the assistant could acknowledge a rename request without having a repo-native implementation path to update the actual session title. This PR closes that gap with the smallest correct change and keeps the scope intentionally limited to title renaming only.

## What Changed

1. Added a new `rename_session` plugin tool in `packages/opencode-config/plugin/lib/session.ts`.
2. Registered the session tool set in `packages/opencode-config/plugin/codenomad.ts`.
3. Added a plugin server endpoint in `packages/server/src/server/routes/plugin.ts` that proxies title updates to the workspace OpenCode instance.
4. Forwarded the current tool execution `directory` so the rename follows the existing worktree-aware session targeting behavior.
5. Corrected the proxy implementation to use `PATCH`, matching the real OpenCode SDK `session.update` contract.

## Scope Boundaries

Included:

- Session title rename tool
- Plugin registration
- Server bridge for title updates
- Worktree-aware forwarding via directory context

Not included:

- Description persistence
- Description rendering in the UI
- Session tooltip changes
- Any broader session metadata model changes

## Technical Notes

The OpenCode SDK contract for session updates is not `POST`; it is `PATCH /session/{sessionID}` with a body containing the updatable fields such as `title`.

That detail matters because the first implementation path could appear successful while failing to apply the rename if the bridge used the wrong HTTP method. The final implementation aligns the server proxy with the generated SDK contract so the rename is applied by the actual OpenCode session update endpoint.

## Files Changed

- `packages/opencode-config/plugin/codenomad.ts`
- `packages/opencode-config/plugin/lib/session.ts`
- `packages/server/src/server/routes/plugin.ts`

## Verification

Performed:

1. Verified the generated OpenCode SDK contract for `session.update`.
2. Confirmed the plugin can call the new bridge endpoint.
3. Confirmed the feature works after switching the proxied request from `POST` to `PATCH`.
4. Rebuilt `packages/server` successfully.

Build note:

- The server build passes.
- The build still emits existing UI/Vite warnings unrelated to this change.

## Risks and Follow-up

1. Description handling remains intentionally postponed and should be handled as a separate follow-up once the storage and UX direction is settled.
2. If future session metadata tools are added, they should continue to reuse the same plugin bridge pattern instead of creating parallel storage behavior.

### DIFF
--- a/packages/opencode-config/plugin/codenomad.ts
+++ b/packages/opencode-config/plugin/codenomad.ts
@@ -1,6 +1,7 @@
 import type { PluginInput } from "@opencode-ai/plugin"
 import { createCodeNomadClient, getCodeNomadConfig } from "./lib/client"
 import { createBackgroundProcessTools } from "./lib/background-process"
+import { createSessionTools } from "./lib/session"
 
 let voiceModeEnabled = false
 
@@ -8,6 +9,7 @@ export async function CodeNomadPlugin(input: PluginInput) {
   const config = getCodeNomadConfig()
   const client = createCodeNomadClient(config)
   const backgroundProcessTools = createBackgroundProcessTools(config, { baseDir: input.directory })
+  const sessionTools = createSessionTools(config)
 
   await client.startEvents((event) => {
     if (event.type === "codenomad.ping") {
@@ -29,6 +31,7 @@ export async function CodeNomadPlugin(input: PluginInput) {
   return {
     tool: {
       ...backgroundProcessTools,
+      ...sessionTools,
     },
     async "chat.message"(_input: { sessionID: string }, output: { message: { system?: string } }) {
       if (!voiceModeEnabled) {
--- a/packages/opencode-config/plugin/lib/session.ts
+++ b/packages/opencode-config/plugin/lib/session.ts
@@ -1,0 +1,49 @@
+import { tool } from "@opencode-ai/plugin/tool"
+import { createCodeNomadRequester, type CodeNomadConfig } from "./request"
+
+type SessionRenameResponse = {
+  sessionID: string
+  title: string
+}
+
+export function createSessionTools(config: CodeNomadConfig) {
+  const requester = createCodeNomadRequester(config)
+
+  const request = async <T>(path: string, init?: RequestInit): Promise<T> => {
+    return requester.requestJson<T>(path, init)
+  }
+
+  return {
+    rename_session: tool({
+      description:
+        "Rename the current session when the user asks to change the chat title. Use a short descriptive title that reflects the current task.",
+      args: {
+        title: tool.schema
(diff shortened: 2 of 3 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
We should close this. I remember having a discussion with VooDoo and we need to first test this as a separate opencode plugin to understand how LLMs handle this.

PR #309, opened 2026-04-09
fix(ui): stop forced auto-follow during streaming

### LINKED ISSUE #308: Fix sticky auto-scroll while chat responses are streaming
# Fix sticky auto-scroll while chat responses are streaming

## Summary

While an assistant response is streaming in chat, auto-scroll is too sticky.

If a user tries to scroll upward to inspect earlier parts of the response before streaming finishes, the view tends to keep pulling them back toward the bottom. In practice, users often have to scroll up repeatedly, quickly, or forcefully to escape follow mode.

## Problem

Long streamed responses are difficult to inspect in real time because the viewport keeps trying to follow the newest output even after the user has started scrolling upward.

That makes it hard to reference earlier lines, compare parts of the same answer, or review content while generation is still in progress.

## Steps to Reproduce

1. Open a chat/session with a long streamed assistant response.
2. Send a prompt that causes the assistant to output many lines over time.
3. While the response is still streaming, try to scroll upward to inspect earlier content.

## Actual Behavior

The chat view keeps trying to follow the newest output, making it difficult to move upward and stay there while the stream is still in progress.

## Expected Behavior

If the user scrolls upward during streaming, auto-scroll should disengage promptly and respect that choice. The user should be able to inspect earlier content without fighting the viewport.

## User Impact

This makes long streamed responses harder to read and review in real time, especially when the user wants to reference earlier lines before generation has finished.

## Notes

The issue appears to be in the chat scroll-follow behavior and the logic that decides when user scroll intent should disable follow mode.

VooDisss:
# PR Draft: Fix sticky auto-scroll during streaming chat responses

Fixes #308

## Summary

This change makes chat auto-scroll easier to escape while assistant output is still streaming.

The goal is to stop the viewport from repeatedly pulling the user back toward the bottom once they begin scrolling upward to inspect earlier content.

## Why

Before this change, streaming updates could keep reasserting bottom-follow behavior during active rendering. That made auto-scroll feel sticky and forced users to scroll repeatedly or forcefully just to review earlier parts of an in-progress response.

The intended behavior is simpler: once the user scrolls upward to leave follow mode, the UI should respect that decision instead of fighting it during subsequent stream updates.

## What Changed

1. Removed render-time force-bottom behavior from the shared follow-scroll helper path.
2. Updated streamed reasoning output to restore scroll without forcing the viewport back to the bottom.
3. Updated streamed tool-call output to use the same non-forcing restore behavior.

## Scope Boundaries

Included:

- Sticky auto-scroll behavior during streamed chat output
- Shared follow-scroll behavior used by streamed nested panes
- Reasoning and tool-call streaming paths that reused the same forced follow behavior

Not included:

- A full rewrite of the virtualized message list follow model
- Broader scroll UX changes outside the streaming follow/escape behavior
- Unrelated UI or plugin configuration changes in the worktree

## Technical Notes

The core problem was not basic auto-scroll itself, but a render-time path that could keep forcing bottom-follow behavior while new streamed content was arriving.

That meant a user's attempt to scroll upward could be overridden repeatedly by subsequent stream updates, which is why the auto-scroll felt sticky. The fix removes that override and keeps render-time restoration dependent on the current follow state instead.

## Files Changed

- `packages/ui/src/lib/follow-scroll.tsx`
- `packages/ui/src/components/message-block.tsx`
- `packages/ui/src/components/tool-call.tsx`

## Verification

Performed:

1. Reproduced the sticky auto-scroll behavior with a long multi-line streaming response.
2. Verified that scrolling upward during streaming now disengages follow more naturally in the affected streamed panes.
3. Ran `npm run typecheck --workspace @codenomad/ui`.
4. Ran `npm run build --workspace @codenomad/ui`.

Build note:

- The UI typecheck passes.
- The UI build succeeds.
- The build still emits existing third-party and chunk-size warnings unrelated to this change.

## Risks and Follow-up

1. The broader scroll-follow model is still more heuristic-heavy than ideal, so there may be future follow-up work to simplify it further.
2. This PR intentionally applies the smallest targeted fix to the known snap-back path instead of rewriting the full chat scroll system.

### DIFF
--- a/packages/ui/src/components/message-block.tsx
+++ b/packages/ui/src/components/message-block.tsx
@@ -1344,9 +1344,7 @@ function ReasoningStreamOutput(props: {
     if (preRef && preRef.textContent !== nextText) {
       preRef.textContent = nextText
     }
-    if (followScroll.autoScroll()) {
-      followScroll.restoreAfterRender({ forceBottom: true })
-    }
+    followScroll.restoreAfterRender()
     notifyContentRendered()
   })
 
--- a/packages/ui/src/components/tool-call.tsx
+++ b/packages/ui/src/components/tool-call.tsx
@@ -454,7 +454,7 @@ function ToolCallDetails(props: {
 
   createEffect(() => {
     if (followScroll.autoScroll()) {
-      scrollHelpers.restoreAfterRender({ forceBottom: true })
+      scrollHelpers.restoreAfterRender()
     }
   })
 
--- a/packages/ui/src/components/tool-call/types.ts
+++ b/packages/ui/src/components/tool-call/types.ts
@@ -47,7 +47,7 @@ export interface ToolScrollHelpers {
   registerContainer(element: HTMLDivElement | null, options?: { disableTracking?: boolean }): void
   handleScroll(event: Event & { currentTarget: HTMLDivElement }): void
   renderSentinel(options?: { disableTracking?: boolean }): JSXElement | null
-  restoreAfterRender(options?: { forceBottom?: boolean }): void
+  restoreAfterRender(): void
 }
 
 export interface ToolRendererContext {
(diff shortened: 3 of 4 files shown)


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

PR #311, opened 2026-04-10
Refactor Git Changes workflow and diff handling

### LINKED ISSUE #310: Improve Git Changes diff workflow and source-control ergonomics
# GitDiffViewer: Current Problems and Why This PR Is Needed

## Summary

The current Git diff experience in CodeNomad has several functional and UX issues that make source-control review harder than it needs to be. These issues affect both correctness and usability.

This document describes the current problems as they exist before the proposed changes in this PR.

---

## 1. Git changes are presented as a flat list

The current Git Changes tab does not clearly separate:

1. staged changes
2. unstaged changes

As a result:

1. users cannot quickly understand what is already staged versus what is still pending
2. review flow is slower because the list does not match the mental model used by common Git tools
3. the panel feels more like a raw diff browser than a source-control workflow

### Why this matters

For Git review, “what is staged” is often the first question a user needs answered. Without that separation, users have to infer state manually.

---

## 2. Diff behavior is not section-aware

The viewer currently lacks a reliable distinction between:

1. staged diff content
2. unstaged diff content

When the same file has both staged and unstaged changes, the current behavior can show the wrong comparison or fail to distinguish which version of the file is being reviewed.

### Why this matters

Git users expect:

1. staged changes to compare `HEAD -> index`
2. unstaged changes to compare `index -> working tree`

If the viewer cannot make that distinction, the review surface becomes misleading.

---

## 3. Selection identity is too weak for Git semantics

When one path can appear in multiple Git states, using only a file path as the selected identity is not sufficient.

This creates problems such as:

1. wrong diff content appearing after refresh
2. selection instability when a file moves between staged and unstaged sections
3. stale async diff results overwriting newer selection state

### Why this matters

In Git UI, a file path alone is not always enough to describe the selected item. The state or section matters too.

---

## 4. Binary-file handling is inconsistent

The current diff path can still imply line-oriented behavior for files that should not be treated as line-based content.

Examples of the resulting problems include:

1. misleading line counts for binaries
2. viewer behavior that implies text semantics for non-text content
3. prompt-context features that are only meaningful for text diffs being exposed around binary cases

### Why this matters

If a file is binary, the UI should not act like text-based line review is valid.

---

## 5. Manual refresh is too important in normal use

Without stronger refresh behavior, the user must rely too much on the refresh button to keep the Git Changes panel current.

This becomes especially noticeable when:

1. files are modified by agent tools
2. a patch is applied externally
3. the selected diff remains stale while the underlying file has already changed

### Why this matters

The Git Changes panel should feel responsive to real file mutations, especially when they are initiated by the same environment that is rendering the panel.

---

## 6. The current row actions waste space and are visually noisy

When action affordances permanently occupy row width, the file list becomes harder to scan.

The result is:

1. less room for file names and counts
2. more visual clutter in idle state
3. a weaker sense of hierarchy between filename, path, counts, and actions

### Why this matters

The file list is primarily a navigation and review surface. Actions should be available without dominating the resting layout.

---

## 7. Stage / unstage is missing from the current Git workflow

Without integrated per-file stage/unstage actions, the user must leave the review flow or use external tools to perform basic SCM operations.

This breaks the intended workflow in the right panel.

### Why this matters

If the panel already presents Git state, it should support the most basic Git transitions directly:

1. stage a file
2. unstage a file
3. review the result immediately

---

## 8. Commit composition is disconnected from staged review

The current experience does not provide a coherent way to:

1. review staged changes
2. write a commit message
3. commit directly from the same workflow surface

### Why this matters

Once staged and unstaged sections exist, the next natural user action is committing the staged set. That should not require switching contexts unnecessarily.

---

## 9. Prompt-context insertion from Git review is underpowered or awkward

The Git diff viewer is a useful place to ask the assistant about specific changed code, but without a smooth path into the prompt, the review-to-assistant workflow is awkward.

Common issues include:

1. no clear line-anchored action from the diff itself
2. extra friction to reference a changed line or range in conversation
3. missing connection between the diff viewer and the active prompt input

### Why this matters

The Git diff viewer is one of the most context-rich places in the app. It should be possible to carry that context into the prompt without manual retyping.

---

## 10. The UI does not align well with established Git review expectations

Compared with tools users are already familiar with, the current Git diff experience lacks several expected affordances:

1. staged / unstaged grouping
2. compact, scan-friendly list layout
3. direct row actions for Git state changes
4. a commit message surface near staged content
5. smoother line-level interaction inside the diff viewer

### Why this matters

Users bring strong expectations from tools like VS Code and GitHub. When the local review surface differs too much, even correct functionality feels harder to use.

---

## What the PR should achieve

To address the current problems, the PR should aim to provide:

1. clear staged / unstaged grouping
2. correct section-aware diff behavior
3. stable selection identity and refresh behavior
4. safe handling of binary and deleted files
5. integrated stage / unstage actions
6. a commit compose flow tied to staged changes
7. line-aware prompt-context insertion from the diff viewer
8. a more compact and navigable row layout
9. reduced dependence on manual refresh

---

VooDisss:
# Git Changes PR Review Context

Fixes: #310 

## Purpose of this document

This document is intended to give a PR reviewer or gatekeeper enough neutral context to review the Git Changes feature series accurately.

## BEFORE/AFTER SNAPSHOT:

image

It distinguishes:

1. the intended scope of the work
2. implementation choices that were deliberate
3. behaviors that were explicitly tested and accepted during development
4. remaining follow-up areas that were not part of the required intent

It should not be treated as a request to approve the PR automatically. It exists to reduce false-positive review findings caused by missing context.

---

## High-level scope

The work in this series refactors and extends the existing `Git Changes` tab in the right panel.

The intended feature scope includes:

1. grouped staged / unstaged change presentation
2. correct section-aware diff loading
3. per-file stage / unstage controls
4. commit message compose box and commit action for staged changes
5. prompt-context insertion from the Git diff viewer
6. auto-refresh behavior that reduces dependence on the manual refresh button

This work is intentionally implemented inside the existing Git Changes vertical slice rather than as a new SCM subsystem.

---

### Server / API surface

The following server areas were intentionally extended:

1. `packages/server/src/api-types.ts`
2. `packages/server/src/events/bus.ts`
3. `packages/server/src/server/http-server.ts`
4. `packages/server/src/server/routes/workspaces.ts`
5. `packages/server/src/workspaces/git-status.ts`
6. `packages/server/src/workspaces/git-mutations.ts`
7. `packages/server/src/workspaces/worktree-directory.ts`
8. `packages/server/src/workspaces/instance-events.ts`

### UI surface

The following UI areas were intentionally extended:

1. `packages/ui/src/components/file-viewer/monaco-diff-viewer.tsx`
2. `packages/ui/src/components/instance/instance-shell2.tsx`
3. `packages/ui/src/components/instance/shell/right-panel/RightPanel.tsx`
4. `packages/ui/src/components/instance/shell/right-panel/git-changes-model.ts`
5. `packages/ui/src/components/instance/shell/right-panel/tabs/GitChangesTab.tsx`
6. `packages/ui/src/components/instance/shell/right-panel/types.ts`
7. `packages/ui/src/components/instance/shell/storage.ts`
8. `packages/ui/src/components/prompt-input.tsx`
9. `packages/ui/src/components/prompt-input/types.ts`
10. `packages/ui/src/components/session/session-view.tsx`
11. `packages/ui/src/lib/api-client.ts`
12. `packages/ui/src/lib/i18n/messages/*/instance.ts`
13. `packages/ui/src/styles/panels/right-panel.css`

---

## Intentional product and architecture decisions

The following outcomes were deliberate and should not be flagged as issues merely because they exist.

### Git status / diff architecture

1. The UI does not rely only on the proxied OpenCode `file.status()` payload.
2. CodeNomad adds server-backed worktree Git status and diff endpoints to expose staged / unstaged semantics correctly.
3. Server-backed worktree mutation endpoints were added for:
   - stage
   - unstage
   - commit
4. The existing event bus / SSE channel is reused for Git invalidation, instead of adding a bespoke invalidation route.

### Git Changes UI structure

1. The file list is grouped into:
   - `Staged Changes`
   - `Changes`
2. Both sections are collapsible.
3. Section open state is persisted.
4. The same file may appear in both sections when Git state genuinely requires that.
5. Rows are filename-first, with parent path as secondary text.
6. Rows are intentionally compact compared to the original flat list.

### Diff behavior

1. Diff loading is section-aware.
2. Deleted files are supported in grouped mode.
3. Binary files are treated as non-line-oriented in the diff viewer.
4. Binary diffs suppress line-based prompt-context affordances.

### Stage / unstage / commit workflow

1. Stage and unstage are per-file row actions.
2. Bulk stage-all / unstage-all was intentionally not added.
3. The commit compose box is intentionally rendered inside the `Staged Changes` section.
4. The commit button is intentionally overlaid inside the commit input area.
5. The current commit compose flow is minimal by design:
   - no push
   - no amend flow
   - no branch management

### Prompt-context insertion

1. Prompt insertion is intentionally an HTML comment marker, not a full diff payload.
2. The expected inserted form is:

   `<!-- Git change context: <path> lines X-Y -->`

3. The trigger UI is intentionally a seam/gutter action in the Monaco diff viewer, not a toolbar button.

### Row action reveal behavior

1. Stage / unstage row actions are intentionally hover-revealed on hover-capable layouts.
2. The row action reveal intentionally uses:
   - delayed hide
   - slight stats fade/shift
   - compact idle width
3. On non-hover layouts, the action remains visible for reliability.

### Auto-refresh behavior

The accepted refresh model is intentionally hybrid:

1. refresh on Git Changes tab activation
2. 20-second polling only while the Git Changes tab is active
3. immediate invalidation from completed raw tool events for:
   - `write`
   - `edit`
   - `apply_patch`

This hybrid model is intentional. Polling remains as a fallback even after tool-event invalidation.

---

## Behaviors explicitly tested during development

The following behaviors were explicitly exercised during development and used to guide fixes.

### Grouped staged / unstaged behavior

1. files appear in the correct staged / unstaged sections
2. section collapse / expand works
3. collapse state persists
4. line counts are section-specific

### Diff behavior

1. staged diff loads differently from unstaged diff
2. deleted-file handling was verified and corrected
3. binary-file rendering was corrected to avoid line-oriented behavior
4. untracked binary files no longer report fake text line counts

### Mutation behavior

1. per-file stage works from `Changes`
2. per-file unstage works from `Staged Changes`
3. stage / unstage selection remapping was exercised and corrected
4. unborn-repo unstage behavior was explicitly hardened

### Prompt-context behavior

1. selected line / range insertion was tested
2. button placement in the Monaco seam/gutter was iterated and verified

### Auto-refresh behavior

1. tab-activation refresh was tested
2. 20-second active-tab polling was tested
3. raw completed tool invalidation was tested in the running UI for:
   - `write`
   - `edit`
   - `apply_patch`
4. stale async overwrite and stale selection restoration bugs were found and fixed through review/testing

---

## Review findings that were investigated and are no longer intended blocker topics

The following areas were previously raised by strict reviews and then either fixed or determined to be acceptable within scope.

### Fixed in the current series

1. duplicate stage / unstage firing
2. stale diff response overwriting newer selection
3. passive refresh restoring a stale selection
4. instance-wide invalidation overreach
5. selected diff staying stale after tool invalidation
6. worktree-switch status races
7. unhandled rejection risk from async invalidation publication
8. queued invalidation intent being lost during in-flight refresh
9. `git-diff` path traversal / absolute path boundary issue

### Investigated and considered non-blocking within current intent

1. split add/delete presentation for tracked rename behavior
   - this was compared against VS Code behavior during manual testing
   - no stage/unstage corruption was observed in the tested flow
   - this is currently treated as a representation tradeoff, not a proven blocker

---

## Remaining non-blocker follow-up areas

The following are still reasonable follow-up topics, but they were not part of the required blocker-fix scope.

1. normalize directory-to-worktree matching more aggressively on Windows so tool invalidation works more reliably from nested directories or path-format variations
2. improve keyboard discoverability of hover-revealed stage / unstage actions
3. reserve textarea space for the overlaid commit button if the overlay tradeoff is reconsidered
4. reduce size/complexity in:
   - `RightPanel.tsx`
   - `right-panel.css`
5. tighten raw SSE tool-event parsing into a more explicit helper if that event bridge grows further

These follow-ups should not be interpreted as evidence that the core implementation is incomplete unless a reviewer finds a new concrete failure.

---

## Suggested review focus

If a gatekeeper or reviewer is evaluating this PR, the most useful focus areas are:

1. whether staged / unstaged behavior is correct for normal Git workflows
2. whether the new server worktree Git endpoints remain narrowly scoped
3. whether auto-refresh remains bounded to the active Git Changes context
4. whether the explicit fixes for stale async behavior and invalidation races are sufficient
5. whether any unintentional server boundary broadening or state corruption remains

Less useful review topics, unless tied to a concrete failure, are:

1. preference disagreements with accepted prompt insertion format
2. preference disagreements with the overlaid commit button placement
3. preference disagreements with keeping polling fallback alongside tool invalidation
4. objections to server-backed Git endpoints purely because they add surface area

---

## Summary

This series intentionally evolves the existing Git Changes tab into a more complete source-control workflow for:

1. grouped staged / unstaged inspection
2. section-aware diffs
3. per-file staging and unstaging
4. commit composition for staged changes
5. prompt-context insertion from Git diffs
6. bounded auto-refresh for both passive viewing and agent-driven file mutations

The intended review standard is to find concrete correctness, layering, or maintenance problems that remain after this series — not to re-argue the already accepted product choices listed above.


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

Any better now?

shantur:
@CodeNomadBot

Any better now?

shantur:
@CodeNomadBot

What are the changes in package/server doing?

shantur:
@CodeNomadBot

When is workspace.fileChanged event sent?
Any other events does server or plugin sends for UI from our codebase?
Be ruthless about the packages/server changes and git handling. Where can it fail, what can lead to issues, does it increase any dependency, does Git client need to be present, what happens if Git is absent? and forgotten edge cases?

shantur:
@VooDisss 

1.
> event.properties.part.type === "tool".
> event.properties.part.state.status === "completed".
> event.properties.part.tool is one of:
> write
> edit
> apply_patch

Depending on the hard coded tool calls might not be the best approach here.  Any new tool call or a custom tool call editing the files will fire the event. Instead we should rely on the earlier event. If I'm not wrong it was session.updated. That brings in the changes. 

2. Missing gig binary needs to be handled better way. The new roads should return a proper clean error message rather than failing blindly. 

3.  Work tree path matching should be improved. It's actually called out by the PR itself. We need to fix it. 

4. If the workspace is not Git repository, that case needs to be handled correctly. 

5. More to come

shantur:
@CodeNomadBot

From your message above

1. Git commit path assumes repo-local config/user setup is valid

What happens in case use isn't setup? What happens to the UI / UX

2.  Binary detection is heuristic

Is the git binary run from the user environment as the node / opencode binary are run in packages/server?

shantur:
@CodeNomadBot

any better now?

PR #318, opened 2026-04-11
Add LLM-Assisted Context Prune Selection

VooDisss:
## User Workflow
1. The user asks the LLM to use the context-pruning tool to select irrelevant messages.
2. The LLM calls the tool with the badge range it wants to stage.
3. The UI automatically selects those badges in the existing delete-selection flow.
4. The user reviews the staged selection and either deletes everything or adjusts the selection manually in the UI before clicking the trashcan.

## Screenshot 
image

## Summary
This PR adds a minimal context-pruning flow where the LLM can stage badge selections for removal, and the user then reviews or adjusts that selection before deleting anything.

## Why
The previous flow required users to enter selection mode and manually pick every badge themselves, which is slow and tedious in longer sessions.

## What Changed
- Added a tool path for selecting context badges by range.
- Reused the existing delete-selection UI instead of adding a second delete flow.
- Kept deletion as a manual user confirmation step.

## Help get this merged, guys
image

Observe the sacred pile of review stones. They are not for throwing at @shantur, no matter how mysteriously delayed the merge may seem. They are for channeling your energy into a respectful but persistent request that he review and merge the context management tooling.

### DIFF
--- a/packages/opencode-config/plugin/codenomad.ts
+++ b/packages/opencode-config/plugin/codenomad.ts
@@ -1,6 +1,7 @@
 import type { PluginInput } from "@opencode-ai/plugin"
 import { createCodeNomadClient, getCodeNomadConfig } from "./lib/client"
 import { createBackgroundProcessTools } from "./lib/background-process"
+import { createContextPruneTools } from "./lib/context-prune"
 
 let voiceModeEnabled = false
 
@@ -8,6 +9,7 @@ export async function CodeNomadPlugin(input: PluginInput) {
   const config = getCodeNomadConfig()
   const client = createCodeNomadClient(config)
   const backgroundProcessTools = createBackgroundProcessTools(config, { baseDir: input.directory })
+  const contextPruneTools = createContextPruneTools(config)
 
   await client.startEvents((event) => {
     if (event.type === "codenomad.ping") {
@@ -29,6 +31,7 @@ export async function CodeNomadPlugin(input: PluginInput) {
   return {
     tool: {
       ...backgroundProcessTools,
+      ...contextPruneTools,
     },
     async "chat.message"(_input: { sessionID: string }, output: { message: { system?: string } }) {
       if (!voiceModeEnabled) {
--- a/packages/opencode-config/plugin/lib/context-prune.ts
+++ b/packages/opencode-config/plugin/lib/context-prune.ts
@@ -1,0 +1,77 @@
+import { tool } from "@opencode-ai/plugin/tool"
+import { createCodeNomadRequester, type CodeNomadConfig } from "./request"
+
+type ContextPruneRouteRequest = {
+  sessionID: string
+  indices: number[]
+}
+
+export function createContextPruneTools(config: CodeNomadConfig) {
+  const requester = createCodeNomadRequester(config)
+
+  return {
+    select_context_range: tool({
+      description: "Select existing conversation badges for context pruning by 1-based range indices so the user can review and delete them in the UI.",
+      args: {
+        range: tool.schema.string().describe("Comma-separated badge indices and ranges, e.g. 1,3-5,6"),
+      },
+      async execute(args, context) {
+        const indices = parseRange(args.range)
+        await requester.requestVoid("/context-prune/select", {
+          method: "POST",
+          body: JSON.stringify({
+            sessionID: context.sessionID,
+            indices,
+          } satisfies ContextPruneRouteRequest),
+        })
+
+        return ""
+      },
+    }),
+  }
+}
+
+function parseRange(input: string): number[] {
+  const raw = (input ?? "").trim()
+  if (!raw) {
+    throw new Error("Range is required")
+  }
+
+  const values = new Set<number>()
+  const tokens = raw.split(",")
+
+  for (const token of tokens) {
+    const part = token.trim()
+    if (!part) {
+      throw new Error("Range contains an empty entry")
+    }
+
+    const rangeMatch = part.match(/^(\d+)-(\d+)$/)
+    if (rangeMatch) {
+      const start = Number(rangeMatch[1])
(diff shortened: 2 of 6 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
No stones reached at me yet VooDoo.

pascalandr:
## Recommended direction for this feature

The current implementation should be redesigned rather than carried forward, primarily because its positional addressing does not establish a reliable shared view between the model and the UI.

### Why badge positions are not a stable protocol

`select_context_range` asks the model to translate entries such as `m0007` into 1-based positions in the UI's current deletable badge list. However, this branch does not expose such a badge manifest to the model, and model messages do not map one-to-one to timeline segments:

- one assistant message can produce several assistant/tool badges;
- some parts are omitted or grouped by the timeline;
- tool parts can be selected independently from their containing message;
- the latest compaction changes the eligible range;
- streaming and part updates rebuild positional timeline segment IDs;
- messages, parts, or compaction state can change between the tool call and UI consumption.

As a result, “position 7 in model-visible context” and “badge 7 in the current deletable timeline” are not guaranteed to identify the same operation. Keeping deletion behind the trash button limits damage, but the staged state still needs to be unambiguous and trustworthy.

### Existing safety issues to address

- A command that resolves to no current badge leaves the previous deletion selection visible. This conflicts with the documented replacement semantics and could make the user review an older selection while believing it came from the latest tool call.
- Pending commands are keyed only by `sessionId`; they need at least `{instanceId, sessionId}` isolation.
- Mixed valid and invalid indices silently stage a partial subset instead of failing atomically.
- Commands are consumed as soon as any timeline content exists, without proving that loading completed or that the compaction boundary still matches.
- The range parser expands a range before enforcing the 1,000-item limit. Inputs such as `1-999999999` can consume excessive CPU/memory before rejection; non-safe integers can make this worse.
- The staging event is sent as an `instance.event`, even though it is a CodeNomad UI command rather than an event originating from the OpenCode stream. In the current architecture this can also affect connection-status handling.
- The implementation has no focused tests for parser limits, replacement behavior, stale state, instance isolation, compaction races, or message/tool-part resolution.

### Suggested replacement protocol

Use a snapshot with stable opaque handles instead of badge positions.

#### 1. Add `inspect_prunable_context`

The tool should return a bounded, model-visible manifest generated from the current session, for example:

```text
p1  message  850 tokens  "Earlier implementation request..."
p2  tool     420 tokens  "Read result for an obsolete approach..."
p3  message  610 tokens  "Superseded assistant response..."
```

Each handle should resolve server-side to a stable deletion target:

```ts
type ContextPruneTarget =
  | { kind: "message"; messageId: string }
  | { kind: "tool"; messageId: string; partId: string }
```

The snapshot should include an ID, creation time, current compaction boundary/revision, and only currently eligible post-compaction targets.

#### 2. Add `stage_context_prune`

Accept the snapshot and handles selected from that manifest:

```json
{
  "snapshotId": "...",
  "handles": ["p1", "p2"],
  "reasons": {
    "p1": "Superseded requirement",
    "p2": "Output is no longer needed"
  }
}
```

The tool should return a clear acknowledgement such as “Staged 2 items for user review; nothing was deleted.”

#### 3. Store the authoritative proposal server-side

Key proposals by `{instanceId, sessionId}` and store:

- `proposalId` and `snapshotId`;
- stable message/part targets;
- compaction boundary or session revision;
- creation time and expiry;
- optional model reasons and estimated token savings.

This avoids losing an event when no UI is connected and lets a reconnecting UI retrieve the current proposal.

#### 4. Publish a typed CodeNomad workspace event

Prefer a first-class event such as `contextPrune.proposalChanged` rather than routing the command through OpenCode's `instance.event` envelope. Include `instanceId`, `sessionId`, and `proposalId`; the UI should then read the authoritative proposal.

#### 5. Revalidate atomically in the UI

Before replacing the current selection, verify that:

- the instance and session match;
- message loading has completed;
- the compaction boundary/revision still matches;
- every message and part still exists and remains eligible.

If any target is stale, reject the complete proposal and ask for regeneration rather than silently staging a subset. Stable message/part IDs should remain the source of truth; derive current timeline IDs only for visual highlighting.

### Preserve the existing confirmation path

The tool must continue to stage only. The user should still be able to add or remove items, cancel the proposal, inspect token savings/reasons, and use the existing trash button as the sole mutation trigger. No new auto-delete route is needed.

### Current-code integration notes

- The active plugin now lives under `packages/opencode-plugin/plugin/`, not `packages/opencode-config/plugin/`.
- Keep the protocol, proposal store, and pure target resolver outside `message-section.tsx`; that component is already very large.
- Use the current authenticated plugin requester and strict bounded Zod schemas.
- Treat the route workspace ID as the authoritative instance ID; never trust an instance ID from the request body.

### Minimum regression coverage

- stable handle resolution for whole messages and exact tool parts;
- latest-compaction eligibility and boundary changes;
- isolation of identical session IDs across instances;
- replacement, expiry, duplicate delivery, and reconnect behavior;
- message/part removal or timeline rebuild before application;
- all-or-nothing rejection of stale targets;
- oversized/malformed payload rejection without range expansion;
- proof that staging never invokes a deletion API;
- cancellation and manual adjustment before trash confirmation.

In short: this PR should be replaced with a focused new PR that keeps the human-in-the-loop workflow and current deletion UI but uses a model-visible snapshot and stable message/part targets.

pascalandr:
OpenCode V2 removed the API behind our V1 pruning system, so I rebuilt pruning from scratch for CodeNomad 0.20.0 using a dedicated plugin and RPC. This PR’s integration is no longer compatible with that replacement, and LLM-assisted selection no longer fits its simpler workflow. This also supersedes my earlier redesign suggestion.

I’m therefore closing this PR. Thanks again for the work and the stones you put into it!

PR #320, opened 2026-04-12
fix(tauri): own Windows CLI subtree with a job object

pascalandr:
## Summary
- Follow-up to #240 to make Windows desktop shutdown reliable this time, even when the tracked CLI wrapper PID exits before its descendants
- Attach the spawned CLI process to a Windows Job Object with `KILL_ON_JOB_CLOSE`, so the desktop app owns the whole subtree instead of relying only on `taskkill /PID <wrapper> /T`
- Keep the current graceful-then-force shutdown path, but add a robust OS-level fallback that reaps orphaned workspace processes when the wrapper is already gone

## Root Cause
The previous Windows shutdown logic still depended on the PID tracked by Tauri. In practice that PID can be a short-lived Node wrapper. Once that wrapper exits, `taskkill` can report success or PID-not-found while descendants remain alive, and the desktop app no longer has a reliable handle to reap them.

## Validation
- `cargo check --manifest-path packages/tauri-app/src-tauri/Cargo.toml`
- `cargo build --release --manifest-path packages/tauri-app/src-tauri/Cargo.toml`
- Manual local test: orphaned processes are cleaned up after desktop shutdown

### DIFF
--- a/packages/tauri-app/src-tauri/Cargo.toml
+++ b/packages/tauri-app/src-tauri/Cargo.toml
@@ -12,20 +12,11 @@ tauri = { version = "2.5.2", features = [ "devtools"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 serde_yaml = "0.9"
-axum = "0.7"
-axum-server = { version = "0.7", features = ["tls-rustls"] }
-bytes = "1"
-futures-util = "0.3"
-rcgen = "0.13"
-rustls = { version = "0.23", features = ["ring"] }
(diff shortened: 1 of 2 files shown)


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

Do we need to align packages/electron-app similar to this change on Windows?

pascalandr:
Sorry, I had accidentally included unrelated manifest changes in the PR earlier. I've cleaned that up now.

shantur:
@CodeNomadBot

Can you suggest a patch for packages/electron-app to match Tauri behavior on Windows

shantur:
Thanks @pascalandr

PR #327, opened 2026-04-16
fix(ui): stabilize timeline follow scroll from bottom

pascalandr:
## Summary
- fix the sticky-bottom state where dragging the scrollbar to the bottom makes `PageUp` jump to the previous timeline block and then snap immediately back down
- keep the change scoped to `virtual-follow-list.tsx`, where follow mode, scroll intent, and bottom pinning are coordinated

## Root Cause
The list only disabled follow mode when it saw an explicit local "user intent" signal. After reaching the bottom through the native scrollbar, `PageUp` could move the viewport without tripping that path, so the next render notification re-enabled the bottom snap immediately.

## Validation
- `npx tsc --noEmit --project packages/ui/tsconfig.json`
- `npm run build --prefix packages/ui`
- manual desktop test: `PageUp` works again from the bottom sticky state

### DIFF
--- a/packages/ui/src/components/virtual-follow-list.tsx
+++ b/packages/ui/src/components/virtual-follow-list.tsx
@@ -175,6 +175,7 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
   let suppressAutoScrollOnce = false
   let pendingInitialScroll = true
   let lastObservedScrollOffset = 0
+  let lastObservedPinnedAtBottom = false
 
   const state: VirtualFollowListState = {
     autoScroll,
@@ -241,6 +242,7 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
 
     const offset = handle.scrollOffset
     const scrolledUp = offset < lastObservedScrollOffset - 1
+    const wasPinnedAtBottom = lastObservedPinnedAtBottom
     const scrollHeight = handle.scrollSize
     const clientHeight = element.clientHeight
     const atBottom = scrollHeight - (offset + clientHeight) <= (props.scrollSentinelMarginPx ?? DEFAULT_SCROLL_SENTINEL_MARGIN_PX)
@@ -256,8 +258,9 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
     // scrollbar). If follow mode stays enabled, the next render notification
     // snaps the list straight back to bottom. A real upward viewport move away
     // from bottom should always break follow unless a hold target is active.
-    if (scrolledUp && autoScroll() && !atBottom && heldItemCount() === null) {
+    if (wasPinnedAtBottom && scrolledUp && autoScroll() && !atBottom && heldItemCount() === null) {
       setAutoScroll(false)
+      lastObservedPinnedAtBottom = false
       return
     }
 
@@ -272,6 +275,8 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
         setAutoScroll(false)
       }
     }
+
+    lastObservedPinnedAtBottom = autoScroll() && atBottom
   }
 
   function scrollToBottom(immediate = true, options?: { suppressAutoAnchor?: boolean }) {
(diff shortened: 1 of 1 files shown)


shantur:
Hey @pascalandr 

Before I review, just to check if this fix works well with the hold the last assistant message feature implemented recently?

Thanks.

shantur:
I can't figure out how to reproduce the issue this PR is fixing in dev. Would you be able to let me know ?

Thanks

pascalandr:
@shantur 

two bugs :
1 - the first one when you scroll the text windows to the bottom by hand with the slider bar, you can't move up with the Page up hotkey as it comes back to bottomn immediatly, you need to scroll again a little by hand to make the page up key work again. Maybe this is a windows only thing ?

2 - I realized the “not fully aligned to the bottom” case I was looking at only happens when `hold` is active on a long last assistant reply.
At the moment, with `hold` enabled, the view can stay a bit above the true bottom even after streaming finishes, which can leave some of the remaining reply partially hidden behind the prompt composer.
Is that the intended behavior of `hold`?
Or should `hold` only pause bottom-pinning while the reply is actively growing, then release at the end so the final lines are fully visible above the prompt?

shantur:
> 1 - the first one when you scroll the text windows to the bottom by hand with the slider bar, you can't move up with the Page up hotkey as it comes back to bottomn immediatly, you need to scroll again a little by hand to make the page up key work again. Maybe this is a windows only thing ?

I tried to reproduce it but couldn't. It shouldn't be windows only thing unless its something with Tauri. Would you be able to confirm this on Electron too, so we know what to target?


> 2 - I realized the “not fully aligned to the bottom” case I was looking at only happens when `hold` is active on a long last assistant reply.
> At the moment, with `hold` enabled, the view can stay a bit above the true bottom even after streaming finishes, which can leave some of the remaining reply partially hidden behind the prompt composer.
> Is that the intended behavior of `hold`?
> Or should `hold` only pause bottom-pinning while the reply is actively growing, then release at the end so the final lines are fully visible above the prompt?

The 'hold' feature is implemented so that long assistant replies should stop scrolling when the top of the reply is about to go out of the view. This is to fox the behavior when every time user gets a long reply, they have to scroll to the top to be able to read it. When the hold is enabled and the assistant reply is growing more than the viewport, the autoscroll is held to keep it in view.

pascalandr:
You're right on the second point: that behavior is coming from the existing hold semantics, not a regression I should fix here. I trimmed the PR to keep only the sticky-bottom/PageUp fix and pushed a follow-up commit accordingly.

pascalandr:
> > 1 - the first one when you scroll the text windows to the bottom by hand with the slider bar, you can't move up with the Page up hotkey as it comes back to bottomn immediatly, you need to scroll again a little by hand to make the page up key work again. Maybe this is a windows only thing ?
> 
> I tried to reproduce it but couldn't. It shouldn't be windows only thing unless its something with Tauri. Would you be able to confirm this on Electron too, so we know what to target?
> 
happens also with electron and older releases, not with all cessions and not every time,.. sometime I have to do the steps several time to trigger it, So I couldn't find the exact trigger to reproduce it every time, but still happen often in most of my cessions, very annoying. Anyway if you can't reproduce it there's no harm with the fix, makes it more robust


> > 2 - I realized the “not fully aligned to the bottom” case I was looking at only happens when `hold` is active on a long last assistant reply.
> > At the moment, with `hold` enabled, the view can stay a bit above the true bottom even after streaming finishes, which can leave some of the remaining reply partially hidden behind the prompt composer.
> > Is that the intended behavior of `hold`?
> > Or should `hold` only pause bottom-pinning while the reply is actively growing, then release at the end so the final lines are fully visible above the prompt?
> 
> The 'hold' feature is implemented so that long assistant replies should stop scrolling when the top of the reply is about to go out of the view. This is to fox the behavior when every time user gets a long reply, they have to scroll to the top to be able to read it. When the hold is enabled and the assistant reply is growing more than the viewport, the autoscroll is held to keep it in view.

ok got it now, I trimmed the PR to keep only the sticky-bottom/PageUp fix.

shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

pascalandr:
Addressed the bot feedback by tightening the heuristic: follow now only breaks on an upward move that exits a previously bottom-pinned state, instead of any upward offset change. And fix still works.
Validation rerun: 
px tsc --noEmit --project packages/ui/tsconfig.json and 
pm run build --prefix packages/ui.

shantur:
@CodeNomadBot

Any good now?

PR #332, opened 2026-04-18
fix(desktop): source shell rc before launching CLI

### LINKED ISSUE #326: [Bug]: Cannot launch CodeNomad through linux app laucher
### App Variant

Tauri

Ubuntu 26.04, CodeNomad 0.14.0

### Issue Summary

Cannot launch CodeNomad through linux app laucher

### Steps to Reproduce

1. use Ubuntu app launcher (Dash / App launcher) to start CodeNomad
2. CodeNomad shows error: CLI exited early: exit status: 127

Image

It's ok to start codenomad in terminal : codenomad-tauri

### Expected Behavior

Should launch without error

### Logs & Screenshots

Image

### Additional Context

CodeNomad 0.14.0.

There was no such issue in 0.13.3

pascalandr:
Fixes #326

## Summary
- source the user's bash or zsh rc before launching the bundled CLI from Tauri
- use `-l -i -c` for zsh so shell-managed Node runtimes are available in launcher-started sessions
- fixes the reproduced Linux launcher case where the app exits with `CLI exited early: exit status: 127` while terminal launches work

## Validation
- reproduced the failure with the released Tauri `v0.14.0` Linux binary
- verified the patched binary succeeds under the same launcher-like environment
- ran `cargo build` on the dev-based PR branch

### DIFF
--- a/packages/electron-app/electron/main/user-shell.ts
+++ b/packages/electron-app/electron/main/user-shell.ts
@@ -20,24 +20,10 @@ function getDefaultShellPath(): string {
   return "/bin/bash"
 }
 
-function wrapCommandForShell(command: string, shellPath: string): string {
-  const shellName = path.basename(shellPath)
-
-  if (shellName.includes("bash")) {
-    return 'if [ -f ~/.bashrc ]; then source ~/.bashrc >/dev/null 2>&1; fi; ' + command
-  }
-
-  if (shellName.includes("zsh")) {
-    return 'if [ -f ~/.zshrc ]; then source ~/.zshrc >/dev/null 2>&1; fi; ' + command
-  }
-
-  return command
-}
-
 function buildShellArgs(shellPath: string): string[] {
   const shellName = path.basename(shellPath)
-  if (shellName.includes("zsh")) {
-    return ["-l", "-i", "-c"]
+  if (shellName.includes("zsh") || shellName.includes("bash")) {
+    return ["-i", "-l", "-c"]
   }
   return ["-l", "-c"]
 }
@@ -59,12 +45,11 @@ export function buildUserShellCommand(userCommand: string): ShellCommand {
   }
 
   const shellPath = getDefaultShellPath()
-  const script = wrapCommandForShell(userCommand, shellPath)
   const args = buildShellArgs(shellPath)
 
   return {
     command: shellPath,
-    args: [...args, script],
+    args: [...args, userCommand],
   }
 }
 
--- a/packages/tauri-app/src-tauri/src/cli_manager.rs
+++ b/packages/tauri-app/src-tauri/src/cli_manager.rs
@@ -1307,11 +1307,11 @@ fn build_shell_args(shell: &str, command: &str) -> Vec<String> {
         .unwrap_or("")
         .to_lowercase();
 
-    if shell_name.contains("zsh") {
-        return vec!["-l".into(), "-i".into(), "-c".into(), command.into()];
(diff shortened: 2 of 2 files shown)


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

pascalandr:
Addressed in `0c7c11c`:
- removed the explicit `source ~/.zshrc` path
- kept `zsh -l -i -c` so zsh manages its own startup files
- explicit rc sourcing remains bash-only

I also rebuilt the branch after the change.

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

1. 
Re
```
if shell_name.contains("bash") {
        return format!(
            "if [ -f ~/.bashrc ]; then source ~/.bashrc >/dev/null 2>&1; fi; {command}"
        );
    }
```
Doesn't bash always load ~/.bashrc, is this needed?

2. For current shell detection, would it be alright to use $SHELL variable?

shantur:
@CodeNomadBot

> 1. Bash does not always load `~/.bashrc` here

Do we need something similar for ~/.zshrc then?

shantur:
@CodeNomadBot

Does bash has similar to zsh -i ?

shantur:
@pascalandr  - Shall we keep it same for both zsh and bash by adding -i ?

pascalandr:
I'd keep bash as-is for now. `bash -i` would turn it into a fully interactive shell in a launcher/background process, which is broader and can bring extra prompt / job-control / shell-hook side effects. The current bash path is narrower: source `~/.bashrc`, then `exec`. So I don't think bash has to match zsh 1:1 here.

shantur:
@pascalandr 

> I'd keep bash as-is for now. `bash -i` would turn it into a fully interactive shell in a launcher/background process, which is broader and can bring extra prompt / job-control / shell-hook side effects. The current bash path is narrower: source `~/.bashrc`, then `exec`. So I don't think bash has to match zsh 1:1 here.

- Doesn't make sense to me, if zsh -i does same stuff and can break then why its ok to use zsh -i?
- We have been using zsh -i / bash -i for electron till now without any issues

shantur:
@CodeNomadBot

Verify my statement above

shantur:
@pascalandr - I think we can align all with zsh/bash -i -l -c to ensure we don't get weird behaviors when zsh/bash is used. It will be difficult to know. 
Can you please implement this for both packages/tauri-app and packages/electron-app

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
@pascalandr - Can it please be aligned.

pascalandr:
Aligned as requested in `92103bd`:
- Electron and Tauri now both use `-i -l -c` for bash and zsh
- removed the extra bash-specific manual `~/.bashrc` sourcing on the Tauri side, so bash no longer gets double-sourced there

I hit some workspace dependency churn while trying to rerun the local build in this worktree, but the code path is now aligned with the updated Electron behavior.

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
`wrap_command_for_shell()` was just a pass-through after the shell-alignment change, so I inlined it as asked.

pascalandr:
wrap_command_for_shell() was suppressed and inlined in uild_shell_command_string()
 
 @CodeNomadBot

verify my correction

shantur:
Thanks @pascalandr

Issue #330, opened 2026-04-17
[Bug] Ubuntu .deb shows a generic icon and the Linux app feels slower than Windows


WadohS:
## Environment
- OS: Ubuntu 24.04.4 LTS
- Install method: Linux .deb package
- Comparison platform: Windows desktop build

## What I noticed
I am using CodeNomad on both Windows and Linux.

On Ubuntu, I am seeing two Linux-specific issues:

1. The desktop/app launcher icon does not appear correctly after installing the .deb package.
   - I only see a generic gear icon instead of the CodeNomad icon.
   - This is not the most important issue, but it makes the Linux packaging/integration feel incomplete.

2. The Linux desktop app feels noticeably slower than the Windows version.
   - When I switch from one workspace to another, there is a visible latency before the UI catches up.
   - I also notice input lag while typing: characters do not always appear immediately and there is a small delay before text is rendered.
   - The same workflows feel more responsive on Windows.

## Reproduction
- Install CodeNomad from the Linux .deb
- Launch it from Ubuntu
- Open multiple workspaces
- Switch between workspaces several times
- Type in the chat/input area

## Expected behavior
- The proper CodeNomad icon should appear in the Ubuntu app launcher.
- Workspace switching and typing should feel as responsive as on Windows, or at least not show visible UI/input lag.

## Notes
I am not sure about the root cause.
It may be related to packaging for the icon issue, and possibly UI/main-thread responsiveness for the Linux latency issue, but that is only a guess from user experience.

If useful, I can help test a fix on Ubuntu.

pascalandr:
icon fix here https://github.com/NeuralNomadsAI/CodeNomad/issues/334

shantur:
Closed with #334

shantur:
Thanks for raising it @WadohS

pascalandr:
@Shantur I'm also working on the second half of the issue (Workspace switching and typing should feel as responsive as on Windows, or at least not show visible UI/input lag.) because my Tauri perf PR should have fixed that, well it does on windows, so looking at what is so specific to GTKweb.
https://github.com/NeuralNomadsAI/CodeNomad/pull/334 only fix the icon part of the issue.

WadohS:
Small follow-up from additional testing: on my Linux side, the icon still does not appear correctly in the Electron build either. So the launcher/app icon issue does not seem fully resolved for me yet, at least in that variant.

PR #334, opened 2026-04-18
fix(desktop): improve Linux desktop icon integration

pascalandr:
Refs #330

## Summary
- add standard Linux hicolor icon sizes to the Tauri package outputs
- enable the GTK app id on Linux and ship a matching reverse-DNS desktop entry alias for shell association
- mark the alias desktop entry `NoDisplay=true` so it does not surface as a duplicate launcher in desktop menus
- include the same alias desktop entry for AppImage so the fix is not limited to deb/rpm packages

## Validation
- confirmed in the Linux VM that the desktop-integrated launch no longer shows the generic taskbar icon
- verified the alias desktop entry is now hidden from app menus via `NoDisplay=true`
- attempted a fresh `tauri build --bundles deb`; the build still hits the known optional `@tauri-apps/cli` native-binding issue in this workspace after prebuild, not a code/config error from this PR


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

Does package/electron-app need the similar changes?

pascalandr:
Addressed CodeNomadBot's feedback in `e2b34bf`:
- the reverse-DNS desktop alias is now `NoDisplay=true`, so it should not show up as a second launcher in menus
- the same alias file is now included for AppImage too, so this is not just deb/rpm-specific anymore

shantur:
@pascalandr - Would you be able to align the packages/electron-app ids too as picked up the gatekeeper? If not that's fine, I will pick up in another PR.

pascalandr:
I'd rather keep this PR scoped to the Tauri/Linux fix, but yes, I can pick up the Electron identifier alignment in a small follow-up PR.

pascalandr:
Picked this up in a separate follow-up PR to keep `#334` scoped: #342

shantur:
@CodeNomadBot

How is it looking now?

PR #335, opened 2026-04-18
fix(desktop): support configured server ports

### LINKED ISSUE #293: [Help] Set server port in the actual macos Tauri app
Is there a way of setting a fixed port for the server hosted inside the Mac apps?

Ive tried:

export CLI_HTTPS_PORT="60599"
export CLI_HTTP_PORT="60599"

But clicking the app in the dock doesnt pick this up

pascalandr:
Fixes #293

## Summary
- read persisted `httpPort` and `httpsPort` values from the desktop config
- pass those fixed ports to the bundled server when launcher-started desktop sessions come up
- keep environment variables as the highest-priority override when they are explicitly set

## Validation
- `npm run bundle:server && cargo build`
- visual end-to-end validation of the configured-port flow is still pending

### DIFF
--- a/packages/electron-app/electron/main/cli-config.test.ts
+++ b/packages/electron-app/electron/main/cli-config.test.ts
@@ -1,0 +1,58 @@
+import assert from "node:assert/strict"
+import fs from "node:fs"
+import os from "node:os"
+import path from "node:path"
+import { describe, it } from "node:test"
+
+import { applyConfiguredPorts, readAppConfigFromPaths, resolveConfiguredPortsFromConfig } from "./cli-config"
+
+describe("resolveConfiguredPortsFromConfig", () => {
+  it("prefers server port values over preferences", () => {
+    const ports = resolveConfiguredPortsFromConfig({
+      preferences: {
+        httpPort: 3000,
+        httpsPort: 3443,
+      },
+      server: {
+        httpPort: 4000,
+        httpsPort: 4443,
+      },
+    })
+
+    assert.deepEqual(ports, [4443, 4000])
+  })
+})
+
+describe("applyConfiguredPorts", () => {
+  it("keeps env vars as the highest-priority override", () => {
+    const args = ["serve"]
+
+    applyConfiguredPorts(args, {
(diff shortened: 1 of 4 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
Follow-up for the bot review, pushed in `3b3ac1d`:
- added focused Rust tests for the precedence path
- covered `server > preferences` port selection
- covered `env var > persisted config` override behavior
- covered YAML config loading for the desktop wrapper

Validation:
- `npm run bundle:server && cargo test` on the PR branch

shantur:
@pascalandr  -

- What updates the config with the server port?
- What happens if there is another process listening on the port?
- Is there any UI component to update the port?
- Electron would need to update to handle this feature too.

pascalandr:
Right, I only looked at the tip of the iceberg. Converting this to draft to rework it properly.

pascalandr:
@shantur 
> @pascalandr -
> 
>     * What updates the config with the server port?
Nothing in this PR. It only honors already-persisted `httpPort` / `httpsPort` values.
>     * What happens if there is another process listening on the port?
 Startup fails, same as any explicit fixed-port bind.
>     * Is there any UI component to update the port?
 Not in this PR. Follow-up : https://github.com/Pagecran/CodeNomad/pull/9
>     * Electron would need to update to handle this feature too.
I’ve now added the same behavior to Electron as well.

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
Hey @pascalandr 

thanks for working on this.

I would rather have it as complete feature as this can be a breaking change.
this needs to be thought through for edge cases as if user sets a wrong port in settings in UI and that port is taken already, then there is no way for user to change in the UI again as the UI server won't startup due to clashing port.

I understand user can go and manually change in the config file but that's not a good UX.

pascalandr:
Closing this for now since honoring persisted fixed ports without a safe recovery path can leave the desktop app unable to start if the configured port is invalid or already in use.

PR #338, opened 2026-04-18
feat(server): expose OpenCode instance URL in env

### LINKED ISSUE #321: Feature request: Add opencode server URL as environment variable
It would be handy to have the OpenCode server URL easily available in the OpenCode environment. For me personally I'm making API calls from tools and plugins to the OpenCode server to accomplish things that aren't currently supported by their plugin architecture. Being able to identify the "correct" OpenCode server to talk to is important for this use case.

✔️  You already provide `OPENCODE_SERVER_USERNAME` and `OPENCODE_SERVER_PASSWORD` in the environment.
❌  The OpenCode server port is not available in the environment.

Request: Add something like `OPENCODE_SERVER_BASE_URL` with the full URL and/or `OPENCODE_SERVER_PORT` with the port number in the environment.

My current workaround is to use `CODENOMAD_INSTANCE_ID` + `CODENOMAD_BASE_URL` to look this up in CodeNomad's API. However this is an extra API call that might require auth, etc.

pascalandr:
Fixes #321

## Summary
- reserve a loopback port before launching each OpenCode workspace runtime
- inject `OPENCODE_SERVER_PORT` and `OPENCODE_SERVER_BASE_URL` into the spawned instance environment
- keep the generated base URL aligned with the loopback host CodeNomad already uses for readiness and proxying

## Validation
- `npm run build --workspace @neuralnomads/codenomad`

### DIFF
--- a/packages/server/src/workspaces/runtime.ts
+++ b/packages/server/src/workspaces/runtime.ts
@@ -1,5 +1,6 @@
 import { ChildProcess, spawn, spawnSync } from "child_process"
 import { existsSync, statSync } from "fs"
+import { createServer } from "net"
 import path from "path"
 import { EventBus } from "../events/bus"
 import { LogLevel, WorkspaceLogEntry } from "../api-types"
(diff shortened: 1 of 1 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
CodeNomadBot is right here: this introduces a real port-selection race, so I'm going to close this PR rather than force a fragile implementation for `OPENCODE_SERVER_PORT` / `OPENCODE_SERVER_BASE_URL`.

PR #342, opened 2026-04-18
fix(desktop): align Electron package and runtime app ids

pascalandr:
Follow-up from #334

## Summary
- align the Electron package `build.appId` with the runtime identifier already used in `app.setAppUserModelId(...)`
- remove the mismatch between packaged desktop identity and runtime desktop identity
- keep the change narrowly scoped to identifier consistency only

## Validation
- verified the previous mismatch in `packages/electron-app/package.json` vs `packages/electron-app/electron/main/main.ts`
- updated the packaging id to match the runtime id exactly

### DIFF
--- a/packages/electron-app/package.json
+++ b/packages/electron-app/package.json
@@ -62,7 +62,7 @@
     "vite-plugin-solid": "^2.10.0"
   },
   "build": {
-    "appId": "ai.opencode.client",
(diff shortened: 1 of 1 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

PR #346, opened 2026-04-20
Don't depend on Node anymore

shantur:
## Summary
- package `packages/server` as a standalone desktop executable so Electron and Tauri no longer depend on a system-installed Node runtime in production
- align Electron and Tauri startup logic around launching the packaged server, resolving binaries from the user shell, and bundling the same server resources into both desktop apps
- replace the workspace instance proxy path that used `@fastify/reply-from` with a direct streaming proxy so packaged standalone builds can talk to spawned `opencode` instances correctly

## Why
Desktop production builds were still depending on a user-provided Node runtime to launch `packages/server`, which made packaging less self-contained and created different behavior across machines. While moving to a standalone server executable, we also found that Bun-compiled standalone builds could start `opencode` successfully but failed when proxying requests to those instances through `reply-from`.

The goal of this change is to make desktop production startup self-contained, keep Electron and Tauri behavior aligned, and restore correct communication with local `opencode` instances in packaged builds.

## What Changed
- added a standalone build path for `packages/server` and bundle `codenomad-server` into desktop resources
- updated Electron production startup to resolve and launch the standalone server executable
- updated Tauri production startup to resolve and launch the standalone server executable with matching cwd and shell behavior
- added runtime path helpers so the packaged server can reliably find its bundled UI, auth templates, config template, and package metadata
- improved bare binary resolution so commands like `opencode` can be resolved from the user's login shell environment
- upgraded the server stack to newer Fastify-compatible packages needed for the standalone/runtime work
- replaced the workspace instance proxy implementation with a direct streaming proxy for requests to spawned `opencode` instances
- updated Electron and Tauri build/prebuild scripts to generate and package the standalone server, while also repairing missing platform-specific optional binaries during packaging

## Benefits
- desktop production builds no longer require Node to be installed on the user's system
- Electron and Tauri now use the same packaged server model in production, reducing platform drift
- packaged desktop apps can successfully create workspaces, launch `opencode`, and proxy health/session traffic to those instances
- the server bundle is more self-contained and resilient to different launch environments
- desktop packaging is more predictable because the required server executable is built and bundled as part of the app build flow

### DIFF
--- a/.github/workflows/build-and-upload.yml
+++ b/.github/workflows/build-and-upload.yml
@@ -53,7 +53,7 @@ on:
 # least-privilege (e.g. dev CI uses read-only; releases grant write).
 
 env:
-  NODE_VERSION: 20
+  NODE_VERSION: 22
 
 jobs:
   build-macos:
--- a/.github/workflows/build-and-upload.yml
+++ b/.github/workflows/build-and-upload.yml
@@ -643,7 +643,8 @@ jobs:
             if [ "$attempt" -gt 1 ]; then
               echo "Retrying Tauri CLI install (attempt $attempt)..."
             fi
-            npm install @tauri-apps/cli@2.10.1 @tauri-apps/cli-linux-x64-gnu@2.10.1 --no-save --no-audit --no-fund --workspaces=false
+            # Tauri CLI 2.10.1 regresses Linux AppImage bundling in CI; keep Linux on the last known-good CLI.
+            npm install @tauri-apps/cli@2.9.4 @tauri-apps/cli-linux-x64-gnu@2.9.4 --no-save --no-audit --no-fund --workspaces=false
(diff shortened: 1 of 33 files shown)


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

How does it look now?

shantur:
@CodeNomadBot

What happens if there is no opencode binary present?

shantur:
@CodeNomadBot

1. How is shell environment setup for user for launching opencode. Tell me what happens in each platform and with different shells bash / zsh

shantur:
@CodeNomadBot

How does it look now?

shantur:
@CodeNomadBot

How is it now?

PR #375, opened 2026-04-27
fix(ui): stabilize streaming chat scroll follow

VooDisss:
## Summary

- Fix sticky chat auto-scroll during streaming by making follow exit/re-entry more reliable in `VirtualFollowList`.
- Prevent visible snapbacks by using DOM scroll state as the source of truth for immediate/manual scroll behavior.
- Keep the upstream surface minimal: this change is scoped to `packages/ui/src/components/virtual-follow-list.tsx` only.

## Testing

- Manually verified:
  - single-notch upward exit from auto-scroll
  - downward re-entry near bottom
  - normal browsing while escaped
  - no visible snapback during rapid streamed output

## Review Notes

- Maintainers may edit or adjust this PR as needed.
- I will ignore gatekeeper bot reviews for this PR.

### DIFF
--- a/packages/ui/src/components/virtual-follow-list.tsx
+++ b/packages/ui/src/components/virtual-follow-list.tsx
@@ -2,6 +2,7 @@ import { Show, createEffect, createMemo, createSignal, type Accessor, type JSX, 
 import { Virtualizer, type VirtualizerHandle } from "virtua/solid"
 
 const DEFAULT_SCROLL_SENTINEL_MARGIN_PX = 48
+const DEFAULT_HOLD_TARGET_TOP_THRESHOLD_PX = 8
 const DEFAULT_NEAR_BOTTOM_THRESHOLD_PX = 48
 const SCROLL_DIRECTION_KEYS = new Set(["ArrowUp", "ArrowDown", "PageUp", "PageDown", "Home", "End", " ", "Spacebar"])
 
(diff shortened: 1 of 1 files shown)


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
@VooDisss
I’m seeing reproducible regressions in the latest PR state during manual testing:
- the overlay up/down scroll buttons can desync the timeline highlight from the visible message
- auto-scroll can sometimes stop short of the actual bottom, as if hold/follow state was not fully released

I’m still narrowing down ...

pascalandr:
So I found multiple reproducible regressions in the latest PR state:

1. Overlay up/down scroll buttons can desync the timeline from the visible message.
`message-section.tsx` drives the overlay arrows through `api.scrollToTop()` / `api.scrollToBottom()`, and both paths set `ignoreNextScrollEvent` in `virtual-follow-list.tsx`. The next `handleScroll()` then returns early after updating scroll buttons only, without recomputing `activeKey` / `onActiveKeyChange`. Since the timeline highlight is driven from `onActiveKeyChange`, the viewport can move while the timeline stays on the previous segment.

I reproduced this in both Tauri and Electron builds with both overlay arrows.

2. Explicit “go to latest” actions do not reliably stay at the actual bottom.
`scrollToBottom()` can immediately be followed by hold logic on the next render via `notifyContentRendered()` -> `updateAutoPinHold()`. Since `holdLongAssistantReplies` is enabled by default, an explicit bottom action (for example the bottom arrow, or `End`) can go to the bottom and then jump slightly back up instead of staying on the latest content.

This is not just a UX preference issue: explicit “go to latest” should not behave as if hold mode silently overrode it.

3. Scrolling during auto-follow can leave streaming effectively stuck.
A single upward wheel event immediately calls `escapeFromFollow()`, which cancels pending follow work, clears hold, sets `escapedFromLock = true`, and disables `autoScroll`. After that, content growth no longer follows until the component explicitly rejoins follow again. In practice this makes streaming feel broken after even a small manual scroll, and the re-entry behavior is inconsistent across wheel, keyboard, and programmatic navigation.

4. Keyboard downward navigation is not handled consistently with wheel downward navigation.
The code computes `ArrowDown` / `PageDown` / `End` as downward intent, but unlike the wheel path it does not use that to rejoin follow or perform the same near-bottom handling. That makes the keyboard behavior inconsistent and is likely part of the “End goes to bottom and then behaves oddly” problem.

@VooDisss  do you see the same problemes ?

pascalandr:
seperseded by #406

Issue #384, opened 2026-05-01
[Bug] Fail to Load Agent Properly When Opening


jollyxenon:
## App Variant

Electron

## Operating System & Version

Windows 11 24H2

## Issue Summary

CodeNomad launches its own OpenCode server with a CodeNomad-specific `OPENCODE_CONFIG_DIR`, which prevents user-configured OpenCode plugins and agents from being loaded.

As a result, agents provided by user plugins such as `oh-my-openagent` are missing from the bottom-left agent selector. After merging the user OpenCode config, another issue appears: native agents such as `build` can be auto-selected and temporarily shown in the primary agent selector, even though they should not be selectable as primary agents.

## Steps to Reproduce

Create a new instance and then enter a conversation.

## Current Behavior

1. CodeNomad starts OpenCode with its packaged config directory.
2. User OpenCode config from `~/.config/opencode` is not loaded.
3. Plugin-provided primary agents such as `Sisyphus` and `Hephaestus` do not appear in the CodeNomad agent selector.
4. When plugin agents are available, the selector may still auto-load `build`.
5. `build` can appear in the selector because the current session agent is forced back into the list.

## Expected Behavior

1. CodeNomad should preserve its own bridge plugin while also loading the user's OpenCode config.
2. User-configured OpenCode plugins should be available in CodeNomad-launched OpenCode instances.
3. The primary agent selector should only include selectable primary agents.
4. Native/internal agents such as `build` should not be auto-selected or shown in the primary selector.
5. New sessions should default to a valid selectable primary agent.

## Logs & Screenshots

Though we expect to see "Sisyphus" and other agents here, it only loaded default agents.
<img width="307" height="172" alt="Image" src="https://github.com/user-attachments/assets/259d9e97-b616-42fa-9c18-c9c92ff0b905" />

## Root Cause

There are two related issues:

1. CodeNomad sets `OPENCODE_CONFIG_DIR` to its own bundled config directory, so user OpenCode plugins are isolated from the OpenCode process launched by CodeNomad.
2. The UI treats any non-subagent agent as selectable. This allows native/internal agents to be used as default primary agents. The selector also force-adds the current agent back into the list, which can make invalid agents such as build appear temporarily.

hugalafutro:
i'm running into the same issue and it honestly sends me crazy

everything was working until about 30minutes ago - orchestrator, council agents from omo-slim in codenomad no problem.
[...]

PR #385, opened 2026-05-01
[Bug] Fix "Fail to Load Agent Properly When Opening"

jollyxenon:
## Summary

- Merge the user's OpenCode config into CodeNomad-launched OpenCode instances while preserving CodeNomad's bridge plugin.
- Stabilize agent fetching when plugins are configured.
- Restrict the bottom-left primary agent selector to valid selectable primary agents only.

## Problem

CodeNomad launches OpenCode with its own `OPENCODE_CONFIG_DIR`, so user-configured OpenCode plugins are not loaded. This makes plugin-provided primary agents, such as Sisyphus and Hephaestus from `oh-my-openagent`, unavailable in CodeNomad even though they work in OpenCode TUI.

After user agents are available, native/internal agents such as `build` can still be auto-selected or temporarily shown in the primary agent selector because the UI treats any non-subagent as selectable.

## Changes

- Generate a merged OpenCode config directory for workspaces:
  - Start from CodeNomad's packaged config so the bridge plugin remains available.
  - Overlay the user's `~/.config/opencode` config.
  - Use the merged directory as `OPENCODE_CONFIG_DIR`.
- Retry `/agent` briefly when plugins are configured so plugin agents have time to register.
- Preserve the SDK `native` field in the UI `Agent` type.
- Add `isSelectablePrimaryAgent(agent)`.
- Use the shared selectable-primary predicate for:
  - New session default agent selection.
  - Bottom-left parent-session agent selector options.
- Stop force-inserting invalid current agents such as `build` into the primary selector list.

## Validation

close issue #384

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -322,6 +322,7 @@ async function main() {
   const binaryResolver = new BinaryResolver(settings)
   const workspaceManager = new WorkspaceManager({
     rootDir: options.rootDir,
+    configDir,
     settings,
     binaryResolver,
     eventBus,
--- a/packages/server/src/opencode-config.ts
+++ b/packages/server/src/opencode-config.ts
@@ -1,4 +1,5 @@
-import { existsSync } from "fs"
+import { cpSync, existsSync, mkdirSync, rmSync } from "fs"
+import os from "os"
 import path from "path"
 import { fileURLToPath } from "url"
 import { createLogger } from "./logger"
@@ -17,15 +18,37 @@ const isDevBuild = Boolean(process.env.CODENOMAD_DEV ?? process.env.CLI_UI_DEV_S
 const templateDir = isDevBuild
   ? devTemplateDir
   : prodTemplateDirs.find((dir) => existsSync(dir)) ?? prodTemplateDirs[0]
+const userConfigDir = path.join(os.homedir(), ".config", "opencode")
+const mergedConfigDir = path.join(os.homedir(), ".config", "codenomad", "opencode-config")
+
+function copyConfigDirectory(source: string, target: string): void {
+  cpSync(source, target, {
+    recursive: true,
+    force: true,
+    dereference: true,
+  })
+}
+
+function prepareMergedConfigDir(): string {
+  rmSync(mergedConfigDir, { recursive: true, force: true })
+  mkdirSync(mergedConfigDir, { recursive: true })
+
+  copyConfigDirectory(templateDir, mergedConfigDir)
+
+  if (existsSync(userConfigDir)) {
+    copyConfigDirectory(userConfigDir, mergedConfigDir)
+    log.debug({ templateDir, userConfigDir, mergedConfigDir }, "Using merged OpenCode config")
+  } else {
+    log.debug({ templateDir, mergedConfigDir }, "Using generated OpenCode config without user overlay")
+  }
+
+  return mergedConfigDir
+}
 
 export function getOpencodeConfigDir(): string {
   if (!existsSync(templateDir)) {
     throw new Error(`CodeNomad Opencode config template missing at ${templateDir}`)
   }
 
-  if (isDevBuild) {
-    log.debug({ templateDir }, "Using Opencode config template directly (dev mode)")
-  }
-
-  return templateDir
+  return prepareMergedConfigDir()
 }
--- a/packages/server/src/opencode-config.ts
+++ b/packages/server/src/opencode-config.ts
@@ -19,7 +19,6 @@ const templateDir = isDevBuild
   ? devTemplateDir
   : prodTemplateDirs.find((dir) => existsSync(dir)) ?? prodTemplateDirs[0]
 const userConfigDir = path.join(os.homedir(), ".config", "opencode")
-const mergedConfigDir = path.join(os.homedir(), ".config", "codenomad", "opencode-config")
 
 function copyConfigDirectory(source: string, target: string): void {
   cpSync(source, target, {
@@ -29,7 +28,16 @@ function copyConfigDirectory(source: string, target: string): void {
   })
 }
 
-function prepareMergedConfigDir(): string {
+function copyBridgePlugin(target: string): void {
+  const bridgePluginDir = path.join(templateDir, "plugin")
+  if (existsSync(bridgePluginDir)) {
+    copyConfigDirectory(bridgePluginDir, path.join(target, "plugin"))
+  }
+}
+
+function prepareMergedConfigDir(baseDir: string): string {
+  const mergedConfigDir = path.join(baseDir, "opencode-config")
+
   rmSync(mergedConfigDir, { recursive: true, force: true })
   mkdirSync(mergedConfigDir, { recursive: true })
 
@@ -37,6 +45,7 @@ function prepareMergedConfigDir(): string {
 
   if (existsSync(userConfigDir)) {
     copyConfigDirectory(userConfigDir, mergedConfigDir)
+    copyBridgePlugin(mergedConfigDir)
     log.debug({ templateDir, userConfigDir, mergedConfigDir }, "Using merged OpenCode config")
   } else {
     log.debug({ templateDir, mergedConfigDir }, "Using generated OpenCode config without user overlay")
@@ -45,10 +54,10 @@ function prepareMergedConfigDir(): string {
   return mergedConfigDir
 }
 
-export function getOpencodeConfigDir(): string {
+export function getOpencodeConfigDir(baseDir: string): string {
   if (!existsSync(templateDir)) {
     throw new Error(`CodeNomad Opencode config template missing at ${templateDir}`)
   }
 
-  return prepareMergedConfigDir()
+  return prepareMergedConfigDir(baseDir)
 }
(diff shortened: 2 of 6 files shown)


jollyxenon:
Addressed Copilot's review comments in the latest push:

- Removed the unused `For` import from `agent-selector.tsx`.
- Updated the merged OpenCode config path to be derived from CodeNomad's configured `configDir` instead of hard-coding `~/.config/codenomad`.
- Re-applied CodeNomad's bundled `plugin/` bridge after overlaying the user's OpenCode config, so user files cannot accidentally override the required bridge plugin.

Validation:

shantur:
@jollyxenon  - Thanks for the PR and using CodeNomad

I want to understand how you are configuring your plugins as CodeNomad keeps `~/.config/opencode`. 
OpenCode is supposed to load `~/.config/opencode` with CONFIG_DIR too. 
Check the docs here - https://opencode.ai/docs/config/#locations

jollyxenon:
Thanks for pointing this out — I may have misunderstood the intended `OPENCODE_CONFIG_DIR` behavior.

My local setup is the standard OpenCode config location on Windows:

- `C:\Users\<user>\.config\opencode`
- plugins/agents are provided there by `oh-my-openagent`
- OpenCode TUI loads those agents correctly

When launched through CodeNomad, I observed that the plugin-provided agents were either missing initially or not equipped correctly, while internal agents always appeared in the primary agent selector.

I interpreted that as CodeNomad’s custom `OPENCODE_CONFIG_DIR` isolating OpenCode from the global config, so I added the merged config-dir approach. But based on the docs you quoted, OpenCode should already load global config plus `OPENCODE_CONFIG_DIR`.

So I think there are two possibilities:

1. The server-side config merge is unnecessary and we should remove that part from this PR.
2. There is a Windows/runtime-specific issue where the global config is not being picked up in the CodeNomad-launched OpenCode process, despite the documented behavior.

The UI-side part of this PR still addresses a separate issue I could reproduce: the primary agent selector should use the same visibility/selectability rules as OpenCode TUI, so native/internal agents like `build` do not get auto-selected or shown when they should be hidden.

I'm happy to simplify the PR if you prefer.

shantur:
@jollyxenon - when last I check the issue was that oMo didn't handle password protected OpenCode instances.
Try starting opencode with `OPENCODE_SERVER_PASSWORD=12345678 opencode serve` and then see if you are able to use oMo in opencode web

jollyxenon:
Thanks — I tested the password-protected case and I don’t think that is the root cause here.

In my setup:

- `OPENCODE_SERVER_PASSWORD` enabled: oMo still works from OpenCode Web
- TUI also works correctly
- On the latest CodeNomad build, the agent selector now shows only `build`, `plan`, and `Atlas` (from OMO)

So this no longer looks like a general “password-protected OpenCode instance” issue to me.

What I’m still seeing is a CodeNomad-specific mismatch with OpenCode TUI/web behavior:
- plugin agents are not fully surfaced the same way in CodeNomad
- the visible/selectable primary-agent set differs from what OpenCode itself exposes

Since `Atlas` does appear, this also suggests the plugin is not completely failing to load. The issue seems more likely to be in how CodeNomad fetches, filters, or stabilizes the agent list after startup.

I can narrow the PR accordingly if that’s preferable:
- keep only the CodeNomad-side agent visibility/filtering fix
- re-evaluate or remove the config-merging part unless we can prove CodeNomad is actually missing part of the OpenCode config at runtime

jollyxenon:
I ran a GUI-side comparison to isolate this.

What I observed:

1. OpenCode Web with only my normal global config (`~/.config/opencode`) shows 4 OMO agents.
2. OpenCode Web with `OPENCODE_CONFIG_DIR` pointed at CodeNomad’s packaged config drops to only `Build`, `Plan`, and `Atlas`.
3. Current CodeNomad `dev` shows the same reduced set: `Build`, `Plan`, and `Atlas`.
4. My PR build restores the expected 4 OMO agents.

So in my environment, the effective runtime behavior is:

- global config alone -> full OMO agent set
- custom `OPENCODE_CONFIG_DIR` -> reduced agent set
- CodeNomad current dev -> same reduced set
- PR build -> full set restored

That strongly suggests the config-merging part of the PR is addressing a real CodeNomad-side issue in practice, even if the documented precedence implies global config should still be loaded.

Also, since `Atlas` still appears in the reduced case, this does not look like a complete plugin-load failure. It looks more like only part of the OMO-provided configuration/agent set is surviving when CodeNomad launches OpenCode with its packaged config directory.

So at this point I believe the config merge should stay in the PR.

jollyxenon:
I dug into this further and ran a few isolated checks.

I think your point about OpenCode’s config loading is correct at the top-level config layer: `~/.config/opencode/opencode.json` is still merged when `OPENCODE_CONFIG_DIR` is set. I confirmed this with `opencode debug config`: the resolved `plugin` list still includes `oh-my-openagent@latest` alongside CodeNomad’s config-dir plugin.

The issue seems to be one layer lower, inside `oh-my-openagent`’s own config discovery.

OMO appears to resolve its user config from `OPENCODE_CONFIG_DIR` when that env var is present. In CodeNomad, that points to CodeNomad’s packaged config dir, which contains the bridge plugin but **does not contain my `oh-my-openagent.jsonc`**. As a result, OMO still loads, but without my user OMO config/model mappings, so the agent set is reduced.

That matches my GUI checks:

- OpenCode Web without `OPENCODE_CONFIG_DIR`: shows the expected 4 OMO agents
- OpenCode Web with `OPENCODE_CONFIG_DIR` pointing at CodeNomad’s packaged config: only `Build`, `Plan`, and `Atlas`
- current CodeNomad `dev`: same reduced set, `Build`, `Plan`, and `Atlas`
- this PR build: restores the expected 4 OMO agents

So I agree OpenCode’s top-level config merge works as documented, but CodeNomad’s custom config dir still changes the runtime config root seen by OMO. That is why the config overlay in this PR seems necessary in practice: it makes the user’s OMO config available from the same config dir that OMO resolves under CodeNomad.

shantur:
> * OpenCode Web with `OPENCODE_CONFIG_DIR` pointing at CodeNomad’s packaged config: only `Build`, `Plan`, and `Atlas`

That's interesting. This changes the focus completely to the CodeNomad plugin. What in the plugin is causing this issue. I think the root cause is somewhere there

jollyxenon:
I dug into the plugin specifically.

What the origin CodeNomad logic does is:

1. With `OPENCODE_CONFIG_DIR` redirecting to `this.opencodeConfigDir`, CodeNomad opens an OpenCode instance (which can be seen at `packages/server/src/workspaces/manager.ts:136-144`).
2. This so-called `this.opencodeConfigDir` is `packages/opencode-config` in dev (seen at `packages/server/src/opencode-config.ts:21-30`).

In `packages/opencode-config`, there is only a plugin used to bridge OpenCode and CodeNomad. And as OpenCode documents mentioned, all settings are merged rather than overwritten, so every global plugin can still be loaded.

However, the logic fails when encountering oh-my-openagent plugin. When not applying `OPENCODE_CONFIG_DIR`, OMO will use global config (`~/.config/opencode/oh-my-openagent.jsonc`); But when applying `OPENCODE_CONFIG_DIR`, OMO only uses private config and **WILL NOT USE GLOBAL CONFIG even if THERE IS NO OMO CONFIG IN PRIVATE CONFIG DIRECTORY**. It means OMO doesn't have a fallback mechanism to ensure the agent configuration is always found.

To solve the problem, I used to put forward this PR, which copies the whole global config to `packages/opencode-config` to ensure the OMO can find its `oh-my-openagent.jsonc`. But maybe we shouldn't completely blame this problem on CodeNomad. CodeNomad is the trigger, but the underlying incompatibility is in OMO’s config discovery. OpenCode treats `OPENCODE_CONFIG_DIR` as an additive custom directory, while OMO treats it as an exclusive user-config root and does not fall back to `~/.config/opencode`.

Maybe the PR should be closed and I ought to turn to oh-my-openagent project and propose there a issue. Ahhhh... I feel exhausted about these days' wrestles.

shantur:
Thanks @jollyxenon for the detailed analysis.

I agree with you that the issue is with OMO's handling. It seems to me it is hardcoding the config location there.
Thanks for your hard work but this needs to be fixed in OMO.

Feel free to re-open this if you feel otherwise.

jollyxenon:
@shantur 

I opened an upstream OMO issue for this [here](https://github.com/code-yeongyu/oh-my-openagent/issues/3846).

I'm really sorry, I immediately changed my mind... XD
The reason is that I think there is still something worth preserving in this PR, namely the agent visibility fix on the UI side.

I believe the agent visibility fix on the UI side should be kept, because what it fixes is not OMO's configuration discovery issue, but **CodeNomad's own display rules**.

The specific reasons are as follows:

1. It consolidates the UI rules into a single entry point, reducing the chance of similar bugs in the future
   - Currently, the main branch hardcodes "what can be selected in the primary session" within the component; the PR changes it to reuse `isSelectablePrimaryAgent`, which essentially centralizes this business rule.
   - Even if this fix is completely decoupled from the OMO issue, it is still worth keeping because it improves the consistency and maintainability of UI behavior.
2. After the modification, the behavior of `availableAgents` is fully aligned with OpenCode
   - The logic before the modification: even if the current agent should not appear according to the rules, it might still be displayed in the dropdown list
   - The logic after the modification: if it does not meet the rules, it directly disappears from the candidate list and triggers the subsequent auto-switching logic
   - In OpenCode's logic, OpenCode does not forcibly put a currently unselectable agent back into the list for display.

In summary, I believe this PR is not entirely unusable. ( •̀ ω •́ )✧

PR #386, opened 2026-05-01
fix(ui): scope session list by workspace folder

jollyxenon:
## Summary

- Scope session listing requests to the current workspace folder by passing `directory: instance.folder` to OpenCode.
- Prevent non-git workspace sessions from leaking across projects when OpenCode stores them under the shared `global` project.
- Include the scoped directory in session list logs for easier debugging.

## Why

OpenCode groups non-git folders under the shared `global` project. CodeNomad previously called `session.list()` without a directory filter, so fresh non-git folders could show sessions from unrelated folders.

## Testing

- `npm ci`

### DIFF
--- a/packages/ui/src/stores/session-api.ts
+++ b/packages/ui/src/stores/session-api.ts
@@ -123,8 +123,8 @@ async function fetchSessions(instanceId: string): Promise<void> {
   })
 
   try {
-    log.info("session.list", { instanceId })
-    const response = await rootClient.session.list()
+    log.info("session.list", { instanceId, directory: instance.folder })
+    const response = await rootClient.session.list({ directory: instance.folder })
 
     const sessionMap = new Map<string, Session>()
 
--- a/packages/ui/src/stores/session-api.ts
+++ b/packages/ui/src/stores/session-api.ts
@@ -123,8 +123,14 @@ async function fetchSessions(instanceId: string): Promise<void> {
   })
 
   try {
-    log.info("session.list", { instanceId, directory: instance.folder })
-    const response = await rootClient.session.list({ directory: instance.folder })
+    const projectResponse = await rootClient.project.current()
+    const projectId = projectResponse.data?.id
+    const sessionListOptions = projectId === "global" ? { directory: instance.folder } : undefined
+
+    log.info("session.list", { instanceId, projectId, directory: sessionListOptions?.directory })
+    const response = sessionListOptions
+      ? await rootClient.session.list(sessionListOptions)
+      : await rootClient.session.list()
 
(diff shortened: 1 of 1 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
@jollyxenon - This is a good catch by CodeNomadBot.

Instead of always, directory should be set only when the project id is "global"

pascalandr:
@jollyxenon
Maybe you could make something like this to adress the blocker:
- non-git workspace: `rootClient.session.list({ directory: instance.folder })`
- git workspace: keep the existing unfiltered `rootClient.session.list()`

That should fix the non-git leakage without regressing the multi-worktree model.

shantur:
> @jollyxenon Maybe you could make something like this to adress the blocker:
> 
> * non-git workspace: `rootClient.session.list({ directory: instance.folder })`
> * git workspace: keep the existing unfiltered `rootClient.session.list()`
> 
> That should fix the non-git leakage without regressing the multi-worktree model.

Opencode has a hardcoded project id for directories that aren't git = 'global' so we don't need to check if the workspace is git/non-git opencode is doing it for us

pascalandr:
@

> > @jollyxenon Maybe you could make something like this to adress the blocker:
> > 
> > * non-git workspace: `rootClient.session.list({ directory: instance.folder })`
> > * git workspace: keep the existing unfiltered `rootClient.session.list()`
> > 
> > That should fix the non-git leakage without regressing the multi-worktree model.
> 
> Opencode has a hardcoded project id for directories that aren't git = 'global' so we don't need to check if the workspace is git/non-git opencode is doing it for us

So you mean we should only apply `session.list({ directory: instance.folder })` when the current OpenCode project ID is `global`, and otherwise keep the existing unfiltered list to avoid regressing git/worktree sessions, right?

shantur:
That's correct

pascalandr:
@CodeNomadBot

Make the same review after this change

jollyxenon:
Thanks, updated in `56f98b1`.

The session list now checks `rootClient.project.current()` first and only passes `directory: instance.folder` when the OpenCode project id is `global`.

For non-global projects it keeps the existing unfiltered `rootClient.session.list()` behavior, so git/worktree session listing should not be narrowed to the root folder anymore.

Verified with:

- `git diff --check origin/fix/issue-355-scope-sessions..HEAD`

shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur (approved):

shantur:
Thanks @jollyxenon

PR #395, opened 2026-05-07
fix(ui): keep idle badges visible until viewed

### LINKED ISSUE #376: [UX/UI] Add a blue tab indicator for finished but unseen jobs
## Summary
It would be very helpful to add a distinct blue indicator on workspace tabs when a job has finished but the tab has not been opened/viewed yet.

## Current behavior
From a user perspective, it is hard to distinguish between:
- a workspace/job that is still running
- a workspace that is simply inactive
- a workspace where the job has already finished, but I have not looked at the result yet

Right now the inactive state does not stand out enough for this use case.

## Proposed behavior
Use a blue dot/badge on the tab when:
- a job/request has finished
- the corresponding tab is not currently open/active
- the result has not been viewed yet

This would make it much easier to quickly spot the projects where work is complete and waiting for attention.

## Why this would help
When working across many workspaces/projects, I often launch requests and continue elsewhere.
Later, I want to quickly identify:
- which tabs are still busy
- which tabs are idle
- which tabs are done, but still unseen

A dedicated blue state for "finished but unseen" would be much clearer than reusing the same visual treatment as a generic inactive tab.

## Suggested state meaning
- current/running: existing in-progress indicator
- inactive: existing green/idle state
- finished but unseen: blue indicator

## Notes
This is mainly a usability improvement for users who juggle multiple workspaces and want to review completed results faster.

pascalandr:
## Summary
- Implements the requested three-state session tab behavior without introducing a new blue indicator.
- Uses the existing yellow/working indicator while a session is actively running, the existing green/idle indicator when work finished in a non-viewed session, and no indicator once that completed session has been viewed.
- Keeps the green completed-but-unseen indicator visible for 2.5 seconds after the session becomes active, preserving the previous transient feedback instead of clearing it instantly.

Fixes #376

## Verification


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
Applied the non-blocking cleanup from the review: removed the unused idle-status time parameter and renamed the person-specific test.

shantur:
@CodeNomadBot

Review again with latest changes

shantur:
@pascalandr - Thanks for implementing this.

1. I feel that instead of 2.5sec maybe we should increase it to 5 or more. 
2. Currently the subagent sessions keep showing IDLE until viewed. Maybe they shouldn't wait at all to be viewed or if the parent session is viewed should update the viewed state for subagent sessions too.

What do you think?

pascalandr:
@shantur,
1 - right, your original timer was 5s but I felt it was maybe too long now with unseen session staying iddle tagged, but I really don't mind, we can change it again later if needed. I set it back to 5s.
2 - I thought about this, either way have an an advantage, I think we should just try as is and change it later if it's annoying but at the momment I prefer to have sub session with the same behaviour, it helps with discoverability, sub session may also be interesting to read (well I like to read them quickly).

shantur:
@pascalandr - I think having subagent sessions as showing IDLE all the time reduces the impact of this PR and feature looks half cooked.
My workflow relies heavily on subagents and this would mean that if I have to get rid of the subagent idle marker, I have to go and view each subagent session which mostly don't happen.
For simplicity, the subagent sessions can just hide idle after 5 seconds and don't need to wait for the visibility.

pascalandr:
> @pascalandr - I think having subagent sessions as showing IDLE all the time reduces the impact of this PR and feature looks half cooked. My workflow relies heavily on subagents and this would mean that if I have to get rid of the subagent idle marker, I have to go and view each subagent session which mostly don't happen. For simplicity, the subagent sessions can just hide idle after 5 seconds and don't need to wait for the visibility.

Really ? I also heavily rely on subagents but that's the reason why I would need this.
But I get your point, what about adding an option to enable or disable this ? (disabled by default )

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

Check how configs are owned in packages/server and there are 2 config and state
Is the config setting for this PR part of config or state?

shantur:
Thanks @pascalandr

PR #400, opened 2026-05-07
feat(ui): add chat history search

shantur:
Add session-scoped chat history search that opens from Cmd/Ctrl+F or the session header search button. The search scans normalized message records rather than rendered DOM so virtualized/offscreen history can be found, while hidden synthetic content and preference-hidden thinking remain excluded from normal results.

Render matches in the message stream and timeline with yellow highlights, jump between matches, expand matching thinking blocks, and scroll the active inline match into view for long messages. Search waits for at least three characters and debounces indexing so long sessions show an explicit searching state instead of a false empty result.

Validation: npm run typecheck --workspace @codenomad/ui; npm run build --workspace @codenomad/ui.


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

Review with latest changes

shantur:
@CodeNomadBot

Check the new state of PR, it was be been reworked

PR #407, opened 2026-05-08
feat(ui): collapse pasted text in chat history

### LINKED ISSUE #266: [Feature] A way to hide portions of a prompt in codenomad display
I have commands that issue a detailed prompt and I would like to hide portions of that in the codenomad display. I was previously doing this by placing `<div style="display:none">` in the prompt. However in a recent version, this loophole was closed and these tags are now being escaped.

My request is that there would be a way to hide portions of the prompt with a codenomad-specific tag. Perhaps this could look like `<codenomad:hide>(portions of prompt to hide)</codenomad:hide>`. Or perhaps it would be something like `<codenomad:start-hide />(portions of prompt to hide)<codenomad:end-hide />`.

In any case this would be specific to codenomad, and ideally these tags would be stripped out before being passed to the LLM.

pascalandr:
## Summary
- Collapse intact placeholder-backed long pasted text blocks by default in user message history for readability.
- Keep pasted content fully visible to the model; this is a display-only history treatment.
- Persist pasted-text collapse metadata across optimistic message replacement and app reloads.
- Fall back to normal plain-text history rendering when the placeholder-backed pasted structure has been edited or is no longer intact before send.

Fixes #266

## Behavior
When a user submits a prompt containing an intact long pasted block from the existing pasted-text placeholder flow, the full pasted text is still sent to the model unchanged, but the chat history renders that pasted section as a collapsed expandable block.

This applies to the existing placeholder-backed long-paste flow only. If the pasted placeholder structure is broken before submission, the message renders in history as normal plain text with no collapse metadata.

## Verification
- `npx tsx --test "packages/ui/src/components/prompt-input/submitPrompt.test.ts" "packages/ui/src/lib/prompt-display-metadata.test.ts" "packages/ui/src/stores/message-prompt-display.test.ts"`

## Local desktop verification
- Built executable: `packages/tauri-app/target/release/codenomad-tauri.exe`
- Built installer: `packages/tauri-app/target/release/bundle/nsis/CodeNomad_0.15.0_x64-setup.exe`
- Launched the built Tauri executable locally and confirmed startup succeeded.

image


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
@CodeNomadBot

Same review again please

shantur:
@pascalandr Thanks for working on this, but not sure we should do it. It might confuse users.

pascalandr:
> @pascalandr Thanks for working on this, but not sure we should do it. It might confuse users.

I'm not sure how it can confuse users, what do you mean ?

shantur:
> > @pascalandr Thanks for working on this, but not sure we should do it. It might confuse users.
> 
> I'm not sure how it can confuse users, what do you mean ?

I see that your are using <codenomad:hide> which is very explicit. My take is

1. I don't see this feature being used by more than 1 user.
2. Sending `Intro<codenomad:hide>Secret</codenomad:hide>Outro` will surely confuse models about what is sent
3. If the user is sending secrets to model, its already in plain text everywhere.

pascalandr:
Oh I see but this is not the use case here : https://github.com/NeuralNomadsAI/CodeNomad/issues/266#issuecomment-4171012876
My wording was bad, I should have used <codenomad:collapse> instead , this is not about secrets but about convenience. This could be very useful when pasting something in your prompt but you want it to stay collapsed for the user for convenience but still not hidden to the model.
but I realize indeed this is exactly how I would like pasted text in my prompt to appear in the message window collapsed by default (like collapsed tool for example) so my implementation is bad at the moment, What if pasted text instead of converted to raw text in bulk could keep its form and be collapsed by defaut (and uncollapsed if needed) if it's possible, does it make sens ?

shantur:
If we can do something without sending additional stuff to LLM then it could used. 
For paste maybe, we can surround with with codeblocks with language paste and detect it automatically and collapse it. Models do understand markdown syntax and could work.

Maybe if needed, a codeblock with language hidden would be a better way, that way model would understand the codeblock and we can treat this codeblock specially and hide it.

TBH, I see this as low value vs high effort and maintenance and not very keen on it. It all adds up in the markdown rendering logic which is a heavy computation part already.

pascalandr:
> If we can do something without sending additional stuff to LLM then it could used. For paste maybe, we can surround with with codeblocks with language paste and detect it automatically and collapse it. Models do understand markdown syntax and could work.
> 
> Maybe if needed, a codeblock with language hidden would be a better way, that way model would understand the codeblock and we can treat this codeblock specially and hide it.
> 
> TBH, I see this as low value vs high effort and maintenance and not very keen on it. It all adds up in the markdown rendering logic which is a heavy computation part already.

ok I'll keep this in mind and will go on only if the trade-off is worthwhile enough

pascalandr:
@CodeNomadBot
this a new implementation
Check if it adds maintenance or computation.

check for regressions
Be the PR gatekeeper

pascalandr:
@CodeNomadBot
Same review

pascalandr:
@CodeNomadBot
Same review

pascalandr:
@CodeNomadBot
can you confirm the implementation trade-off is worthwhile for this feature

pascalandr:
@shantur I think this is ready to merge. I didn’t realize I needed this feature at first, but after using it, I really really like it.

shantur:
@CodeNomadBot

Give me a very brief overview of how this feature is implemented in the latest version?
Don't give me code reference, just technical overview

shantur:
@pascalandr - I think if we want this we should implement using a markdown codeblock with `pasted` or some other language. If I understand it correctly the pasted text state is stored only in UI which resets as soon as the workspace is opened again. 

The markdown block would fix that but at the same time it happens that I add pasted text in the middle of texts which I expect to be read inline rather than a block so we need to think how to allow user to handle 2 types of pasted texts.

pascalandr:
> @pascalandr - I think if we want this we should implement using a markdown codeblock with `pasted` or some other language. If I understand it correctly the pasted text state is stored only in UI which resets as soon as the workspace is opened again.
> 
> The markdown block would fix that but at the same time it happens that I add pasted text in the middle of texts which I expect to be read inline rather than a block so we need to think how to allow user to handle 2 types of pasted texts.

Not quite, I don't think you understand it correctly, it does persist across reloads.
If you try it, I think the behavior is actually pretty nice, and it also helps performance because large pasted blocks do not need to be rendered in the UI until expanded.

shantur:
@CodeNomadBot

> That metadata is attached to the optimistic user message

What metadata is attached to the message? Give me some more info how its attached

shantur:
@pascalandr 

Can you pleaes change the pasted text to be rendered exactly as the markdown text block with all the copy and other controls. Maybe use the same component by just replacing it to be ``` ` ` `<pasted Text> ` ` ` ``` and collapsing it by default.

pascalandr:
@shantur 
collaped (default nothing loaded in the UI)
image
expanded (lazy loading)
image

shantur:
@CodeNomadBot

Re-review the new changes

shantur:
@CodeNomadBot

Try again

shantur:
@CodeNomadBot

Try again

shantur:
@CodeNomadBot

Please work

shantur:
@CodeNomadBot

Check latest

shantur:
@pascalandr - Please resolve conflicts last time, I will merge it straight in if bot is ok

pascalandr:
@CodeNomadBot
I think you are mistaking, there is no existing drop-in component in the current codebase that gives:
- standard code-block UI
- built-in collapse
- and preserved Markdown rendering for arbitrary pasted content

Standard code-block rendering exists, but not as a reusable collapsible primitive and preserved markdown for this case.
So on the contrary, the current implementation is not ignoring an obvious existing path, it is making a different rendering tradeoff on purpose to respect the requested “render exactly as the markdown text block with all the copy and other controls” behavior.
Please update your review.

pascalandr:
@CodeNomadBot
Please confirm the conflict merge resolution didn't break anything.

pascalandr:
@shantur 
ready to merge

shantur:
Thanks @pascalandr

Issue #383, opened 2026-04-29
[feature] ssh remote


linonetwo:
<img width="1157" height="428" alt="Image" src="https://github.com/user-attachments/assets/fce25467-f4e9-4ba2-8d99-8de3ad080a58" />

I hope I can use this to replace the VS Code SSH remote and open code in the remote.

You can show local configured SSH remotes, like VS Code does, and connect to a remote folder in a click.

shantur:
Check Sidecars

https://github.com/NeuralNomadsAI/CodeNomad#sidecars

linonetwo:
@shantur I see, it allows opening VS Code on the side. Actually, I mean doing something like VS Code, but not using vscode, where I want to control opencode running on a remote server using an SSH tunnel like VS Code could do for me.

shantur:
@linonetwo  - for that you can run CodeNomad server in the remote server

linonetwo:
And setup ssh port forward manually? I think It could be further automated to improve the experience.Like VS Code does

shantur:
> And setup ssh port forward manually? I think It could be further automated to improve the experience.Like VS Code does

No you don't need to port forward SSH or anything. You can use ttyd as in the README to get terminal. TBH I am not sure what is the VS Code behaviour, any video would be helpful

linonetwo:
Hi, I mean this one https://code.visualstudio.com/docs/remote/ssh#_remember-hosts-and-advanced-settings

<img width="786" height="500" alt="Image" src="https://github.com/user-attachments/assets/4c7494b2-f7a1-444c-a190-fb583ca05da2" />

It automatically shows all the servers available in user ssh config. Clicking a button auto-connects to it and handles all the port forwarding without using the terminal. If no VS Code server exists, it will auto-download the server using SSH.

shantur:
So what you can do with CodeNomad is

1. Run CodeNomad server using 1 command on the ssh servers.
2. Connect the the remote codenomad server once with "Connect CodeNomad Server"
3. The server gets added to the list of the  Servers for easy connect next time.
4. Each server will be connected as its own window.

<img width="1106" height="523" alt="Image" src="https://github.com/user-attachments/assets/3398adfc-4c7b-4040-8119-d89f4a54e1f7" />

Issue #254, opened 2026-03-27
[Feature] Support for Docker Sandboxes


clayliddell:
Docker Sandboxes allow for spinning up OpenCode instances in a secure sandboxed environment (more details [here](https://docs.docker.com/ai/sandboxes/). I find it really useful for allowing agents a bit more freedom without having to worry about my host machine.

Is there anyway you could allow new OpenCode instances to be created using a custom command? Could you allow connecting an OpenCode server? Or another TUI instance some how?

Thanks for the awesome project btw 😄

PR #408, opened 2026-05-08
feat(runtime): add execution profiles

pascalandr:
## Summary
- Add server-scoped execution profiles for `local`, `wsl`, `docker`, `command`, and `ssh` runtimes so new workspaces can choose how the OpenCode runtime is launched.
- Add Execution Profile management in Settings, a Home screen selector for new workspaces started by this CodeNomad server, active runtime metadata in Instance Information, and duplicate actions for saved execution profiles.
- Add `Preview command` and `Test profile` flows so users can inspect launch commands and validate local/WSL/Docker/custom command/SSH profiles before using them.
- Harden runtime handling with WSL distro-aware launches/previews/tests, Docker-aware URL/env rewriting and localhost port publishing, SSH remote-shell port forwarding plus OpenCode config sync, sensitive environment/argument redaction in logs and previews, safer profile normalization, and cleanup when a spawned runtime fails readiness.

## Scope
- This PR is scoped to Execution Profiles only: local binary / WSL / Docker / custom command / SSH remote shell launch modes for workspace runtimes started by the current CodeNomad server.
- SSH execution profiles copy the packaged OpenCode config template to the remote host, run `opencode serve` on the configured remote host through `ssh`, forward the remote OpenCode runtime back to the current CodeNomad server, and reverse-forward callback traffic back to the current server.
- Connecting to another CodeNomad server remains covered by the existing `Connect CodeNomad Server` flow; network transport such as VPN, SSH tunnel, Tailscale, or reverse proxy stays external to CodeNomad.

## Validation
- `node --import tsx --test "packages/server/src/settings/binaries.test.ts" "packages/server/src/workspaces/execution-launch.test.ts" "packages/server/src/workspaces/__tests__/spawn.test.ts" "packages/server/src/workspaces/__tests__/git-clone.test.ts"`

## Notes
- Keeping this PR in draft until real manual validation is done against Docker, WSL, and SSH remote-shell targets.
- Pending manual checks: Linux VM, Hackintosh VM, WSL, Docker, SSH remote shell execution profile.
- Refs #254
- Refs #383
- Refs #5
- Refs #482


shantur:
@pascalandr - We need to discuss this as it won't work as it is. It depends on the opencode-config folder that resides on the server.

pascalandr:
> @pascalandr - We need to discuss this as it won't work as it is. It depends on the opencode-config folder that resides on the server.

Resolved in latest push. The SSH execution profile no longer points the remote OpenCode process at the server-local config path. Before launching the SSH runtime, CodeNomad now archives the packaged OpenCode config template and streams it over ssh into a per-workspace directory on the remote host under /tmp, then sets OPENCODE_CONFIG_DIR to that remote path for the launched opencode serve process. Cleanup is best-effort on exit/startup failure. Docker/WSL/local continue using their existing local/mapped config paths.

pascalandr:
Updated the SSH config sync to avoid the tar dependency. SSH profiles now clean the remote temp config directory with ssh, copy the packaged OpenCode config with scp -r, and keep using that remote path for OPENCODE_CONFIG_DIR.

Validated after rebasing on the latest branch:
- node --import tsx --test "packages/server/src/settings/binaries.test.ts" "packages/server/src/workspaces/execution-launch.test.ts" "packages/server/src/workspaces/__tests__/spawn.test.ts" "packages/server/src/workspaces/__tests__/git-clone.test.ts"

shantur:
I was thinking about how to make opencode_config implementation better. Essentially it is an opencode plugin that is used for codenomad. 
I will take it up as part of #419 but the idea is

1. Opencode_config will be bundled as a npm package in CodeNomad 
2. When starting opencode, CodeNomad will inject OPENCODE_CONFIG_CONTENT env var to add CodeNomad plugin npm package. It can be via file:// or http:// depending on the requirement. If OPENCODE_CONFIG_CONTENT is present, it will be easy to inject another plugin.
3. We free up the OPENCODE_CONFIG_DIR for users and plugin can be provided to other execution profiles too

shantur:
opencode-config features is being worked in #433

pascalandr:
@shantur
Thanks, with #433, #408 can stop copying `opencode-config` and only make the packaged plugin `.tgz` reachable from the target runtime: local/command use it directly,
WSL keeps the path rewrite, Docker mounts it into the container, and SSH copies it to `/tmp` then points `OPENCODE_CONFIG_CONTENT` at that remote file.

shantur:
@pascalandr -  I have been thinking about this and we can support any opencode environment ( even a custom script ) until

1. CodeNomad and opencode variables and arguments passed into the script is passed onto the final opencode instance
2. Script ensures that codenomad-plugin is copied and available to final opencode instance and the variable is adjusted accordingly.
3. STDOUT of opencode binary is routed by script to its stdout.

Hope I haven't forgotten anything else but If a script can support these then it can run opencode anywhere over SSH, Docker, K8s.
User just need to select that script in CodeNomad.

We can think of a good UX for easy setup without building all of it in CodeNomad.

pascalandr:
@shantur  Yes, I think we agree on the direction, but I'm not sure one single config can cover Docker, SSH, WSL, K8s, and custom scripts without becoming too abstract. So I’d still keep the current UX for common cases as presets.

the balance I’m aiming for is:
- one shared launcher contract underneath
- built-in profiles as presets/templates over that shared path
- and `command` as the generic advanced path

That is also what I just pushed: the backend now converges on a shared launcher flow while keeping the current specialized profile UX.

Still need to be tested against each case...

shantur:
@pascalandr 

> but I'm not sure one single config can cover Docker, SSH, WSL, K8s, and custom scripts without becoming too abstract.

100% agreed. I am more leaning towards this feature to come as docs / samples / helper scripts rather than building all in UI
There are lots of variables on each config that are implementation dependent and I don't think its would be feasible for us to take over all that in code. This is an advanced feature.

Maybe in future it can come up with some skill that people can point and get their LLMs to generate launch scripts for their own cases.

pascalandr:
closing this as discussed in Discord

PR #411, opened 2026-05-08
fix(ui): return to active projects from home

### LINKED ISSUE #281: [UX/UI] Improve Project Navigation and Prevent Duplicate Project Instances
### 🚀 Problem Statement
The current navigation flow when adding or switching projects is unintuitive and leads to redundant project instances.

**The issues are twofold:**
1. **Low Discoverability:** The "X" button in the top-right corner used to return to the chat/project view is not obvious to users.
2. **Duplicate Instances:** Clicking on a project that is already active opens a second instance of that project instead of switching back to the existing one.

### 💡 Proposed Solution
- **Singleton Project Pattern:** Implement a check to see if a project is already active. If it is, clicking the project card should navigate the user back to the existing session.
- **Enhanced Navigation UI:** Provide a more prominent "Return to Active Project" button or clear breadcrumbs to make the exit path from the entry interface more intuitive.

### 📝 Steps to Reproduce
1. Open a project and enter the chat interface.
2. Click the **"+ Add Project"** button to return to the entry/selection screen.
3. Observe the small "X" button (unintuitive exit).
4. Click the same project name again.
5. Observe that a duplicate project instance is opened.

### ✅ Expected Behavior
- Users should be able to clearly identify how to return to their work.
- The system should recognize active projects and prevent multiple instances of the same project from running simultaneously.

pascalandr:
## Summary
- Return to an existing open project when users choose that from the already-open folder dialog.
- When users click a recent folder that is already open, ask whether to switch to the open project or intentionally open another instance.
- Surface a prominent Return to Active Project action so leaving the home screen no longer depends on the subtle top-right close affordance.
- Keep multi-instance workflows available without requiring users to remember a separate row-level button.

Fixes #281

## Why
Users can accidentally open the same workspace several times from the add-project screen. The app now detects that ambiguity when the already-open folder is clicked and asks what should happen, instead of making the user remember a special alternate button ahead of time.

The dialog keeps the copy short: the title states that the project is already open, and the buttons carry the actual choices.

## Verification
- `git diff --check`
- `npm run build:tauri`

### DIFF
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -34,6 +34,7 @@ import {
 import { useConfig } from "./stores/preferences"
 import {
   createInstance,
+  getExistingInstanceForFolder,
   instances,
   stopInstance,
   disconnectedInstance,
@@ -257,15 +258,26 @@ const App: Component = () => {
 
   const launchErrorMessage = () => launchError()?.message ?? ""
 
-  async function handleSelectFolder(folderPath: string, binaryPath?: string) {
+  async function handleSelectFolder(folderPath: string, binaryPath?: string, options?: { forceNew?: boolean }) {
     if (!folderPath) {
       return
     }
-    setIsSelectingFolder(true)
     const selectedBinary = binaryPath || serverSettings().opencodeBinary || "opencode"
+    recordWorkspaceLaunch(folderPath, selectedBinary)
+    clearLaunchError()
+
+    if (!options?.forceNew) {
+      const existingInstance = getExistingInstanceForFolder(folderPath)
+      if (existingInstance) {
+        selectInstanceTab(existingInstance.id)
+        setShowFolderSelection(false)
+        log.info("Selected existing instance", { instanceId: existingInstance.id, folderPath })
+        return
+      }
+    }
+
+    setIsSelectingFolder(true)
     try {
-      recordWorkspaceLaunch(folderPath, selectedBinary)
-      clearLaunchError()
       const instanceId = await createInstance(folderPath, selectedBinary)
       selectInstanceTab(instanceId)
       setShowFolderSelection(false)
@@ -607,6 +619,7 @@ const App: Component = () => {
                 onSelectFolder={handleSelectFolder}
                 isLoading={isSelectingFolder()}
                 onOpenSidecar={handleOpenSidecarPicker}
+                activeProjectLabel={activeInstance()?.folder ?? null}
                 onClose={() => {
                   setShowFolderSelection(false)
                   clearLaunchError()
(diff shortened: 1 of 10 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
@CodeNomadBot
same review again

shantur:
Hey @pascalandr 

Thanks for fixing the issue.
I don't think we should fix this. It's more a PEBCAK.

pascalandr:
ok but should I close the issue then ?

shantur:
Yes please.

pascalandr:
Opening again, actually I changed my mind about this and think this should be discussed more.

I also personally would like to have this feature, I often have so many tabs in codenomad I can open open the same workspace several times by mistakes not by choice. In my implementation here, I just add helpers to avoid that.
@Shantur what is your exact problem with that, maybe I can get a better approch ?

shantur:
> Opening again, actually I changed my mind about this and think this should be discussed more.
> 
> I also personally would like to have this feature, I often have so many tabs in codenomad I can open open the same workspace several times by mistakes not by choice. In my implementation here, I just add helpers to avoid that. @shantur what is your exact problem with that, maybe I can get a better approch ?

I see it again as PEBCAK 
I do open multiple instances for same workspace many times. If it's really needed, I would just add a check if the workspace is already opened and user clicks on it to present a dialog that asks user if they want to create a new session or switch to that workspace. That would be a simpler and cleaner approach IMO

pascalandr:
did you see the proposition ?
image
you think a dialog is a better option ?

shantur:
> you think a dialog is a better option ?

Yes, as then it becomes more intuitive and app handles situation smartly. If first time then new instance, if already running an instance, provides an option

pascalandr:
I'm not sure the screenshot is clear enough, the proposed solution offer exactly the same result as you described but without the additionnal dialog,  same result but faster with less clics, I don't get it. Did you build the PR and tried it ?

shantur:
> I'm not sure the screenshot is clear enough, the proposed solution offer exactly the same result as you described but without the additionnal dialog, same result but faster with less clics, I don't get it. Did you build the PR and tried it ?

I have, the different button is what user needs to  remember to click when they want to launch another instance. In the proposed way, the app is smart enough to ask user what needs to be done.

pascalandr:
like this ?
image

shantur:
Yes, @pascalandr

pascalandr:
@CodeNomadBot
same review again

shantur (approved):

shantur:
Thanks @pascalandr

PR #367, opened 2026-04-25
fix(ui): show pending tool prompts

pascalandr:
Fixes #290

## Summary
- keep tool calls expanded while a permission or question is attached, even before active interruption reconciliation catches up
- use the same pending-interruption rule for diagnostics expansion and manual toggle guards
- make permission asks more reliably visible in the message stream when tool calls are normally collapsed

## Testing
- `npm run typecheck --workspace @codenomad/ui`
- `npm run build --workspace @codenomad/ui`

PR #422, opened 2026-05-09
fix(ui): harden permission event reconciliation

### LINKED ISSUE #290: [Bug]: sometime missing display ask
### App Variant

Electron

win11

### Issue Summary

sometime missing display ask

### Steps to Reproduce

Image

### Expected Behavior

.

### Logs & Screenshots

_No response_

### Additional Context

_No response_

pascalandr:
Fixes #290
Supersedes #367

## Summary
- keep permission replies idempotent by clearing both the global queue and message-v2 state after a successful reply
- reconcile pending permissions when tool parts arrive, matching the existing question flow
- prevent stale or duplicate permission events from leaving mismatched active/attached permission state
- drain YOLO auto-accept from the permission flow itself instead of the rendered instance shell
- preserve optimistic user attachments across SSE event reordering until official server parts arrive

## Bugs and Fragile Flows Fixed
- **Reply cleanup depended on `permission.replied` SSE**: `sendPermissionResponse()` removed only the global permission queue entry, so if the replied event was delayed or missed, inline tool-call UI could keep showing a stale permission.
- **Delayed permission events could resurrect handled prompts**: a late `permission.updated` for an already accepted permission could re-add it locally until reload corrected state from `permission.list()`.
- **Duplicate permission payloads were discarded or downgraded**: repeated events for the same permission ID could drop richer session/worktree/tool-call metadata when a later payload was partial.
- **Pending session badges could drift**: replacing a permission payload with a different session ID did not rebalance pending permission counters.
- **Permission/tool-call attachment was asymmetric with questions**: live `message.part.updated` re-linked questions, but permissions depended on narrower incidental rebind paths or full message reload.
- **message-v2 could keep multiple attachment keys for one permission**: a permission could remain under an old global/ID key and later also be inserted under the resolved tool part.
- **message-v2 active permission could stay stale**: active state was only refreshed in limited cases instead of being derived from the ordered pending queue.
- **Optimistic user attachments could disappear during ID replacement**: `message.updated` could replace a synthetic message ID before user part events arrived and clear the optimistic parts, including resolved pasted/file attachment content.
- **YOLO auto-accept depended on shell rendering**: the auto-accept drain lived in `instance-shell2`, so heavy message rendering or permission UI churn could delay replies even though the request was already queued.
- **YOLO retry could amplify UI churn or outlive permissions**: transient reply failures were retried through a reactive render effect, and retry state was not fully cleaned up when permissions were removed, queues were cleared, or YOLO was disabled.

## Implementation
- Add short-lived local tombstones for replied permission IDs to ignore stale permission updates after a successful local or SSE reply.
- Merge duplicate permission entries conservatively in both the global queue and message-v2, preserving existing IDs and nested metadata/time/tool data when incoming events are partial.
- Update duplicate permission entries in the global queue, including pending session counters and worktree slug mapping.
- Remove permissions from both `instances.ts` and `message-v2` after successful replies so the UI does not depend on receiving `permission.replied`.
- Run `reconcilePendingPermissionsV2()` for every live tool part update.
- Make `message-v2` enforce one `byMessage` attachment per permission ID and recalculate active permission from the sorted queue.
- Allow permission reconciliation to repair entries whose `partId` exists but whose `byMessage` attachment was lost.
- Preserve optimistic user message parts when temporary message IDs are replaced, then remove matching synthetic text/file parts only when the official server part arrives.
- Move YOLO draining into `permission-auto-accept.ts` and trigger it from permission sync, live permission enqueue, and the Status-tab toggle.
- Add bounded retry backoff for failed YOLO replies while verifying the permission is still pending before retrying.
- Clear YOLO retry/in-flight state when permissions are removed, queues are cleared, or per-session auto-accept is disabled.

## YOLO Mode
- No product behavior change: YOLO remains per-session and continues to reply with `once` only.
- This only decouples auto-accept from rendered UI so queued permissions can be answered without waiting for permission/tool-call UI to render.

## Validation
- `git diff --check`
- final read-only gatekeeper review: clear

## Notes
- `packages/ui` does not currently define a test script/runner, so this PR is validated with typecheck/build rather than new automated unit tests.
- Build still emits the existing `virtua` JSX import-source warning and large chunk warnings.

### DIFF
--- a/packages/ui/src/components/instance/instance-shell2.tsx
+++ b/packages/ui/src/components/instance/instance-shell2.tsx
@@ -36,7 +36,7 @@ import { serverApi } from "../../lib/api-client"
 import { loadBackgroundProcesses } from "../../stores/background-processes"
 import { BackgroundProcessOutputDialog } from "../background-process-output-dialog"
 import { useI18n } from "../../lib/i18n"
-import { getPermissionQueue, getPermissionQueueLength, getQuestionQueueLength, sendPermissionResponse } from "../../stores/instances"
+import { getPermissionQueueLength, getQuestionQueueLength } from "../../stores/instances"
 import SessionSidebar from "./shell/SessionSidebar"
 import { useSessionSidebarRequests } from "./shell/useSessionSidebarRequests"
 import RightPanel from "./shell/right-panel/RightPanel"
@@ -58,13 +58,7 @@ import { useDrawerHostMeasure } from "./shell/useDrawerHostMeasure"
 import { useDrawerResize } from "./shell/useDrawerResize"
 import { useSessionCache } from "./shell/useSessionCache"
 import { useInstanceSessionContext } from "./shell/useInstanceSessionContext"
-import { getPermissionSessionId } from "../../types/permission"
-import {
-  canAutoRespondPermission,
-  finishAutoRespondPermission,
-  getPermissionAutoAcceptInFlightVersion,
-  isPermissionAutoAcceptEnabled,
-} from "../../stores/permission-auto-accept"
+import { isPermissionAutoAcceptEnabled } from "../../stores/permission-auto-accept"
 
 const log = getLogger("session")
 const OPEN_SESSION_SEARCH_EVENT = "codenomad:open-session-search"
@@ -269,8 +263,6 @@ const InstanceShell2: Component<InstanceShellProps> = (props) => {
     return permissions + questions > 0
   })
 
-  const permissionQueue = createMemo(() => getPermissionQueue(props.instance.id))
-
   const activePromptInputApi = createMemo(() => {
     const sessionId = activeSessionIdForInstance()
     if (!sessionId || sessionId === "info") return null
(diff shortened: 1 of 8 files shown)


pascalandr:
Final gatekeeper pass is clear after the latest fixes.

Validated locally:
- `git diff --check`
- read-only gatekeeper review focused on duplicate permission merge, message-v2 optimistic attachment preservation, YOLO retry cleanup, StatusTab delegation, and YOLO once-only/session-scoped behavior

Latest commit pushed: `440e5668 fix(ui): preserve permission state during yolo drain`.

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
Addressed the non-blocking cleanup notes from the latest gatekeeper pass in `6f288b64`:

- replied permission tombstones are now pruned when replies are marked or checked
- replied tombstones are cleared when an instance is removed
- removed the dead `updatedQueue` local from `removePermissionFromQueue()`

I intentionally left the duplicate permission merge helper in place for this PR because extracting a shared helper would touch both queue/message-v2 boundaries again and add churn after the behavior is already validated. If this logic grows further, a shared helper is the right follow-up.

Revalidated:
- `git diff --check`

pascalandr:
@CodeNomadBot
same review

shantur:
@pascalandr  - Were you able to reproduce this issue?

pascalandr:
we already had this discussion https://github.com/NeuralNomadsAI/CodeNomad/pull/367#issuecomment-4322312238

I already had all kind of buggy behavior with permission this is why I have dig into it and found several issues.

shantur:
TBH, this is touching too many core parts and I am not sure why synthetic parts are touched while fixing permissions?
We really need this to be targeted to be able to review it properly.
Why are we keeping a map of replied permissions in UI when the source of truth is the opencode server?
Would you be able to document the sure shot way to reproducing this? Currently the PR seems a combination of optimistic fixes done by LLM

pascalandr:
Well, I described the permission problems I had observed, then explained what I thought could be causing them and looked for existing bugs or weak spots matching those behaviors.
For the YOLO mode issue specifically, I’m pretty sure the fix matches the observed problem exactly.
That said, I agree this should be reworked to be more targeted, conservative, and explained.

pascalandr:
I split this broad PR into smaller targeted PRs for better review:

- #424: YOLO-only, drains auto-accept outside shell render.
- #425: stale permission events, prevents recently answered prompts from being resurrected by delayed pending events/sync.
- #426: permission/tool-call reconciliation, aligns permission re-linking with the existing question flow.

For reminder, these are issues I have personally hit multiple times. I cannot reproduce them deterministically on demand because they happen when the UI is under heavy pressure, but they happened often enough to reveal a consistent pattern.

PR #423, opened 2026-05-09
fix(ui): follow up idle badge behavior

pascalandr:
## Summary
- Follow-up to #395 after the idle badge persistence behavior regressed around `keepUnseenSubagentIdleStatus`.
- Restores `keepUnseenSubagentIdleStatus` handling across session, active-session, and instance-tab badge helpers.
- Starts a 5s fade when a viewed idle session/thread is seen, then clears each exact idle transition after the animation even if focus moves away.
- Keeps unseen subagent idle badges visible when the preference is enabled, and keeps aggregate instance badges solid when any other visible idle contributor is still unseen.

## Validation
- `git diff --check`
- Gatekeeper re-review: no blocking findings remain for the prior preference/race findings.

## Notes
- `node --test packages/ui/src/stores/session-status.test.ts` was attempted, but the repo's extensionless TypeScript imports are not directly resolvable by Node's test runner.

### DIFF
--- a/packages/ui/src/components/instance-tab.tsx
+++ b/packages/ui/src/components/instance-tab.tsx
@@ -1,8 +1,9 @@
-import { Component, Show, createMemo } from "solid-js"
+import { Component, Show, createMemo, createSignal, onCleanup } from "solid-js"
 import type { Instance } from "../types/instance"
-import { getInstanceSessionIndicatorStatus } from "../stores/session-status"
+import { getInstanceIdleFadeClass, getInstanceSessionIndicatorStatus } from "../stores/session-status"
 import { FolderOpen, ShieldAlert, X } from "lucide-solid"
 import { useI18n } from "../lib/i18n"
+import { useConfig } from "../stores/preferences"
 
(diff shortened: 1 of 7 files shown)


shantur:
@pascalandr - Can you please let me know what was the regression as I had slightly changed the implementation around the idle badges for child sessions

pascalandr:
Well the option to let child session behave consistently was broken, some sessions where iddle/green forever for some reason. The other problem was we had to stay more than 5 s in a a tab or in a session to trigger the change to inactive/grey, this was not how it was supposed to behave, not the asked feature. Now as soon as you select select a tab or a session it start to fad to grey during five 5s but doesn't stop if you change of tab or session earlier than 5 s.
Hum but I see still a problem in default mode sorry the subsessions were supposed to fade at the same time as the master session but it's not the case anymore. Fixing this.

pascalandr:
ok now it works exactly as expected for both options

@CodeNomadBot
same review please

shantur (approved):

PR #424, opened 2026-05-09
fix(ui): drain yolo permissions outside shell render

pascalandr:
## Summary
- Split out from #422 as the YOLO-only permission fix.
- Moves YOLO auto-accept draining out of `InstanceShell` render effects and into the permission queue flow.
- Keeps behavior unchanged: YOLO remains per-session and replies with `once` only.
- Keeps in-flight cleanup for auto-accept attempts so duplicate sends are guarded outside UI render timing.

## Why
In YOLO mode, the app could still appear to wait on a permission or block the UI, even though auto-accept was enabled and the permission was already queued.

The auto-accept drain was tied to an `InstanceShell` render effect, so the reply path depended on a specific UI shell rendering. Draining from permission sync/enqueue and from the YOLO toggle makes auto-accept run from the permission queue itself instead of UI render timing.

## Validation
- `git diff --check`

### DIFF
--- a/packages/ui/src/stores/permission-auto-accept.ts
+++ b/packages/ui/src/stores/permission-auto-accept.ts
@@ -4,8 +4,6 @@ import { getPermissionSessionId } from "../types/permission"
 import { getLogger } from "../lib/logger"
 
 const STORAGE_KEY = "codenomad:permission-auto-accept:v1"
-const RETRY_BASE_DELAY_MS = 1_000
-const RETRY_MAX_DELAY_MS = 10_000
 
 const log = getLogger("api")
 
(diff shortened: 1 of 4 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

packages/ui/src/stores/permission-auto-accept.ts
```diff
@@ -54,28 +65,109 @@ export function setPermissionAutoAcceptEnabled(instanceId: string, sessionId: st
...
+  inFlight.add(requestKey)
+
+  void responder(instanceId, sessionId, permission.id, "once")
+    .then(() => {
+      clearRetry(requestKey)
+    })
+    .catch((error) => {
+      log.error("Failed to auto-accept permission", error)
+      if (isPending(instanceId, permission.id) && isPermissionAutoAcceptEnabled(instanceId, sessionId)) {
+        scheduleRetry(instanceId, permission, responder, isPending, requestKey)
```
shantur: Do we really need retry? It's a local server

pascalandr: Not really, you are right.

shantur:
@pascalandr - I am not sure how to test this manually, any ideas?

pascalandr:
> @pascalandr - I am not sure how to test this manually, any ideas?

I asked "look at the opencode code in my local repository "
while my workspace is not this repo.
And while the agent is doing that i change of session and tab and still work, we can see the original tab becoming red for 0.1 s every few seconds without displaying the permission UI. And making a report about the code ah end.
Works nicely on my side.

shantur:
Will take your word for it.

PR #425, opened 2026-05-09
fix(ui): ignore stale permission events after reply

pascalandr:
## Summary
- Split out from #422 as the stale-permission-events fix.
- Clears permission UI state immediately after a successful local permission reply instead of waiting for `permission.replied` SSE.
- Tracks replied permission IDs until a newer pending-permission sync observes that the server no longer reports them pending.
- Marks SSE `permission.replied` events into the same replied-ID path so delayed pending events/sync results cannot resurrect prompts that were just answered.

## Why
A permission prompt could remain on screen after clicking Allow/Deny, or clicking Allow could look like it did not take effect and require another click. Reloading could fix the UI, which pointed to stale local permission state rather than the server still waiting.

The UI receives permission state from local replies, SSE events, and pending-permission sync. If an older pending event or sync result is processed after a confirmed reply, the UI can re-add a permission that was already answered. Replied IDs stay suppressed until a sync started after the local reply proves the server has dropped that permission from the pending list.

## Validation
- `git diff --check`
- `node --test packages/ui/src/stores/permission-replies.test.ts`

### DIFF
--- a/packages/ui/src/stores/instances.ts
+++ b/packages/ui/src/stores/instances.ts
@@ -32,6 +32,12 @@ import { setSessionPendingPermission, setSessionPendingQuestion } from "./sessio
 import { setHasInstances } from "./ui"
 import { messageStoreBus } from "./message-v2/bus"
 import { upsertPermissionV2, removePermissionV2, upsertQuestionV2, removeQuestionV2 } from "./message-v2/bridge"
+import {
+  clearRepliedPermissions,
+  hasRepliedPermission,
+  markPermissionReplied,
+  pruneRepliedPermissions,
+} from "./permission-replies"
 import { clearCacheForInstance } from "../lib/global-cache"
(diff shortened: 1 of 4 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
@CodeNomadbot
same review

shantur:
@pascalandr - When does this happen?

pascalandr:
> @pascalandr - When does this happen?

It happens when I have many processes running at the same time, or maybe bigger compilation I think. I’m not sure about a specific trigger, but I get it regularly. Reloading fix the UI but the permission prompt is then dead.

shantur:
Thanks @pascalandr 

Once the conflict is resolved, I can merge this in

shantur:
Thanks @pascalandr

PR #426, opened 2026-05-09
fix(ui): reconcile permission tool attachments

### LINKED ISSUE #290: [Bug]: sometime missing display ask

pascalandr:
## Summary
- Split out from #422 as the permission/tool-call reconciliation fix.
- Reconciles pending permissions whenever live tool parts update, matching the existing question re-link path.
- Keeps one message-v2 attachment per server permission ID and recalculates the active permission from queue order.
- Conservatively merges duplicate or out-of-order permission updates so known session/message/tool routing metadata is not lost.
- Fixes #290

## Why

The observed failure shape is that permission prompts can appear missing, frozen, or attached in unexpected places when permission events and tool-call parts are observed in different orders. In those cases, the server-side permission may exist, but the UI can temporarily attach it globally, attach it to the wrong tool location, or fail to move it when the matching tool part arrives later.

This PR focuses on the UI-side attachment/order problem: one UI attachment per server permission ID, re-linking permissions when tool parts arrive, and preserving known routing metadata across duplicate/out-of-order updates. It does not attempt semantic deduplication across different permission IDs that happen to ask for the same logical approval.

## Validation
- `git diff --check`
- `npm exec --no -- tsx --test packages/ui/src/types/permission.test.ts packages/ui/src/stores/message-v2/instance-store.test.ts`
- `node --test packages/ui/src/stores/permission-replies.test.ts`


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
@CodeNomadBot
same review

shantur:
> later duplicate permission payloads can contain less metadata

Is this really true?

pascalandr:
> > later duplicate permission payloads can contain less metadata
> 
> Is this really true?

bad wording, modified

shantur (dismissed):

packages/ui/src/types/permission.ts
```diff
@@ -42,6 +42,48 @@ export interface PermissionReplyEventPropertiesLike {
...
+  "messageID",
+  "messageId",
+  "callID",
+  "callId",
+  "partID",
+  "partId",
+  "toolCallID",
+  "toolCallId",
+] as const
+
```
shantur, line 64: What's going on here?

pascalandr: It keeps the same fields that tell the UI where the permission belongs.

shantur (approved):

PR #433, opened 2026-05-11
Package CodeNomad OpenCode plugin

shantur:
## Summary
- Rename the OpenCode config template into a versioned npm-packable CodeNomad plugin package.
- Build and package the plugin through the server bundle, with Electron/Tauri carrying it via existing server resources.
- Replace OPENCODE_CONFIG_DIR injection with JSONC-aware OPENCODE_CONFIG_CONTENT merging that appends the CodeNomad plugin while preserving user config.

## Validation
- npm run build --workspace @codenomad/codenomad-opencode-plugin
- npm run prepare-plugin --workspace @neuralnomads/codenomad
- npm run typecheck --workspace @neuralnomads/codenomad
- npm run typecheck --workspace @neuralnomads/codenomad-electron-app
- node --import tsx --test \"src/opencode-plugin.test.ts\" \"src/workspaces/__tests__/spawn.test.ts\"

## Notes
- Production plugin loading uses an explicit npm file alias for the packaged tarball.
- Dev loading still references the TypeScript plugin entry directly.

PR #436, opened 2026-05-11
fix(electron): avoid shell for resource prep

pascalandr:
## Summary
- Avoid routing the Electron resource-prep Node invocation through the Windows shell.
- Fix local `build:win` when Node is installed under `C:\Program Files\nodejs\node.exe`.

## Validation
- `npm run build:win --workspace @neuralnomads/codenomad-electron-app`

## Notes
- This targets `opencode-config-merge` so it can be merged into #433.


shantur:
Thanks for PR

PR #444, opened 2026-05-14
fix(ui): recover legacy sessions with export fallback

pascalandr:
## Summary
- recover legacy OpenCode sessions that still appear in `session.list` but fail to open in CodeNomad because the transcript endpoint rejects their historical message payload
- add a narrow recovery path that falls back to `opencode export` only for that known legacy validation failure
- keep normal sessions on the existing `session.messages` flow and harden the fallback so it does not become a new source of hangs or memory spikes

## Problem
The broken sessions are not missing from OpenCode entirely.

They still show up in `session.list`, so CodeNomad can discover them and render them in the session sidebar. The failure happens later, when CodeNomad tries to hydrate the transcript by calling the OpenCode v2 `session.messages` endpoint.

For some historical sessions, at least one assistant message does not satisfy the current v2 response schema anymore. The concrete legacy shape we reproduced is a missing `info.agent` field inside a transcript message. When that happens, the OpenCode endpoint rejects the entire `session.messages` response during validation and returns a `400` body-validation error instead of returning partial transcript data.

That means CodeNomad never receives a message list to hydrate at all. This is why the session can still be visible in the UI while opening it fails.

The error we reproduced on the affected sessions is:

```text
Missing key
  at [1]["info"]["agent"]
```

## Why This Is Not Just A UI Hydration Fix
A UI-only compatibility layer would only help if CodeNomad received the old transcript payload and merely failed to render or normalize it.

That is not what happens here.

The blocking failure occurs inside the OpenCode `session.messages` API path itself, before CodeNomad can hydrate anything. Since the endpoint returns a validation error instead of message data, CodeNomad cannot repair the payload in memory because it never gets access to the transcript through the normal API route.

That is why this PR uses `opencode export` as a read fallback for a very specific legacy failure mode rather than trying to reinterpret `session.messages` client-side.

## Recovery Strategy
The new behavior is:

1. CodeNomad first uses the normal `session.messages` request path.
2. If that call succeeds, nothing changes.
3. If that call fails with the specific legacy body-validation signature for missing `info.agent`, CodeNomad falls back to `opencode export <sessionId>`.
4. The exported transcript is then used only as a replacement read source for hydration.

Important scope limits:
- this does **not** export and re-import sessions
- this does **not** mutate or migrate session data on disk
- this does **not** bypass normal API loading for healthy sessions
- this does **not** swallow arbitrary schema failures; it is restricted to the known legacy missing-agent validation path

## Hardening In This PR
Because the affected transcripts can be very large, the fallback path is also hardened so the recovery mechanism does not introduce a second failure mode.

Specifically:
- the fallback matcher is restricted to the reproduced legacy validation path (`["info"]["agent"]`), so unrelated schema regressions still surface normally
- the server exports to a temporary file and streams that file back instead of buffering the full export payload in memory
- the export subprocess is abortable and protected by a timeout so a stuck export cannot hang the request forever
- temporary export files are cleaned up when the request finishes or is aborted

## Files Changed
- `packages/ui/src/stores/session-message-source.ts`
  - adds the client-side message-loading fallback entry point
- `packages/ui/src/stores/session-message-fallback.ts`
  - isolates the strict legacy error matcher
- `packages/server/src/workspaces/manager.ts`
  - adds the server-side export-to-temp-file helper with timeout and cleanup
- `packages/server/src/server/routes/workspaces.ts`
  - exposes the export stream endpoint used only by the fallback path
- `packages/ui/src/stores/session-api.ts`
  - routes transcript hydration through the guarded loader
- `packages/ui/src/stores/session-state.ts`
  - reuses the same guarded loader for deep blank-session inspection
- `packages/ui/src/lib/api-client.ts`
  - adds the typed server API call for the export endpoint

## Validation
- `npx tsx --test packages/ui/src/stores/session-message-fallback.test.ts`

## Notes
- `npm run typecheck --workspace @neuralnomads/codenomad` currently fails on the moving `dev` base because of unrelated server dependency/type drift, not because of this change
- touched large files worth future refactor attention:
  - `packages/server/src/workspaces/manager.ts`
  - `packages/ui/src/stores/session-api.ts`
  - `packages/ui/src/stores/session-state.ts`

### DIFF
--- a/packages/server/src/api-types.ts
+++ b/packages/server/src/api-types.ts
@@ -59,6 +59,11 @@ export interface WorkspaceDeleteResponse {
   status: WorkspaceStatus
 }
 
+export interface WorkspaceSessionExportResponse {
+  info: Record<string, unknown>
+  messages: unknown[]
+}
+
 export type WorktreeKind = "root" | "worktree"
 
 export interface WorktreeDescriptor {
--- a/packages/server/src/server/routes/workspaces.ts
+++ b/packages/server/src/server/routes/workspaces.ts
@@ -103,6 +103,16 @@ export function registerWorkspaceRoutes(app: FastifyInstance, deps: RouteDeps) {
   })
 
   app.get<{
+    Params: { id: string; sessionId: string }
+  }>("/api/workspaces/:id/sessions/:sessionId/export", async (request, reply) => {
+    try {
+      return await deps.workspaceManager.exportSessionData(request.params.id, request.params.sessionId)
+    } catch (error) {
+      return handleWorkspaceError(error, reply)
+    }
+  })
+
+  app.get<{
     Params: { id: string }
     Querystring: { path?: string }
   }>("/api/workspaces/:id/files", async (request, reply) => {
--- a/packages/server/src/server/routes/workspaces.ts
+++ b/packages/server/src/server/routes/workspaces.ts
@@ -104,8 +104,8 @@ export function registerWorkspaceRoutes(app: FastifyInstance, deps: RouteDeps) {
   })
 
   app.get<{
-    Params: { id: string; sessionId: string }
-  }>("/api/workspaces/:id/sessions/:sessionId/export", async (request, reply) => {
+    Params: { id: string; slug: string; sessionId: string }
+  }>("/api/workspaces/:id/worktrees/:slug/sessions/:sessionId/export", async (request, reply) => {
     const controller = new AbortController()
     const handleAbort = () => controller.abort()
     request.raw.on("close", handleAbort)
@@ -132,8 +132,28 @@ export function registerWorkspaceRoutes(app: FastifyInstance, deps: RouteDeps) {
     reply.raw.on("finish", handleReplyDone)
 
     try {
+      const workspace = deps.workspaceManager.get(request.params.id)
+      if (!workspace) {
+        await cleanup()
+        reply.code(404)
+        return { error: "Workspace not found" }
+      }
+
+      const directory = await resolveWorktreeDirectory({
+        workspaceId: workspace.id,
+        workspacePath: workspace.path,
+        worktreeSlug: request.params.slug,
+        logger: request.log,
+      })
+      if (!directory) {
+        await cleanup()
+        reply.code(404)
+        return { error: "Worktree not found" }
+      }
+
       exportFile = await deps.workspaceManager.exportSessionDataToFile(request.params.id, request.params.sessionId, {
         signal: controller.signal,
+        directory,
       })
       const stream = createReadStream(exportFile.filePath)
       stream.on("error", () => {
--- a/packages/server/src/server/routes/workspaces.ts
+++ b/packages/server/src/server/routes/workspaces.ts
@@ -1,3 +1,4 @@
+import { createReadStream } from "fs"
 import { FastifyInstance, FastifyReply } from "fastify"
 import { z } from "zod"
 import { WorkspaceManager } from "../../workspaces/manager"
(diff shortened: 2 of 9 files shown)


shantur:
Hey @pascalandr 

> * recover legacy OpenCode sessions that appear in `session.list` but fail transcript hydration because `session.messages` rejects malformed historical assistant metadata

Interesting. Is it failing in OpenCode end or CodeNomad end?

pascalandr:
only in codenomad (any version on windows), opencode have no problem with the malformed metadata

shantur:
If I understand correctly, this tries to export and import the old style session.
If, that's true, I would expect the session hydration to support the old session data instead of export / import

pascalandr:
> If I understand correctly, this tries to export and import the old style session. If, that's true, I would expect the session hydration to support the old session data instead of export / import

right, still working on it, but there's no import, only export because this is the only working way:
The UI cannot hydrate these sessions directly because the OpenCode session.messagesendpoint itself rejects the legacy payload, so this uses opencode export as a read fallback. I have updated the description with lot more details

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
@CodeNomadBot
same review

shantur:
hey @pascalandr 

> 3\. If that call fails with the specific legacy body-validation signature for missing `info.agent`, CodeNomad falls back to `opencode export <sessionId>`.

1. Do we get this error from OpenCode?
2. How does OpenCode work with the same session. You mentioned it works with opencode, is it still the case ?
3. If even OpenCode isn't supporting those old sessions, why should we?

pascalandr:
To clarify: OpenCode itself still reads these sessions fine; the incompatibility is in the SDK used by CodeNomad, where `session.messages` rejects some session payloads that the main OpenCode client still accepts.
I need this fix because a quarter of my used cessions are broken without it in codenomad.
These are not so old sessions; they are sessions whose payload is still accepted by OpenCode itself but rejectedd  by the newer SDK path used by CodeNomad.

shantur:
You maybe highlighting some deeper issue here, OpenCode Desktop uses the same SDK that CodeNomad does.
If it is able to work, it means CodeNomad is
 
a. Either doing something incorrect
b. Using wrong / older version of SDK

I think instead of trying a band-aid fix, we should investigate why it fails in CodeNomad.

Would you be able to 

1. Try loading the same session in OpenCode desktop with Developer Tools to capture the complete request and response.
2. Try loading the same session in CodeNomad and compare the request and response.

pascalandr:
oh I never spoked about Opencode desktop, did you ?

shantur:
TUI also uses the same SDK, IIRC

pascalandr:
well as I said already this cessions open with zero issue or message in opencode cli and doesn't open at all in codnomad. that all I can say, the agent found a solution by testing against the sdk. https://github.com/NeuralNomadsAI/CodeNomad/pull/444#issue-4447075881

shantur:
CLI / TUI also works via the SDK so if it can work, then definitely some issue in CodeNomad.
Agent might not know and is trying one way to make this work.

Would you be able to send me one of your problematic sessions by exporting?
I can figure out the cause and try to fix it properly.

pascalandr:
once the cession is exported it isn't broken anymore in Codenomad. So I agree we don't understand well what's happening here, working on another solution by trying to repair the cessions instead of supporting them.

shantur:
We can understand the issue better if you try to load the session in opencode desktop and keep the dev tools open to capture request / response. Once we have that we can see what different request / response is CodeNomad doing.

A curl request can also be used to query the session messages to understand what's going on.

pascalandr:
ok so I'm going to install opencode desktop

pascalandr:
Error: Missing key
  at [1]["info"]["agent"]
    at wrapClientError (oc://renderer/assets/main-DlaiLIQY.js:62826:12)
    at request (oc://renderer/assets/main-DlaiLIQY.js:59067:28)
    at async retry (oc://renderer/assets/main-DlaiLIQY.js:63575:14)
    at async fetchMessages (oc://renderer/assets/main-DlaiLIQY.js:65394:24)
    at async loadMessages (oc://renderer/assets/main-DlaiLIQY.js:65418:7)
    at async Promise.all (index 1)
    at async oc://renderer/assets/main-DlaiLIQY.js:65576:13



────────────────────────────────────────
Causé par :
{
  "body": {
    "name": "BadRequest",
    "data": {
      "message": "Missing key\n  at [1][\"info\"][\"agent\"]",
      "kind": "Body"
    }
  },
  "status": 400
}

pascalandr:
and still no issue in TUI, so I guess finally the sdk is not used the same way in each of them

pascalandr:
supersed by #450

PR #450, opened 2026-05-15
feat(ui): add OpenCode session repair tool

pascalandr:
## Summary
- add a command-palette-driven OpenCode session repair tool that analyzes the global OpenCode database and reports broken, hidden, and merely incomplete sessions separately
- provide scoped repair actions for safe metadata normalization, directory reconciliation, or both, with an automatic database backup before any write
- surface the analysis and repair flow in a dedicated modal so users no longer need manual SQLite edits to recover affected sessions

## Validation

image

supersed #444


shantur:
@pascalandr - The correct fix for the issue is here https://github.com/anomalyco/opencode/pull/27702

I will get this implemented in OpenCode.

pascalandr:
yes but the tool here is not just about this bug

shantur:
What else does it do?

pascalandr:
Directory mismatch, and metadata malformation in general not specificaly the one I started with in #444

shantur:
TBH, I don't envision such tools to be part of CodeNomad. It maybe better suited as separate utilities.
CodeNomad targets to be a Coding platform, don't want it to be come an opencode manager / toolbox.

pascalandr:
could be in opencode itself, wil try to talk about that with opencode 's team

PR #458, opened 2026-05-16
fix(ui): allow status panel sections to collapse

pascalandr:
## Summary
- Stop the right panel from immediately re-opening collapsed Status sections.
- Keep the existing default expanded sections, including session changes.

## Why
Users could click Status sub-panels to collapse them, but the UI immediately expanded them again, making the accordion controls feel broken.

## Validation
- npm run typecheck --workspace @codenomad/ui
- git diff --check
- npm run build --workspace @codenomad/ui

PR #459, opened 2026-05-16
fix(mobile): tappable instance/project tab bar while session drawer is open

omercnet:
## Summary

Fixes a mobile UX bug where, with the session list (left drawer) open on phone layout, the still-visible instance/project tab bar at the top was non-interactive. Tapping a tab did nothing — users had to close the drawer first and then switch projects/instances.

After this change, tapping a tab in that bar switches the instance/project **and** the drawer auto-closes in a single gesture, matching user expectation.

## Root cause

On phone (`max-width: 767px`), `InstanceShell` renders the left session sidebar via MUI's `Drawer variant="temporary"`. The drawer paper is offset down to `floatingTopPx()` so the instance tab bar remains visually visible above it. However, MUI's `Modal` Backdrop is `position: fixed; inset: 0` and covers the entire viewport — including the area over the tab bar. The backdrop is styled `backgroundColor: transparent` (so it's invisible) but it still captures pointer events at `z-index: 60`. Taps over the tab bar hit the transparent backdrop → MUI calls `onClose` (closing the drawer) but the tab's click handler never fires.

## Implementation

1. **`packages/ui/src/components/instance/instance-shell2.tsx`** — Constrain the MUI Drawer Backdrop via `sx` overrides on both the left (session list) and right drawers so the backdrop is bound to the drawer paper's vertical range (`top: floatingTopPx(); height: floatingHeight();`) instead of fullscreen. Taps over the tab bar now reach the tab buttons.

2. **`packages/ui/src/styles/panels/tabs.css`** — Lift `.tab-bar-instance` to `position: relative; z-index: 70` so it stacks deterministically above the drawer (z-index 60) across browsers as defense-in-depth.

3. **`packages/ui/src/components/instance/shell/useDrawerChrome.ts` + `instance-shell2.tsx`** — Expose `closeFloatingDrawersIfAny` from `useDrawerChrome` and call it from an `InstanceShell` effect that fires when `props.isActiveInstance` flips `true → false`. This closes any open floating drawer on the instance the user just switched away from, so its previously-open state doesn't bleed back when the user returns to that tab later.

Tablet (>=768px) and desktop (>=1280px) layouts use pinned drawers (no temporary modal), so the backdrop constraint and z-index lift are inert there. The fix applies symmetrically to the right drawer on phone.

## Verification

- `tsc --noEmit` clean.
- `vite build` clean.
- Manual mobile-emulation checklist in the task file (`tasks/done/058-mobile-session-list-blocks-tab-switch.md`) and SUMMARY in `evidences/058-mobile-session-list-blocks-tab-switch/`.

## Reviewer manual check (phone viewport <=767px)

- Open session list drawer. Tap a different instance tab → it activates AND drawer closes.
- Tap "+" while drawer open → folder picker opens, drawer closes.
- Tap Settings / Notifications / Remote while drawer open → action fires, drawer closes.
- Same checks against the right drawer.
- Tap area below tab bar but outside drawer paper → drawer still closes (existing backdrop dismissal preserved).
- Resize to tablet and desktop widths → pinned drawers unaffected.
- No visual regression in light or dark theme.

### DIFF
--- a/packages/ui/src/components/instance/instance-shell2.tsx
+++ b/packages/ui/src/components/instance/instance-shell2.tsx
@@ -194,6 +194,7 @@ const InstanceShell2: Component<InstanceShellProps> = (props) => {
     unpinRight: unpinRightDrawer,
     closeLeft: closeLeftDrawer,
     closeRight: closeRightDrawer,
+    closeFloatingDrawersIfAny,
     leftAppBarButtonLabel,
     rightAppBarButtonLabel,
     leftAppBarButtonIcon,
@@ -201,6 +202,19 @@ const InstanceShell2: Component<InstanceShellProps> = (props) => {
     handleLeftAppBarButtonClick,
     handleRightAppBarButtonClick,
   } = drawerChrome
+
+  // When the user switches away from this instance (e.g., taps a different
+  // instance/project tab while a floating drawer is open on phone), close any
+  // open floating drawers so the previous instance's drawer doesn't remain
+  // visually or interactively open when its tab regains focus later.
+  let wasActiveInstance = Boolean(props.isActiveInstance)
+  createEffect(() => {
+    const isActive = Boolean(props.isActiveInstance)
+    if (wasActiveInstance && !isActive) {
+      closeFloatingDrawersIfAny()
+    }
+    wasActiveInstance = isActive
+  })
 
   createEffect(() => {
     const instanceId = props.instance.id
@@ -607,7 +621,13 @@ const InstanceShell2: Component<InstanceShellProps> = (props) => {
         ModalProps={modalProps}
         sx={{
           zIndex: 60,
+          // The outer modal layer is transparent to pointer events; only the
+          // paper and the (constrained) backdrop intercept. This lets taps on
+          // the still-visible instance tab bar above the drawer reach the tab
+          // buttons in a single gesture on phone layout.
+          pointerEvents: "none",
           "& .MuiDrawer-paper": {
+            pointerEvents: "auto",
             width: isPhoneLayout() ? "100vw" : `${sessionSidebarWidth()}px`,
             boxSizing: "border-box",
             borderInlineEnd: isPhoneLayout() ? "none" : "1px solid var(--border-base)",
@@ -620,8 +640,14 @@ const InstanceShell2: Component<InstanceShellProps> = (props) => {
             height: floatingHeight(),
           },
 
+          // Constrain the backdrop so it does not overlap the instance tab bar
+          // above the drawer; backdrop dismissal still works in the remaining
+          // area below the tab bar.
           "& .MuiBackdrop-root": {
+            pointerEvents: "auto",
             backgroundColor: "transparent",
+            top: floatingTopPx(),
+            height: floatingHeight(),
           },
         }}
       >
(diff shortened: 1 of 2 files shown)


shantur:
Hey @omercnet 

Thanks for the PR

1. Seems like you are using NomadWorks for CodeNomad. We haven't got it enabled for the project so can you please keep it disabled for now or don't commit the task and non-code files
2. This forces the tab bar to be visible when new tab button is pressed and it goes under it.
3. In my testing, tapping the tab bar auto closes the drawer in the current `dev` version and the second tab is needed to activate the tab. I understand not the best of UX but does it not work like that for you?

shantur:
@CodeNomadBot

Understand the issue here and also check my comments and suggest a limited fix

omercnet:
Thanks for the review @shantur and @CodeNomadBot — addressed all three points in the force-push (`bfaa4a8` → `f18b155`):

1. **Removed all non-code artifacts.** Dropped `tasks/done.md`, `tasks/done/058-…md`, and `evidences/058-…/SUMMARY.md`. NomadWorks stays out of the PR.
2. **Reverted the `.tab-bar-instance` `position: relative; z-index: 70` change** — that was the regression making the new-tab / folder-selection overlay render under the tab bar. Tab bar default stacking is restored.
3. **Scoped the fix to the temporary drawer only**, exactly as suggested:
   ```ts
   sx={{
     zIndex: 60,
     pointerEvents: "none",
     "& .MuiDrawer-paper": { pointerEvents: "auto", /* …existing… */ },
     "& .MuiBackdrop-root": {
       pointerEvents: "auto",
       backgroundColor: "transparent",
       top: floatingTopPx(),
       height: floatingHeight(),
     },
   }}
   ```
   Applied symmetrically to the right drawer. The outer modal layer is now pointer-transparent, so taps over the still-visible tab bar reach the real tab buttons; the constrained backdrop continues to dismiss the drawer when tapped below the tab bar.

I also rebased onto the latest `dev` (`311b60b`) so this is up to date with `#458` and `#461`.

Kept the small `closeFloatingDrawersIfAny` export from `useDrawerChrome` plus the `isActiveInstance true → false` effect in `InstanceShell` — it closes a previous instance's floating drawer when its tab loses focus, so the drawer doesn't visually bleed back when the user returns to that tab. Happy to drop those two if you'd prefer the absolute minimum patch — let me know.

Final diff: 2 files, 34 insertions.

To answer your direct question about the two-tap behavior on current `dev`: I do see exactly what you describe — first tap dismisses the drawer, second tap then activates the tab. This PR collapses it into a single tap, which matches typical mobile drawer/overlay UX (e.g., tapping outside-content while a sheet is open should both dismiss the sheet and act on the tapped target when that target is a top-level navigation chrome above the sheet).

shantur:
@CodeNomadBot

Check again now

shantur:
Pushed a follow-up commit to the fork: `6fb00ce fix(mobile): close floating drawers from tab bar gaps`.

What changed:
- Kept the drawer paper and backdrop constrained below the instance tab bar, so tab-bar controls remain reachable while a floating drawer is open.
- Added active-instance click-away handling for floating drawers: if a pointer starts outside the left/right drawer content, any open floating drawer closes.
- Updated the comments around the drawer/backdrop styling to match this model.

Why:
The previous scoped fix made the tab bar clickable by moving the backdrop below it, but blank/non-interactive areas of the tab bar no longer behaved like outside clicks, so the drawer stayed open. The new behavior treats the tab bar as outside the drawer: clicking a real tab/action still activates it in the same gesture, while clicking tab-bar gaps now closes the drawer as expected.

shantur:
@CodeNomadBot

Check PR again now

shantur:
@CodeNomadBot

And now?

shantur:
Thanks @omercnet

PR #461, opened 2026-05-16
fix(ui): prevent accidental prompt clear from tab focus

### LINKED ISSUE #460: Pressing Tab focus the clear button and Space deletes the whole prompt with no undo
While typing in the input field, I sometimes accidentally press `Tab`.  
This moves focus to the clear (`X`) button next to the text field. If I do not notice and then press `Space`, it activates the button and clears the entire text.

The main issue is that this action does not seem to be recoverable with `Ctrl+Z`, so it is very easy to lose everything by mistake.

**Steps to reproduce**
1. Type some text in the input field.
2. Press `Tab`.
3. Press `Space`.

**Current behavior**  
After pressing `Tab`, the clear (`X`) button gets focus.  
Pressing `Space` then clears the whole input.  
The deleted text cannot be restored with `Ctrl+Z`.

**Expected behavior**  
Accidentally pressing `Tab` and then `Space` should not result in losing the entire input with no way to recover it.

**Possible fixes**
1. Make the clear action part of the undo history so `Ctrl+Z` restores the deleted text.
2. Change the tab order so the clear button is not the next focused element after the text input.

shantur:
## Summary
- Fixes #460 by changing the prompt control DOM order so tabbing from the textarea reaches expand before clear.
- Clears prompt text through the textarea native delete path so Cmd/Ctrl+Z can restore cleared text where the browser supports native undo.
- Keeps the existing app-state fallback for environments where native editing commands are unavailable or throw.

## Validation
- Reviewed fallback behavior: `document.execCommand` is checked for callability and wrapped in `try/catch`, so unavailable native editing commands still fall back to app-state clearing.

## Manual QA Notes
- In a supported desktop/browser host: type text, click clear, then press Cmd/Ctrl+Z and confirm text restores.
- Type text, press Tab and confirm focus moves to expand before clear; pressing Tab again reaches clear.

### DIFF
--- a/packages/ui/src/components/prompt-input.tsx
+++ b/packages/ui/src/components/prompt-input.tsx
@@ -463,8 +463,26 @@ export default function PromptInput(props: PromptInputProps) {
     textareaRef?.focus()
   }
 
+  function clearTextareaWithUndo() {
+    const textarea = textareaRef
+    if (!textarea || textarea.disabled) return false
+
+    textarea.focus()
+    textarea.setSelectionRange(0, textarea.value.length)
+
+    const cleared = typeof document !== "undefined" && document.execCommand("delete")
+    if (!cleared || textarea.value.length > 0) {
+      textarea.value = ""
+    }
+
+    textarea.dispatchEvent(new Event("input", { bubbles: true }))
+    return cleared
+  }
+
   function handleClearPrompt() {
-    clearPrompt()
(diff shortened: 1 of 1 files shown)


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

Check again

shantur:
@CodeNomadBot

Now?

PR #472, opened 2026-05-17
revert: remove tracked NomadWorks repo state

pascalandr:
## Summary
- revert the NomadWorks initialization artifacts that introduced tracked `.nomadworks` state, root `codemap.yml`, SCR registry files, and task registry files
- keep the tracked `.opencode` project configuration/commands/agent files because they are part of the expected CodeNomad repo setup
- preserve unrelated product commits while cleaning up only the NomadWorks-generated repo state


shantur:
Thanks @pascalandr 

But we need `.opencode` stuff

pascalandr:
@shantur 
do we need  .opencode/agent/web_developer.md  ?

shantur:
Yes, I use it when I have to use subagents for a bigger feature

shantur (approved):

shantur:
Thanks @pascalandr

PR #479, opened 2026-05-18
fix(ui): preserve question queue order when upserting duplicate requests

### LINKED ISSUE #448: [Bug]: questions do not open and garble leading to deadlock "Waiting for earlier responses"

JDis03:
## Summary
- Fixes #448 - Questions deadlock with "Waiting for earlier responses" when a question arrives first as a global entry and later resolves to a tool part with a newer timestamp.

## Root Cause
`upsertQuestion` replaced the entire queue entry on duplicate request id, losing the original `enqueuedAt` value. If question-1 arrived at t=1000 (global), then question-2 at t=1500, and question-1 resolved to a tool part at t=2000, the newer `enqueuedAt` (2000) would place question-1 behind question-2 in the sorted queue, breaking active interruption ordering.

## Fix
- Add `mergeQuestionEntry` mirroring `mergePermissionEntry` behavior
- Preserve original `enqueuedAt` using `Math.min(existing, new)`
- Keep newer `messageId`/`partId` when resolving from global to tool part
- Clean up stale `byMessage` entries for the same request id
- Sort queue by `enqueuedAt` and recalculate active after each upsert

## Validation
- `node --import tsx --test packages/ui/src/stores/message-v2/instance-store.test.ts`
- Added 3 regression tests covering duplicate resolution and ordering

### DIFF
--- a/packages/ui/src/stores/message-v2/instance-store.test.ts
+++ b/packages/ui/src/stores/message-v2/instance-store.test.ts
@@ -35,3 +35,60 @@ describe("message-v2 permission state", () => {
     assert.equal(store.getPermissionState(undefined, "permission-2")?.active, true)
   })
 })
+
+describe("message-v2 question state", () => {
+  it("keeps one question attachment when a duplicate moves from global to a tool part", () => {
+    const store = createInstanceMessageStore("instance-1")
+
+    store.upsertQuestion({
+      request: { id: "question-1", sessionID: "session-1", questions: [] },
+      enqueuedAt: 1_000,
+    })
+    store.upsertQuestion({
+      request: { id: "question-1", sessionID: "session-1", questions: [] },
+      messageId: "message-1",
+      partId: "part-1",
+      enqueuedAt: 2_000,
+    })
+
+    assert.equal(store.state.questions.queue.length, 1)
+    assert.equal(store.getQuestionState(undefined, "question-1"), null)
+    assert.equal(store.getQuestionState("message-1", "part-1")?.entry.request.id, "question-1")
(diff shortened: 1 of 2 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
I’m not sure this fully fixes #448.
This PR fixes a narrow question upsert/order case, but the issue discussion points to a broader problem: the blocking interruption is not visible/actionable, and the user gets stuck on “Waiting for earlier responses”.
This PR does not touch `activeInterruption`, the permission/question center, auto-opening behavior, or any way to act on the blocking item.
Reproduction validation from the reporter is needed.

pascalandr:
Closing as superseded by #576, which already includes the generalized question deduplication and enqueue-time preservation. #448 remains open for the broader unresolved deadlock symptoms.

PR #483, opened 2026-05-18
feat(ui): add retry button for failed tool calls

### LINKED ISSUE #473: Is it possible to retry when tool fails?
it happened to me that I reached my subscription limit to one provider. But I have subscription to another provider I could switch to in order to continue. However I cant figure out how. UI tells me to wait although I tried to change LLM 

Image

JDis03:
## Summary
Fixes #473. When a tool call fails (rate limit, provider error, etc.), users can now retry by reverting to the original user message.

## How it works
- A **Retry** action appears in the tool call overflow menu when the tool state is `error`.
- Clicking retry calls the existing revert flow, which finds the user message before the failed tool call, reverts all messages after it, and restores the original prompt text in the input field.
- The user can then change the model, edit the prompt, and resend.

## Files changed
- `packages/ui/src/components/message-block.tsx`: Added `onRevert` prop to `ToolCallItem`, `handleRetry` function, and retry menu item in `actionMenuItems()`.
- All locale `toolCall.ts` files: Added `toolCall.retry.label` and `toolCall.retry.title` strings (en, es, fr, ja, ru, he, zh-Hans).

## Validation

### DIFF
--- a/packages/ui/src/components/message-block.tsx
+++ b/packages/ui/src/components/message-block.tsx
@@ -1,5 +1,5 @@
 import { For, Index, Match, Show, Suspense, Switch, createEffect, createMemo, createSignal, lazy, onCleanup, untrack, type Accessor } from "solid-js"
-import { ChevronsDownUp, ChevronsUpDown, ExternalLink, FoldVertical, ListStart, Trash, Volume2 } from "lucide-solid"
+import { ChevronsDownUp, ChevronsUpDown, ExternalLink, FoldVertical, ListStart, RefreshCw, Trash, Volume2 } from "lucide-solid"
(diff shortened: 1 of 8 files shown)


shantur:
Hey @JDis03 ,

Revert to last user message is already possible.
Giving user an option to retry failed tool call and reverting to last message can be destructive. OpenCode reverts changes on disk too by keeping snapshots. 
From a user point of view retrying a tool call would mean actually just retry the tool call. I don't think this will be a good UX and can be confusing.

Thanks

shantur:
Closing as per comment above, if you feel it shouldn't, please re-open

PR #512, opened 2026-05-30
fix: handle plugin base URL for host binding

omercnet:
## Summary
- Resolve the plugin fetch/SSE base URL from actual listener bindings instead of always publishing 127.0.0.1.
- Preserve loopback plugin loading for default local, localhost, wildcard, and mixed HTTP/HTTPS listener setups.
- Make the git clone destination test portable so server tests run with zero skips.

## Validation
- node --import tsx --test packages/server/src/server/__tests__/listener-base-url.test.ts packages/server/src/server/__tests__/network-addresses.test.ts packages/server/src/workspaces/__tests__/git-clone.test.ts — 15 pass / 0 fail / 0 skipped
- shopt -s globstar && node --import tsx --test packages/server/src/**/*.test.ts — 59 pass / 0 fail / 0 skipped

## Notes
- Nomad task/evidence artifacts intentionally excluded from the commit per maintainer request.

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -23,6 +23,7 @@ import { AuthManager, BOOTSTRAP_TOKEN_STDOUT_PREFIX, DEFAULT_AUTH_COOKIE_NAME, D
 import { resolveHttpsOptions } from "./server/tls"
 import { RemoteProxySessionManager } from "./server/remote-proxy"
 import { resolveNetworkAddresses, resolveRemoteAddresses } from "./server/network-addresses"
+import { resolvePluginBaseUrl } from "./server/listener-base-url"
 import { startDevReleaseMonitor } from "./releases/dev-release-monitor"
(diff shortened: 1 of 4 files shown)


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur (approved):

shantur:
Hey @omercnet 

Thanks for the PR
Can you explain what was failing?
IIRC the codenomad plugin is fetched via file:// not http url and we don't manage other plugins

omercnet:
I noticed an error in opencode logs, seems that when host is customized (in my case to a tailscale ip) it's not load via file but defaults to https://127.0.0.1:9898

shantur:
@CodeNomadBot

Can you confirm what url the codenomad sets in the inline opencode variable for loading codenomad plugin from packages/codenomad-plugin packages/server

PR #515, opened 2026-06-01
fix: port permission idempotency model to question replies

omercnet:
## Problem

Answering a `question` prompt after a reconnect, process restart, or rapid double-submit could POST a reply for a request the backend no longer tracks, producing opencode's `reply for unknown request` warning. The reply was either silently dropped or mis-routed to a fallback (`root`) target.

Root cause: the `question` request lifecycle in the UI lacked the idempotency and stale-guard layer that the `permission` lifecycle already had. It reproduces even with a single root worktree (not primarily a worktree issue).

## Fix

Ports the existing permission idempotency model to questions (entirely `packages/ui`, no opencode core changes):

- **New `stores/question-replies.ts` ledger** (`markQuestionReplied` / `hasRepliedQuestion` / `pruneRepliedQuestions` / `clearRepliedQuestions`) — a structural mirror of `permission-replies.ts`. No shared abstraction (YAGNI; prune semantics may diverge).
- **Reconcile-before-POST**: `sendQuestionReply` / `sendQuestionReject` check `question.list()` before POSTing. If the request is no longer pending, the UI surfaces a localized "this prompt expired — send your answer as a message instead" notice rather than POSTing to a fallback. No error-string parsing.
- **`syncPendingQuestions`** adopts the full timestamp-based prune + `hasRepliedQuestion` filter so replied questions are never rehydrated on reconnect.
- **Ledger retired only via timestamp prune** — never cleared on `rehydrateInstance` (only on instance removal), so in-flight stale replies can't slip through.
- **Stale guards** on `handleQuestionAsked` / `handleQuestionAnswered`.
- **Synchronous double-submit guard** at the top of `handleQuestionSubmit` (Enter + button can no longer both fire).
- New i18n key `toolCall.question.errors.expired` added to all 7 locales.

Also fixes two pre-existing module-load crashes uncovered while making the UI suite green: `client-identity.ts` dereferenced `window` storage before its `typeof window` guard, and `server-events.ts` opened an `EventSource` at module load without a guard. Both minimal and production-safe.

## User-visible behavior

- Expired question prompts show a clear "answer as a message" notice instead of failing silently or mis-routing.
- Rapid double-submit can no longer fire two replies.
- Reconnect no longer re-surfaces an already-answered question.

## Validation

- UI suite **49/49 pass** (`bun test --conditions browser`)
- UI typecheck clean (`tsc --noEmit`)
- Rebased on `origin/dev` (upstream default branch)

### DIFF
--- a/packages/ui/src/components/tool-call.tsx
+++ b/packages/ui/src/components/tool-call.tsx
@@ -5,7 +5,14 @@ import { messageStoreBus } from "../stores/message-v2/bus"
 import { useTheme } from "../lib/theme"
 import { useGlobalCache } from "../lib/hooks/use-global-cache"
 import { useConfig } from "../stores/preferences"
-import { activeInterruption, sendPermissionResponse, sendQuestionReject, sendQuestionReply } from "../stores/instances"
+import {
+  activeInterruption,
+  hasRepliedQuestion,
+  QuestionExpiredError,
+  sendPermissionResponse,
+  sendQuestionReject,
+  sendQuestionReply,
+} from "../stores/instances"
 import { copyToClipboard } from "../lib/clipboard"
 import type { PermissionRequestLike } from "../types/permission"
 import { getPermissionSessionId } from "../types/permission"
@@ -275,8 +282,16 @@ function ToolCallDetails(props: {
   }
 
   async function handleQuestionSubmit() {
+    // Synchronous double-submit guard BEFORE any await: Enter + button (or a
+    // rapid re-submit) must never fire two `question.reply` calls.
+    if (questionSubmitting()) return
     const request = questionDetails()
     if (!request || !props.isQuestionActive()) {
+      return
+    }
+    // Second defense: a stale request already replied locally (e.g. an SSE
+    // `question.replied` raced ahead) is a no-op.
+    if (hasRepliedQuestion(props.instanceId, request.id)) {
       return
     }
     const answers = (questionDraftAnswers()[request.id] ?? []).map((x) => (Array.isArray(x) ? x : []))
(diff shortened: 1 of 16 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
Until we know how to reproduce this, we can't be sure this fixes issues.
@omercnet - Would you be able to help us reproduce the issue first?

omercnet:
yes, happy to asisst - it happens often for me but now sure why

pascalandr:
@omercnet Thanks for the work on this. The underlying stale/duplicate question-reply problem remains plausible, but we should close this PR rather than merge or rebase it.

Since this was opened, PRs #521 and #526 substantially changed question routing, workspace handling, and the legacy/V2 question APIs. The implementation here now targets an obsolete lifecycle and conflicts with current dev.

More importantly, the list-before-reply approach introduces a correctness risk: if every question.list() probe fails transiently, the code treats the question as expired, removes it, and discards the user's answer. A successful preflight list would still leave a race between the list and the reply.

We also still lack a reliable reproduction showing the exact trigger and user impact. Without that, rebasing this 489-line change would carry significant integration risk without proving that it fixes the current behavior.

If the issue is reproduced on current dev, the safer path is a new focused PR using the current legacy/V2 routing:

- deduplicate in-flight replies by instance/request ID;
- preserve the prompt on transport failures;
- remove it only after confirmed success or a typed terminal response;
- optionally use a small tombstone only to suppress delayed SSE rehydration;
- test legacy, V2, double-submit and delayed-event paths directly.

PR #516 is stacked on this commit and should be rebuilt directly on current dev, excluding f7618722.

Closing this PR does not mean the reported issue is invalid. It means the implementation is now too stale and risky to merge safely, and any future fix should start from a confirmed reproduction against the current architecture.

PR #514, opened 2026-05-31
Migrate worktree mappings to session metadata

shantur:
## Summary
- Upgrade `@opencode-ai/sdk` to `1.15.13` and adapt UI types for the current session/diff API surface.
- Add a session metadata store helper that safely updates OpenCode session metadata via read-merge-full-replace while preserving non-CodeNomad metadata.
- Migrate CodeNomad worktree assignments from `.codenomad/worktreeMap.json` into `metadata.codenomad.worktreeSlug`, with legacy fallback, stale-entry pruning, and server-side deletion of the legacy map once empty.
- Add `.codenomad/background_processes/` to CodeNomad-managed `.git/info/exclude` entries.

## Details
- New sessions and worktree reassignment now persist explicit session worktree assignments in OpenCode session metadata under:
  `metadata.codenomad.worktreeSlug`
- Worktree resolution now prefers session metadata, then falls back to legacy `parentSessionWorktreeSlug`, then `root`.
- Legacy migration runs after session/map loading; after the full session list is known, missing legacy session IDs are treated as stale and pruned.
- The UI no longer uses `defaultWorktreeSlug`; the fallback is `root`.
- `writeWorktreeMap` deletes `.codenomad/worktreeMap.json` server-side when no legacy parent-session mappings remain.

## Validation
- `npm run typecheck --workspace @codenomad/ui`
- `npm run typecheck --workspace @neuralnomads/codenomad`

PR #516, opened 2026-06-01
fix(ui): refresh + auto-switch worktrees on session idle

omercnet:
> **Stacked on #515** (task 058 question-reply idempotency). This PR's diff currently includes the 058 commit because it carries a baseline test fix (`client-identity.ts` / `server-events.ts` window/EventSource guards) required for the UI suite to be green. **Merge #515 first**; this PR's net change is the single commit `fix(ui): refresh + auto-switch worktrees on session idle`.

## Problem

The opencode agent creates worktrees via `git worktree add` mid-session (outside CodeNomad's managed flow). CodeNomad cached its worktree list and never invalidated it for out-of-band git changes, so the files view stayed bound to a stale tree until a manual reload or restart.

## Fix

`packages/ui` only — no server cache (server enumeration via `git worktree list --porcelain` is already live per request):

- **New debounced `refreshWorktreesOnIdle(instanceId)`** in `worktrees.ts`, triggered from the existing `handleSessionIdle` path (no new sse-manager hook). Per-instance ~600ms trailing debounce, deduped on parent session id. Adds a `reloadLoads` in-flight guard to `reloadWorktrees` (it had none) to avoid overlap with `rehydrateInstance`.
- **Constrained auto-switch rule** (PO-confirmed): on idle, diff the worktree list before/after reload. If **exactly one** new worktree appeared, auto-switch the active session's files view to it via `setWorktreeSlugForParentSession` (metadata-backed, per #514). Guarded: switch only when the idle session is the active session AND the user hasn't manually switched in the meantime. Zero or multiple new worktrees → refresh list only, no switch (ambiguous attribution is not reliable without a per-session signal).

## Why the constraint

CodeNomad cannot reliably attribute a newly-created worktree to a specific session (no per-session cwd signal). The single-new-worktree heuristic covers the common case (agent creates one worktree mid-turn) safely without yanking the user's view.

## Validation

- UI suite **56/56 pass** (`bun test --conditions browser`) — 49 existing + 7 new
- UI typecheck clean (`tsc --noEmit`), build OK
- Tests cover: reload-on-idle, debounce coalescing, in-flight guard, single-new→switch, zero/multi→no-switch, different-active-session→no-switch

### DIFF
--- a/packages/ui/src/components/tool-call.tsx
+++ b/packages/ui/src/components/tool-call.tsx
@@ -5,7 +5,14 @@ import { messageStoreBus } from "../stores/message-v2/bus"
 import { useTheme } from "../lib/theme"
 import { useGlobalCache } from "../lib/hooks/use-global-cache"
 import { useConfig } from "../stores/preferences"
-import { activeInterruption, sendPermissionResponse, sendQuestionReject, sendQuestionReply } from "../stores/instances"
+import {
+  activeInterruption,
+  hasRepliedQuestion,
+  QuestionExpiredError,
+  sendPermissionResponse,
+  sendQuestionReject,
+  sendQuestionReply,
+} from "../stores/instances"
 import { copyToClipboard } from "../lib/clipboard"
 import type { PermissionRequestLike } from "../types/permission"
 import { getPermissionSessionId } from "../types/permission"
@@ -275,8 +282,16 @@ function ToolCallDetails(props: {
   }
 
   async function handleQuestionSubmit() {
+    // Synchronous double-submit guard BEFORE any await: Enter + button (or a
+    // rapid re-submit) must never fire two `question.reply` calls.
+    if (questionSubmitting()) return
     const request = questionDetails()
     if (!request || !props.isQuestionActive()) {
+      return
+    }
+    // Second defense: a stale request already replied locally (e.g. an SSE
+    // `question.replied` raced ahead) is a no-op.
+    if (hasRepliedQuestion(props.instanceId, request.id)) {
       return
     }
     const answers = (questionDraftAnswers()[request.id] ?? []).map((x) => (Array.isArray(x) ? x : []))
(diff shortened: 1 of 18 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
Hey @omercnet,

Git worktrees were recently improved in CodeNomad + OpenCode.
Instead of refreshing it everytime, as its an agent, maybe we should first check if opencode provides any SSE event when it detects a new worktree. Can you check in opencode source code please

pascalandr:
@omercnet Closing this PR because its implementation is no longer suitable for the current dev architecture.

The underlying stale-inventory issue remains valid, but the SSE approach suggested by @shantur is now available. Current OpenCode emits a `worktree.ready` event after an OpenCode-managed worktree has been created and bootstrapped. Its global event envelope includes the exact worktree `directory`, while the payload includes its `name` and optional `branch`.

CodeNomad's existing instance event bridge already forwards arbitrary OpenCode events and preserves that global `directory`, so polling on every `session.idle` event is unnecessary. A clean replacement based on current `dev` is available in #583. It routes `worktree.ready` through the UI SSE manager, reloads the Git worktree inventory, and then synchronizes OpenCode workspace mappings.

The replacement deliberately does not auto-switch a session. Although the event identifies the newly created worktree, it does not contain a session ID and therefore cannot reliably identify the originating session. Refreshing the inventory is deterministic; assigning the worktree to whichever session happens to be active would retain the same unsafe attribution problem as this PR.

This branch was also stacked on #515, which is now closed, and carries assumptions that predate the current recursive session hierarchy, root-client routing, and workspace mapping paths. Those issues, together with the idle debounce and in-flight reload races identified during review, make this version unsuitable to revive.

Thank you for identifying and implementing the stale-inventory path. #583 supersedes this implementation with the event-driven approach.

PR #521, opened 2026-06-02
fix(worktrees): route OpenCode calls through workspaces

### LINKED ISSUE #518: [Bug]: Losing sessions on codenomad worktrees on restarts
### App Variant

Tauri

Windows 11

### Issue Summary

I keep losing sessions (they just dissapear) on codenomad worktrees when I restart codenomad

### Steps to Reproduce

This issue started appearing right after Opencode introduced the session-related changes.

I can still see the affected sessions in the TUI, but in CodeNomad they frequently disappear after a restart.

For example:

* I create a worktree and start a session in it. That session remains visible.
* Later, I create a second session in the same worktree.
* If I restart CodeNomad or log in through the web interface, the second session sometimes disappears completely.

The behavior is inconsistent. Sometimes it happens, sometimes it doesn't, and I haven't found a reliable way to reproduce it. However, one thing I've noticed is that if I connect through the web interface and a session is missing there, then after restarting the Tauri CodeNomad application that same session is guaranteed to be gone locally as well.

I've also noticed that this only seems to affect sessions created in worktrees. So far, I have not been able to reproduce the issue with sessions created on the main branch.

I'm using the Tauri version locally, in case that helps with troubleshooting.

Image

In that image, the "TRABAJANDO" session is a new recreated one . If i open the TUI i see this:

Image

Many sessions lost

### Expected Behavior

Just don't lose sessions

### Logs & Screenshots

_No response_

### Additional Context

_No response_

shantur:
## Summary

This fixes #518 by moving CodeNomad worktree execution from directory-header routing to OpenCode experimental workspace routing.

OpenCode changed existing-session routing so session routes can prefer the session's stored directory over `x-opencode-directory`. That made the old CodeNomad worktree proxy model unreliable for sessions that should execute in a worktree. This PR switches CodeNomad to resolve an OpenCode workspace ID for each CodeNomad worktree and pass that workspace ID on worktree-scoped OpenCode calls.

## What changed

- Start OpenCode with `OPENCODE_EXPERIMENTAL_WORKSPACES=true`.
- Change the OpenCode server base URL from `/workspaces/:id/worktrees/root/instance` to `/workspaces/:id/instance`.
- Add a root OpenCode client helper in `packages/ui/src/stores/opencode-client.ts`.
- Add OpenCode workspace sync/cache helpers in `packages/ui/src/stores/opencode-workspaces.ts`.
- Sync OpenCode workspaces after CodeNomad worktree hydration and after worktree create/delete flows.
- Map CodeNomad worktree slugs/directories to OpenCode `workspace.id` values discovered by `experimental.workspace.syncList` and `experimental.workspace.list`.
- Replace all OpenCode worktree clients with the root client plus explicit `workspace` payloads where the active session/worktree requires it.
- Route session, permission/question, file browser reads, SDK git status, and prompt/action calls through root OpenCode client + workspace ID.
- Keep CodeNomad local git worktree server APIs intact; those still need filesystem directories for local git operations.

## Review fix

- Fixed right-panel file saves so they no longer write to the root workspace after reading from a selected worktree.
- The existing CodeNomad file-content API now accepts an optional `worktree` query parameter.
- Right-panel saves pass `worktreeSlugForViewer()`, and the server resolves that slug to the same worktree directory used by local git worktree APIs before writing.
- Root saves continue to use the original root workspace path.

## Removed old routing

- Removed `/workspaces/:id/worktrees/:slug/instance` OpenCode proxy routes.
- Removed directory override proxy support using `/__dir/<encoded>`.
- Removed proxy injection of `x-opencode-directory`.
- Removed the remaining background-process completion prompt `x-opencode-directory` header.
- Removed `getOrCreateWorktreeClient`, `getOrCreateWorktreeClientWithDirectoryOverride`, and worktree proxy path helpers from the UI worktree store.
- Simplified `sdkManager.createClient` because clients are no longer keyed by worktree slug.

## Why this fixes #518

Worktree sessions can now stay visible under the root project session listing while worktree execution is selected through OpenCode's workspace routing model. CodeNomad no longer depends on `x-opencode-directory` to override existing session directories, so sessions should not disappear from the root-directory list or execute in the wrong directory because of stale session directory fallback behavior.

Fixes #518

## Validation

- `git diff --check`

## Notes

Some touched files are already oversized and were not refactored as part of this migration: `packages/server/src/server/http-server.ts`, `packages/ui/src/stores/instances.ts`, `packages/ui/src/stores/session-api.ts`, `packages/ui/src/stores/session-state.ts`, `packages/ui/src/components/instance/shell/right-panel/RightPanel.tsx`, `packages/ui/src/stores/session-events.ts`, and `packages/server/src/background-processes/manager.ts`.

### DIFF
--- a/packages/server/src/background-processes/manager.ts
+++ b/packages/server/src/background-processes/manager.ts
@@ -637,7 +637,6 @@ export class BackgroundProcessManager {
     const targetUrl = `http://127.0.0.1:${port}/session/${encodeURIComponent(notify.sessionID)}/prompt_async`
     const headers: Record<string, string> = {
       "content-type": "application/json",
-      "x-opencode-directory": /[^\x00-\x7F]/.test(notify.directory) ? encodeURIComponent(notify.directory) : notify.directory,
     }
 
     const authorization = this.deps.workspaceManager.getInstanceAuthorizationHeader(workspaceId)
--- a/packages/server/src/server/http-server.ts
+++ b/packages/server/src/server/http-server.ts
@@ -9,8 +9,6 @@ import { connect as connectTls, type TLSSocket } from "tls"
 import { fetch, type Headers } from "undici"
 import type { Logger } from "../logger"
 import { WorkspaceManager } from "../workspaces/manager"
-import { isValidWorktreeSlug, listWorktrees, resolveRepoRoot } from "../workspaces/git-worktrees"
-import { resolveWorktreeDirectory } from "../workspaces/worktree-directory"
 
 import type { SettingsService } from "../settings/service"
 import { FileSystemBrowser } from "../filesystem/browser"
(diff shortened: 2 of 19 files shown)


shantur:
@CodeNomadBot

check for regressions
check if the PR could be better implemented in any other way
Be the PR gatekeeper

shantur:
@CodeNomadBot

PR is changed completely go through it again.

check for regressions
check if the PR could be better implemented in any other way
Be the PR gatekeeper

shantur:
@CodeNomadBot

Check now

Issue #525, opened 2026-06-05
[Bug]: CodeNomad doesn't work correctly with OpenCode version 1.16.0+


shantur:
### App Variant

Electron

### Operating System & Version

All

### Issue Summary

SDK change in 1.16.0 is breaking CodeNomad

### Steps to Reproduce

1. Upgrade OpenCode to 1.16.0+
2. CodeNomad doesn't work properly

### Expected Behavior

CodeNomad to work properly

### Additional Context

Workaround : Downgrade opencode to 1.15.3

Aersaud:
Seems to be working fine with opencode v1.16.2.

PR #526, opened 2026-06-06
Support OpenCode SDK 1.16 runtime APIs

shantur:
## Summary
- upgrade the UI OpenCode SDK usage for the 1.16 v2 session list/response shapes
- migrate session list/search handling to v2 while preserving unsupported legacy session actions
- support mixed legacy/v2 permission and question queues, events, hydration, and reply routing during the 1.16 transition

## Validation
- npm run typecheck --workspace packages/ui
- npx tsx --test "packages/ui/src/types/permission.test.ts" "packages/ui/src/stores/message-v2/instance-store.test.ts"

PR #527, opened 2026-06-07
fix: update plugin for OpenCode 1.16.0+ compatibility and add schema migration

High-cla:
## Summary

Fixes two compatibility issues between CodeNomad and OpenCode v1.16.0+:

### Issue #525 — Plugin Compatibility with OpenCode 1.16.0+

**Changes:**
- `packages/opencode-plugin/package.json`: Bumped `@opencode-ai/plugin` from `1.3.7` to `^1.16.0` and added the `"oc-plugin"` field for npm plugin discovery
- `packages/opencode-plugin/plugin/codenomad.ts`: Updated to the new `Plugin` type from `@opencode-ai/plugin`, replaced the old `chat.message` hook with `experimental.chat.system.transform`, and added `export const server: Plugin` for the PluginModule format (required by the v1.16+ loader)
- `packages/opencode-plugin/plugin/lib/background-process.ts`: Changed `tool` import from `@opencode-ai/plugin/tool` to `@opencode-ai/plugin` (tool is now re-exported from the main entry)

### Issue #31204 — `session_message.seq` NOT NULL Constraint Failure

**Changes:**
- `packages/server/src/workspaces/migration.ts` (new): Non-blocking schema migration utility that checks the OpenCode database and fixes the `seq` column if needed
- `packages/server/src/workspaces/runtime.ts`: Calls `checkAndFixOpencodeSchema()` before spawning OpenCode instances

The migration handles three cases:
1. **Column missing** → `ALTER TABLE ADD COLUMN seq INTEGER NOT NULL DEFAULT 0`
2. **Column exists, no default** → Full table recreation via SQL
3. **Already has DEFAULT 0** → No-op

### Testing
- `tsc --noEmit` on both packages: Zero errors
- Integration tests added for migration module (4 tests, all passing)
- All existing tests continue to pass
- Uses `opencode db` command (no external dependencies)

### DIFF
--- a/packages/opencode-plugin/package.json
+++ b/packages/opencode-plugin/package.json
@@ -5,6 +5,7 @@
   "license": "MIT",
   "type": "module",
   "main": "dist/codenomad.js",
+  "oc-plugin": "dist/codenomad.js",
   "files": [
     "dist",
     "README.md"
@@ -13,7 +14,7 @@
     "build": "node -e \"require('fs').rmSync('dist',{recursive:true,force:true})\" && tsc -p tsconfig.json"
   },
   "dependencies": {
-    "@opencode-ai/plugin": "1.3.7"
+    "@opencode-ai/plugin": "^1.16.0"
   },
   "devDependencies": {
     "@types/node": "^22.18.0",
--- a/packages/opencode-plugin/plugin/codenomad.ts
+++ b/packages/opencode-plugin/plugin/codenomad.ts
@@ -1,14 +1,10 @@
-import type { PluginInput } from "@opencode-ai/plugin"
+import type { Plugin } from "@opencode-ai/plugin"
 import { createCodeNomadClient, getCodeNomadConfig } from "./lib/client.js"
 import { createBackgroundProcessTools } from "./lib/background-process.js"
 
 let voiceModeEnabled = false
 
-export async function CodeNomadPlugin(input: PluginInput): Promise<{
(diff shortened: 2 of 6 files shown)


shantur:
Hi @High-cla 

Thanks for the PR. I have made some fixes in dev for the 1.16+ support. Can you check if that works for you

pascalandr:
Thanks @High-cla 
We don’t think the DB/schema migration part should live in CodeNomad. OpenCode owns that database, and anomalyco/opencode#32093 addresses this at the opencode layer with `db doctor` / `db repair`, dry-run/apply separation, backups, transactions, and post-checks. CodeNomad should not silently mutate the OpenCode DB on workspace launch.

For the OpenCode 1.16+ compatibility concern, I’m not sure there is still an issue to fix here. Issue #525 appears to have been addressed by #526 (“Support OpenCode SDK 1.16 runtime APIs”), and there is also feedback on #525 saying OpenCode v1.16.2 works fine.

Before keeping any plugin compatibility changes, could you provide a fresh repro/evidence for what is still broken on current `dev` with OpenCode 1.16.2+? For example: plugin failing to load, tools missing, voice-mode hook not running, or a specific error/log.

If there is still a real plugin-specific regression, I think you should narrow the PR to it.

shantur:
Also as per my investigation, there aren't any changes needed to plugin too.

shantur:
Please re-open this if you feel otherwise.

PR #528, opened 2026-06-07
fix(ui): scope primary agent selector to selectable agents

### LINKED ISSUE #384: [Bug] Fail to Load Agent Properly When Opening

pascalandr:
## Summary

- rescope the UI portion of #385 to the remaining main-session selector issue from #384
- keep main-session primary selectors limited to selectable primary agents
- preserve child/subagent behavior by keeping the current hidden agent available for steering
- add focused tests for selector eligibility behavior

This intentionally excludes the OMO/config-directory merge and plugin retry scope from #385, since that part was determined to belong outside CodeNomad. Credit to @jollyxenon for the original #385 investigation and UI direction.

## Validation

- ode --test packages/ui/src/types/session.test.ts ✅
- pm run typecheck --workspace @codenomad/ui ⚠️ fails on existing SDK/type drift outside this diff

Fixes #384
Based on #385

### DIFF
--- a/packages/ui/src/components/agent-selector.tsx
+++ b/packages/ui/src/components/agent-selector.tsx
@@ -2,7 +2,7 @@ import { Select } from "@kobalte/core/select"
 import { Show, createEffect, createMemo } from "solid-js"
 import { agents, fetchAgents, sessions } from "../stores/sessions"
 import { ChevronDown } from "lucide-solid"
-import { isSelectablePrimaryAgent, type Agent } from "../types/session"
+import { getSelectableAgentsForSession, type Agent } from "../types/session"
 import { useI18n } from "../lib/i18n"
 import { getLogger } from "../lib/logger"
 const log = getLogger("session")
@@ -29,12 +29,7 @@ export default function AgentSelector(props: AgentSelectorProps) {
   })
 
   const availableAgents = createMemo(() => {
-    const allAgents = instanceAgents()
-    if (isChildSession()) {
-      return allAgents.filter((agent) => !agent.hidden)
-    }
-
-    return allAgents.filter(isSelectablePrimaryAgent)
+    return getSelectableAgentsForSession(instanceAgents(), props.currentAgent, isChildSession())
   })
 
   createEffect(() => {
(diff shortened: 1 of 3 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

PR #533, opened 2026-06-08
fix: stabilize hold-mode assistant autoscroll

pascalandr:
## Summary
- Keep Hold mode latched until the user manually reaches the true bottom.
- Restrict Hold targets to user-readable Assistant answer text.
- Exclude tool/status/output and reflection/reasoning blocks from Hold target eligibility.

## Validation
- node --test packages/ui/src/components/virtual-follow-behavior.test.ts


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr:
Gatekeeper blocker from https://github.com/NeuralNomadsAI/CodeNomad/pull/533#issuecomment-4653040748 addressed in c236b7f7. The bottom-pin suspension now applies only to external suspension or an actively latched Hold target, so ordinary escaped streaming can rejoin near the bottom while active Hold still remains latched until true bottom. Added regression coverage; local validation passed 37/37 virtual-follow tests, UI typecheck, UI build, and diff checks. The medium note about measuring a dedicated inner Assistant text element remains a documented non-blocking follow-up.

pascalandr:
@CodeNomadBot
again

shantur:
## Hold-mode scroll state follow-up

I checked the current Hold + prompt-submit behavior and found a few state-machine issues that can still cause stale Hold state or unwanted bottom-follow jumps.

### 1. Issues found

- **Hold stays active after manual scroll.** Any manual scroll while a message is actively held should leave active Hold immediately and forget the held message id. The current PR keeps `holding` latched in several manual-scroll cases.
- **Stale held ids can still affect later scroll work.** Once Hold exits for message `X`, delayed alignment/restoration must not be able to scroll back to `X`.
- **Streaming completion jumps to bottom.** When the active hold target disappears at stream completion, the transition should forget the held id without emitting a bottom-scroll effect. Otherwise the viewport jumps to the bottom when streaming ends.
- **Tiny scroll while held can rejoin bottom-follow.** The first manual scroll exits Hold, but a follow-up scroll/virtualizer event can see `escaped + down-scroll + canPinToBottom` and magnet the viewport to bottom while the old streaming target is still current.
- **But true bottom should still rejoin.** Suppression should only block the near-bottom magnet path, not `atBottom` rejoin. If the viewport is genuinely at bottom, follow mode should be restored.

### 2. How we fixed it locally

- Changed `holding + user-scroll` to always become `escaped` with no scroll effect.
- Changed `hold-target-changed` for an active hold to become `escaped` with no scroll effect, so streaming completion preserves the viewport.
- Added a defensive guard in `alignHoldTarget(id)` so stale/delayed align work no-ops unless `id` is still the active held key.
- Added `suppressedHoldTargetKey` in `VirtualFollowList` so a just-released held target cannot immediately re-enter Hold or magnet back to bottom while it remains the current streaming target.
- Added `suppressRejoin` to the user-scroll transition path. This blocks the near-bottom bottom-follow magnet for a released hold target, but still allows true `atBottom` rejoin.
- Added regression coverage for manual hold exit, streaming completion without bottom jump, stale target suppression, and true-bottom rejoin.

Validation run locally:

```text
node --test packages/ui/src/components/virtual-follow-behavior.test.ts  # 41 passed
npm run typecheck --workspace @codenomad/ui                            # passed
git diff --check                                                       # passed
```

### 3. Actual diff from local changes

```diff
diff --git a/packages/ui/src/components/virtual-follow-behavior.test.ts b/packages/ui/src/components/virtual-follow-behavior.test.ts
index daaa0d7a..c144e4ca 100644
--- a/packages/ui/src/components/virtual-follow-behavior.test.ts
+++ b/packages/ui/src/components/virtual-follow-behavior.test.ts
@@ -12,8 +12,8 @@ import {
   type ScrollControllerMetrics,
 } from "./virtual-follow-behavior.ts"
 
-const userScroll = (direction: "up" | "down" | null, atBottom: boolean, canPinToBottom = false) =>
-  ({ type: "user-scroll", direction, atBottom, canPinToBottom }) as const
+const userScroll = (direction: "up" | "down" | null, atBottom: boolean, canPinToBottom = false, suppressRejoin = false) =>
+  ({ type: "user-scroll", direction, atBottom, canPinToBottom, suppressRejoin }) as const
 
 function metrics(offset: number, scrollHeight = 3000, clientHeight = 600): ScrollControllerMetrics {
   return {
@@ -39,6 +39,20 @@ describe("virtual follow behavior", () => {
     assert.deepEqual(next.effect, { type: "none" })
   })
 
+  it("does not magnet to bottom after releasing hold while the stale target is current", () => {
+    const next = transitionFollowMode({ type: "escaped" }, userScroll("down", false, true, true))
+
+    assert.deepEqual(next.mode, { type: "escaped" })
+    assert.deepEqual(next.effect, { type: "none" })
+  })
+
+  it("rejoins follow at true bottom after releasing hold", () => {
+    const next = transitionFollowMode({ type: "escaped" }, userScroll("down", true, true, true))
+
+    assert.deepEqual(next.mode, { type: "following" })
+    assert.deepEqual(next.effect, { type: "none" })
+  })
+
   it("rejoins follow and pins bottom when escaped user scrolls down with pin permission", () => {
     const next = transitionFollowMode({ type: "escaped" }, userScroll("down", false, true))
 
@@ -53,10 +67,17 @@ describe("virtual follow behavior", () => {
     assert.deepEqual(next.effect, { type: "none" })
   })
 
-  it("keeps hold latched when the user scrolls down above bottom", () => {
+  it("exits hold when the user scrolls down above bottom", () => {
     const next = transitionFollowMode({ type: "holding", key: "message-1" }, userScroll("down", false, true))
 
-    assert.deepEqual(next.mode, { type: "holding", key: "message-1" })
+    assert.deepEqual(next.mode, { type: "escaped" })
+    assert.deepEqual(next.effect, { type: "none" })
+  })
+
+  it("does not rejoin bottom follow from a held manual down-scroll reported at bottom", () => {
+    const next = transitionFollowMode({ type: "holding", key: "message-1" }, userScroll("down", true, true))
+
+    assert.deepEqual(next.mode, { type: "escaped" })
     assert.deepEqual(next.effect, { type: "none" })
   })
 
@@ -74,10 +95,10 @@ describe("virtual follow behavior", () => {
     assert.deepEqual(next.effect, { type: "none" })
   })
 
-  it("keeps hold latched for directionless user scroll away from bottom", () => {
+  it("exits hold for directionless user scroll away from bottom", () => {
     const next = transitionFollowMode({ type: "holding", key: "message-1" }, userScroll(null, false, true))
 
-    assert.deepEqual(next.mode, { type: "holding", key: "message-1" })
+    assert.deepEqual(next.mode, { type: "escaped" })
     assert.deepEqual(next.effect, { type: "none" })
   })
 
@@ -103,17 +124,17 @@ describe("virtual follow behavior", () => {
     assert.deepEqual(next.effect, { type: "align-hold", key: "message-1" })
   })
 
-  it("keeps hold latched when the hold target disappears", () => {
+  it("exits hold without jumping when the hold target disappears", () => {
     const next = transitionFollowMode({ type: "holding", key: "message-1" }, { type: "hold-target-changed", key: null, canPinToBottom: true })
 
-    assert.deepEqual(next.mode, { type: "holding", key: "message-1" })
+    assert.deepEqual(next.mode, { type: "escaped" })
     assert.deepEqual(next.effect, { type: "none" })
   })
 
-  it("keeps hold latched when a later hold target is reported", () => {
+  it("exits hold without jumping when a later hold target is reported", () => {
     const next = transitionFollowMode({ type: "holding", key: "message-1" }, { type: "hold-target-changed", key: "message-2", canPinToBottom: true })
 
-    assert.deepEqual(next.mode, { type: "holding", key: "message-1" })
+    assert.deepEqual(next.mode, { type: "escaped" })
     assert.deepEqual(next.effect, { type: "none" })
   })
 
@@ -124,6 +145,17 @@ describe("virtual follow behavior", () => {
     assert.deepEqual(next.effect, { type: "scroll-bottom", immediate: true, suppressHold: true })
   })
 
+  it("explicit top jumps leave hold and forget the held id", () => {
+    const controller = new VirtualScrollController(true)
+    controller.holdCandidate("message-1", true)
+
+    const result = controller.jumpTop(true)
+
+    assert.deepEqual(result.state.mode, { type: "escaped" })
+    assert.deepEqual(result.effect, { type: "scroll-top", immediate: true })
+    assert.equal(controller.heldKey(), null)
+  })
+
   it("prompt submission overrides a stale hold latch and returns to bottom follow", () => {
     const controller = new VirtualScrollController(true)
     controller.holdCandidate("old-assistant-answer", true)
@@ -184,7 +216,7 @@ describe("virtual follow behavior", () => {
     assert.deepEqual(next.effect, { type: "scroll-bottom", immediate: true, suppressHold: false })
   })
 
-  it("keeps auto-pin suspended while a hold target is actively latched", () => {
+  it("suspends auto-pin while active but exits hold on manual scroll", () => {
     const suspend = shouldSuspendAutoPinToBottomForHold({
       externalSuspend: false,
       activeHoldTargetKey: "streaming-assistant-answer",
@@ -194,7 +226,7 @@ describe("virtual follow behavior", () => {
     const next = transitionFollowMode({ type: "holding", key: "streaming-assistant-answer" }, userScroll("down", false, !suspend))
 
     assert.equal(suspend, true)
-    assert.deepEqual(next.mode, { type: "holding", key: "streaming-assistant-answer" })
+    assert.deepEqual(next.mode, { type: "escaped" })
     assert.deepEqual(next.effect, { type: "none" })
   })
 
@@ -238,14 +270,15 @@ describe("virtual follow behavior", () => {
     assert.deepEqual(result.effect, { type: "none" })
   })
 
-  it("does not resume or snap when a held target disappears", () => {
+  it("forgets the held id without scrolling when a held target disappears", () => {
     const controller = new VirtualScrollController(true)
     controller.holdCandidate("message-1", true)
 
     const result = controller.holdTargetChanged(null, true)
 
-    assert.deepEqual(result.state.mode, { type: "holding", key: "message-1" })
+    assert.deepEqual(result.state.mode, { type: "escaped" })
     assert.deepEqual(result.effect, { type: "none" })
+    assert.equal(controller.heldKey(), null)
   })
 
   it("lets fresh user upward movement escape even during a programmatic window", () => {
@@ -300,7 +333,7 @@ describe("virtual follow behavior", () => {
     assert.deepEqual(result.effect, { type: "none" })
   })
 
-  it("keeps hold latched until downward movement reaches actual bottom", () => {
+  it("exits hold immediately on downward movement before actual bottom", () => {
     const controller = new VirtualScrollController(true)
     controller.holdCandidate("message-1", true)
     controller.recordProgrammaticOffset(2100, false)
@@ -308,14 +341,9 @@ describe("virtual follow behavior", () => {
 
     const nearBottom = controller.observeViewport(metrics(2220), 100, false, true)
 
-    assert.deepEqual(nearBottom.state.mode, { type: "holding", key: "message-1" })
+    assert.deepEqual(nearBottom.state.mode, { type: "escaped" })
     assert.deepEqual(nearBottom.effect, { type: "none" })
-
-    controller.setUserIntent("down", 800)
-    const atBottom = controller.observeViewport(metrics(2400), 200, false, true)
-
-    assert.deepEqual(atBottom.state.mode, { type: "following" })
-    assert.deepEqual(atBottom.effect, { type: "none" })
+    assert.equal(controller.heldKey(), null)
   })
 
   it("still escapes follow on upward movement at bottom", () => {
diff --git a/packages/ui/src/components/virtual-follow-behavior.ts b/packages/ui/src/components/virtual-follow-behavior.ts
index 84e111e3..fd5cd57d 100644
--- a/packages/ui/src/components/virtual-follow-behavior.ts
+++ b/packages/ui/src/components/virtual-follow-behavior.ts
@@ -11,7 +11,7 @@ export type FollowEffect =
   | { type: "align-hold"; key: string }
 
 export type FollowEvent =
-  | { type: "user-scroll"; direction: "up" | "down" | null; atBottom: boolean; canPinToBottom: boolean }
+  | { type: "user-scroll"; direction: "up" | "down" | null; atBottom: boolean; canPinToBottom: boolean; suppressRejoin?: boolean }
   | { type: "jump-top"; immediate: boolean }
   | { type: "jump-bottom"; immediate: boolean; explicit: boolean }
   | { type: "jump-key"; key: string; block: ScrollLogicalPosition; smooth: boolean; followAfter: boolean }
@@ -95,15 +95,12 @@ export function transitionFollowMode(mode: FollowMode, event: FollowEvent): Foll
   switch (event.type) {
     case "user-scroll": {
       if (mode.type === "holding") {
-        if (event.atBottom && event.direction !== "up") {
-          return { mode: { type: "following" }, effect: noFollowEffect }
-        }
-        return { mode, effect: noFollowEffect }
+        return { mode: { type: "escaped" }, effect: noFollowEffect }
       }
       if (event.direction === "up") {
         return { mode: { type: "escaped" }, effect: noFollowEffect }
       }
-      if (mode.type === "escaped" && event.direction === "down" && event.canPinToBottom) {
+      if (!event.suppressRejoin && mode.type === "escaped" && event.direction === "down" && event.canPinToBottom) {
         return {
           mode: { type: "following" },
           effect: { type: "scroll-bottom", immediate: true, suppressHold: false },
@@ -143,7 +140,13 @@ export function transitionFollowMode(mode: FollowMode, event: FollowEvent): Foll
       return { mode, effect: noFollowEffect }
 
     case "hold-target-changed":
-      return { mode, effect: noFollowEffect }
+      if (mode.type !== "holding" || event.key === mode.key) {
+        return { mode, effect: noFollowEffect }
+      }
+      return {
+        mode: { type: "escaped" },
+        effect: noFollowEffect,
+      }
 
     case "clear-hold":
       if (mode.type !== "holding") {
@@ -267,7 +270,7 @@ export class VirtualScrollController {
     return this.result(next.effect)
   }
 
-  observeViewport(metrics: ScrollControllerMetrics, now: number, programmatic: boolean, canPinToBottom = false): ScrollControllerResult {
+  observeViewport(metrics: ScrollControllerMetrics, now: number, programmatic: boolean, canPinToBottom = false, suppressRejoin = false): ScrollControllerResult {
     const previousOffset = this.state.lastObservedOffset
     const offset = metrics.offset
     const scrolledUp = offset < previousOffset - 1
@@ -295,6 +298,7 @@ export class VirtualScrollController {
       direction,
       atBottom,
       canPinToBottom: canMagnetToBottom,
+      suppressRejoin,
     })
     this.state.mode = next.mode
     this.state.lastObservedAtBottom = this.isAutoFollowing() && atBottom
diff --git a/packages/ui/src/components/virtual-follow-list.tsx b/packages/ui/src/components/virtual-follow-list.tsx
index f1a0a257..33ee36b0 100644
--- a/packages/ui/src/components/virtual-follow-list.tsx
+++ b/packages/ui/src/components/virtual-follow-list.tsx
@@ -210,6 +210,7 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
   let lastResetKey: string | number | undefined
   let suppressAutoScrollOnce = false
   let suppressHoldUntilTargetChanges = false
+  let suppressedHoldTargetKey: string | null = null
   let lastItemsReference = props.items()
   let restoreToken = 0
 
@@ -265,6 +266,7 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
   }
 
   function resetStaleHoldForBottomIntent() {
+    suppressedHoldTargetKey = activeHoldTargetKey() ?? holdTargetKey() ?? suppressedHoldTargetKey
     restoreToken += 1
     scrollController.setRestoring(false)
     setDidTriggerHoldForCurrentTarget(false)
@@ -332,8 +334,12 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
   }
 
   function syncControllerResult(result: ScrollControllerResult) {
+    const previousHeldKey = activeHoldTargetKey()
     setFollowMode(result.state.mode)
     if (result.state.mode.type !== "holding") {
+      if (previousHeldKey !== null) {
+        suppressedHoldTargetKey = previousHeldKey
+      }
       clearHeldAnchor()
     } else if (heldAnchorKey !== null && heldAnchorKey !== result.state.mode.key) {
       clearHeldAnchor()
@@ -363,6 +369,7 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
           performance.now(),
           hasProgrammaticScrollIntent(),
           canRejoinFollowFromDownScroll(metrics),
+          shouldSuppressBottomRejoinForReleasedHold(),
         )
         break
       }
@@ -517,6 +524,7 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
     if (!streamingActive()) return false
     if (effectiveSuspendAutoPinToBottom()) return false
     if (activeHoldTargetKey() !== null) return false
+    if (shouldSuppressBottomRejoinForReleasedHold()) return false
     const items = props.items()
     if (items.length === 0) return false
     if (isAtBottom(metrics)) return true
@@ -527,6 +535,10 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
     return viewportEndIndex >= Math.max(items.length - DEFAULT_REJOIN_LAST_ITEM_COUNT, 0)
   }
 
+  function shouldSuppressBottomRejoinForReleasedHold() {
+    return suppressedHoldTargetKey !== null && holdTargetKey() === suppressedHoldTargetKey
+  }
+
   function getSnapshotMetrics(element: HTMLDivElement, handle?: VirtualizerHandle) {
     const scrollTop = handle?.scrollOffset ?? element.scrollTop
     const scrollHeight = handle ? getCurrentScrollSize(element, handle) : element.scrollHeight
@@ -728,7 +740,13 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
     setShowScrollBottomButton(hasItems && !atBottom)
     setShowScrollTopButton(hasItems && !atTop)
 
-    const result = scrollController.observeViewport(metrics, now, programmatic, canRejoinFollowFromDownScroll(metrics))
+    const result = scrollController.observeViewport(
+      metrics,
+      now,
+      programmatic,
+      canRejoinFollowFromDownScroll(metrics),
+      shouldSuppressBottomRejoinForReleasedHold(),
+    )
     if (result.state.mode.type !== followMode().type || result.effect.type !== "none") {
       suppressHoldUntilTargetChanges = false
     }
@@ -881,6 +899,7 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
   }
 
   function alignHoldTarget(key: string) {
+    if (activeHoldTargetKey() !== key) return
     const element = scrollElement()
     if (!element) return
     const target = resolveHoldTargetElement(key)
@@ -920,6 +939,7 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
     if (!autoScroll()) return
     if (externalSuspendAutoPinToBottom()) return
     if (!targetKey) return
+    if (targetKey === suppressedHoldTargetKey) return
     if (didTriggerHoldForCurrentTarget()) return
     if (suppressHoldUntilTargetChanges) return
 
@@ -1003,6 +1023,7 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
     itemElements.clear()
     setDidTriggerHoldForCurrentTarget(false)
     suppressHoldUntilTargetChanges = false
+    suppressedHoldTargetKey = null
     pendingBottomRepinAfterHold = false
     clearHeldAnchor()
   }))
@@ -1014,6 +1035,9 @@ export default function VirtualFollowList<T>(props: VirtualFollowListProps<T>) {
     if (nextTargetKey !== prevTargetKey && !hasActiveBottomFollowIntent()) {
       suppressHoldUntilTargetChanges = false
     }
+    if (nextTargetKey !== suppressedHoldTargetKey && nextTargetKey !== null) {
+      suppressedHoldTargetKey = null
+    }
     if (activeHoldTargetKey() === null) return
     if (nextTargetKey === activeHoldTargetKey()) return
     dispatchFollowEvent({ type: "hold-target-changed", key: nextTargetKey, canPinToBottom: !externalSuspendAutoPinToBottom() })
```

shantur:
@pascalandr  - Please apply the changes suggested and test. Thanks

pascalandr:
Thanks @Shantu for digging into this. but I think your patch is solving a somewhat different problem than the one this PR is trying to preserve.
The intended semantics here are:
once Hold has actually latched onto an Assistant answer, it stays latched until the user reaches the true bottom
manual scrolling above bottom should not immediately exit active Hold
normal escaped-mode near-bottom rejoin should still work when Hold is not active
Because of that, the two main transitions in your patch look wrong to me as defaults:

holding + user-scroll -> escaped
holding + hold-target-changed -> escaped
Those change the meaning of Hold itself rather than fixing a narrow bug. After that change, the patch has to reintroduce suppressedHoldTargetKey / suppressRejoin to prevent the stale target from immediately reclaiming behavior, which makes me think we are replacing one explicit state (holding until true bottom) with several secondary suppression rules.

I also do not yet see the “stream completion jumps to bottom” path on the current PR head. In the current code:
- hold-target-changed does not emit a bottom-scroll effect
- held content-grew does not emit a bottom-scroll effect
- the delayed repin path should not arm while we are in holding

So before taking that change, I want to be sure to understand exactly what you try to solve here, I mean the  alignHoldTarget(key) guard seems reasonable on its own, but I would not take the full patch as-is because it changes the core Hold contract.

shantur:
@pascalandr - Below are the following issues I saw related to auto scrolling + hold mode in the PR before the patch

Scale the window to have smaller height for easy testing

Prompt : "Return with a markdown table for 300 rows, each row contains the row number only, nothing else"

Issues
1. While the table is streaming and is longer than view port, any attempt to manually scroll will take it down straight to the bottom. I want to scroll and see the next hidden rows.
2. When the table is finished streaming, when adding any new prompt, the scroll is taken back to the beginning of the table.

The expectation is correct

1. When a message is streaming and user tries to scroll, we leave the hold on that message.
2. When message is streaming, user tries to scroll and only scrolls a bit, bottom follow isn't enabled.
3. When message is streaming, user scrolls to the bottom of message that is still streaming, bottom follow is enabled.

shantur:
I haven't taken the PR through the bot review, so it might need some cleaning up but the after the patch functionally it works.

pascalandr:
@CodeNomadBot
again

pascalandr:
@shantur
I couldn't reproduce issue 1 exactly in this PR (but was happening before this PR).
To trigger the same problem I had to scroll down to bottomn then scroll up , then next every scroll down was trigering the magnetism which is supposed to be completely disabled with Hold mode on.

Can you confirm this works now for you please ?

shantur:
@pascalandr 
Thanks for the update.

Check the video from 30 seconds 

https://www.filemail.com/d/hjdqdluzkwprcyp

shantur:
I can't reproduce the issue I posted in the video.
Seems like very edge case, we can get this in and keep an eye on it.

PR #536, opened 2026-06-08
fix(ui): throttle SSE part deltas and prevent stale delta text duplication

JDis03:
## Problem

Streaming responses send many `message.part.delta` events per second. Each event triggers a Solid store mutation + re-render, causing:

1. **Mobile freezes** — KaTeX/highlight.js re-renders on every delta block the main thread
2. **Text duplication** — When `message.part.updated` replaces a part with the server's complete state, previously enqueued deltas would flush AFTER the replacement, concatenating stale text onto the fresh part

### 1. Delta throttling (50ms buffer)
Accumulate deltas in a buffer and flush every 50ms instead of updating on every event. This batches Solid mutations and cuts main-thread blocking during streaming.

### 2. Stale delta cleanup
When a full `message.part.updated` arrives, clear any pending deltas for that part before applying the update. The update already contains the complete server state, so accumulated deltas are stale.

## Changes

- `packages/ui/src/stores/session-events.ts`: Added `enqueueDelta`, `flushDeltas`, `clearPendingDeltasForPart`; updated `handleMessagePartDelta` and `handleMessageUpdate`

## Verification

- Build passes: `npm run build:ui`
- Tested on mobile — eliminates visible freezes during long streaming responses
- No duplicate text observed at end of assistant messages after part updates

### DIFF
--- a/packages/ui/src/stores/delta-buffer.test.ts
+++ b/packages/ui/src/stores/delta-buffer.test.ts
@@ -1,0 +1,102 @@
+import assert from "node:assert/strict"
+import { afterEach, beforeEach, describe, it } from "node:test"
+import { setTimeout as delay } from "node:timers/promises"
+
+import {
+  clearPendingDeltasForPart,
+  enqueueDelta,
+  flushPendingDeltasForMessage,
+  resetDeltaBufferForTests,
+  setFlushCallback,
+} from "./delta-buffer.ts"
+
+type DeltaBatch = Array<{ instanceId: string; messageId: string; partId: string; field: string; delta: string }>
+
+describe("delta buffer", () => {
+  beforeEach(() => {
+    resetDeltaBufferForTests()
+  })
+
+  afterEach(() => {
+    resetDeltaBufferForTests()
+  })
+
+  it("concatenates matching deltas and flushes them once", async () => {
+    const flushed: DeltaBatch[] = []
+    setFlushCallback((batch) => flushed.push(batch))
+
+    enqueueDelta("instance-1", "message-1", "part-1", "text", "hello")
+    enqueueDelta("instance-1", "message-1", "part-1", "text", " world")
+
+    await delay(75)
+
+    assert.equal(flushed.length, 1)
+    assert.deepEqual(flushed[0], [
+      { instanceId: "instance-1", messageId: "message-1", partId: "part-1", field: "text", delta: "hello world" },
+    ])
+  })
+
+  it("clears pending deltas for a full part update before a stale timer flush", async () => {
+    const flushed: DeltaBatch[] = []
+    setFlushCallback((batch) => flushed.push(batch))
+
+    enqueueDelta("instance-1", "message-1", "part-1", "text", "stale")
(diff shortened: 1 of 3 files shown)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
> ### 2. Stale delta cleanup
> When a full `message.part.updated` arrives, clear any pending deltas for that part before applying the update. The update already contains the complete server state, so accumulated deltas are stale.

I am not sure if this is valid. Can you first confirm in opencode source code if this is expected and matches with how TUI behaves?

JDis03:
@shantur Regarding `clearPendingDeltasForPart`:

**Yes, this matches how the TUI behaves.** Here is the reasoning:

In the TUI (terminal), each `message.part.delta` is applied immediately — there is no batching or buffering. When `message.part.updated` arrives, it replaces the part content atomically. Any previously rendered deltas are naturally overwritten by the replacement.

In our web UI, we buffer deltas for 50ms to batch Solid store mutations and avoid main-thread blocking during streaming (KaTeX/highlight.js re-renders are expensive on mobile). This buffering introduces a race condition that does NOT exist in the TUI:

1. Deltas for part X accumulate in the buffer
2. `message.part.updated` for part X arrives with the complete server state
3. We apply the complete part (replaces content)
4. The buffer timer fires and flushes the old deltas → **stale text appended to fresh content**

`clearPendingDeltasForPart` prevents this race. The server sent the authoritative complete part — any deltas still in the buffer for that part are now stale and must be discarded. This is equivalent to what the TUI does naturally (no buffer = no stale deltas).

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

shantur:
@pascalandr  - Please hold onto this one, I would like to play with it too.
I would want to improve the markdown / syntax highlighting pipeline too so might be linked

pascalandr:
@JDis03 
Greptile review here https://github.com/Pagecran/CodeNomad/pull/17#issuecomment-4664140378

pascalandr:
> @pascalandr - Please hold onto this one, I would like to play with it too. I would want to improve the markdown / syntax highlighting pipeline too so might be linked

@Shantur I don’t see it as a blocker for this PR. This PR reduces delta-driven render frequency and also fixes stale buffered deltas causing text duplication, which is independent from markdown rendering. A markdown/highlighting refactor would reduce the cost per render; this PR reduces the number of renders and preserves event ordering. Unless you see a concrete conflict with the pipeline work, I’d prefer to keep this narrow and land it, then handle markdown/highlighting as a follow-up.

shantur:
@pascalandr - Go for it, once you are happy testing it

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr (approved):
Works nicely in my tests and improvement confirmed especially for Electron.

pascalandr:
@CodeNomadBot
again

pascalandr:
Thanks @JDis03 !

PR #519, opened 2026-06-02
fix(ui): add retry logic to SSE pong to improve connection resilience

JDis03:
## Problem

When the client receives a ping from the server, it responds with a pong via HTTP POST. On unstable networks (mobile, WiFi with poor signal, network switches), this POST can fail in multiple ways:

- **Hung requests**: `fetch()` never rejects, blocking retries indefinitely
- **Network disconnection**: `Failed to fetch`
- **Brief disconnections**: transient network errors

Previously, a single missed pong would cause the server to close the SSE connection after 45s, leaving message responses stuck in queue until the next message triggered a reconnection.

## Solution

Three improvements to make the pong POST resilient:

### 1. Request timeout (10s)
Each pong POST is now bounded with a 10s `AbortSignal` timeout. Hung requests fail fast instead of blocking indefinitely, allowing retries to start before the server's stale connection sweep.

### 2. Selective retry with `isRetryableError()`
Only retries transient failures where recovery is possible:
- `AbortError` / `TimeoutError` (hung or timed-out requests)
- `Failed to fetch` (network disconnected)
- `NetworkError` (browser network errors)

Non-retryable errors like `404 Client connection not found` (permanently closed connection) fail immediately instead of wasting retry attempts.

### 3. Exponential backoff (3 attempts, 100ms → 2000ms)
Handles burst failures gracefully without hammering the server.

## Changes

- **`packages/ui/src/lib/retry-utils.ts`** (new): Reusable retry utility with `timeoutMs` and `shouldRetry` predicate support
- **`packages/ui/src/lib/server-events.ts`**: Updated pong handler to use bounded timeout + selective retry

## Verification

- Build passes: `npm run build:ui` ✅
- Manually verified in production logs: retries now visible as `Pong failed after retries` instead of single immediate failure
- SSE monitor log at `~/.codenomad/logs/sse-monitor.log` shows `PONG_OK` / `PONG_FAIL` / `STALE` events for ongoing monitoring

Issue #543, opened 2026-06-11
[Bug]: Corrupted time_archived in SQLite causes HTTP 400 on session list — zero sessions shown


raz123:
## Bug

A single session with a malformed `time_archived` value in the SQLite database causes the entire session list API to fail with HTTP 400. This makes CodeNomad display zero sessions with no error indication.

## Root Cause

The OpenCode session list response is a paginated array. When serializing, the schema for `SessionInfo.time.archived` expects `DateTime.Utc` format, but the DB stores a raw millisecond timestamp:

```
1781132189927  (milliseconds since epoch)
```

When the encoder reaches the session with this bad value, serialization fails:
```
Expected DateTime.Utc, got 1781132189927
  at ["items"][37]["time"]["archived"]
```

## Impact

Any session list request with a page size larger than the index of the corrupted session fails with HTTP 400. CodeNomad requests 50-200 sessions per page — if the bad session is within that range, ZERO sessions return. The UI never gets data and shows the empty "Start a conversation" state.

## Reproduction

1. Set any session's `time_archived` to a raw millisecond number in the shared SQLite DB
2. Open a workspace with that session in CodeNomad
3. Session list returns 0 items (HTTP 400 swallowed silently)

## Fix

Two fixes needed:
1. **Data cleanup**: Set the bad `time_archived` to NULL
2. **Resilience**: The API should skip or gracefully handle individual session serialization failures rather than failing the entire response

## Environment
- CodeNomad v0.17.0 (Tauri) and v0.17.1 (web)
- OpenCode binary v1.15.13

shantur:
Hi @raz123 ,

Please use the latest opencode binary.

Thanks

shantur:
@raz123 - Is this still an issue with latest CodeNomad and latest OpenCode?

Issue #542, opened 2026-06-11
[Bug]: OpenCode returns { items } but CodeNomad expects { data } — session list empty


raz123:
## Bug

When OpenCode v1.15.13 serves the V2 session list API, it returns:
```json
{ "items": [...], "cursor": {...} }
```

But CodeNomad's session accessor function (`getV2SessionItems` / compiled name `Qc`) expects:
```js
function Qc(e) { return e.data }
```

Since the response has no `.data` key, the session list is never populated. UI shows "Start a conversation" (empty state) despite having 75+ sessions in the database.

## Root Cause

The `response.data` access in `packages/ui/src/stores/session-api.ts:145-147`:
```ts
function getV2SessionItems(response: V2SessionsResponse): SessionV2Info[] {
  return response.data
}
```

The actual server response is `{ items: [...], cursor: {...} }` — the key is `.items`, not `.data`.

## Fix

Change to check both formats:
```ts
function getV2SessionItems(response: V2SessionsResponse): SessionV2Info[] {
  return (response as any).items ?? response.data
}
```

## Impact

Session listing across ALL workspaces returns empty. No error displayed — UI just shows the blank "Start a conversation" state. User has no indication the fetch failed.

Confirmed on OpenCode binary v1.15.13. Possibly related to #525 (format change in 1.16.0+).

## Environment
- CodeNomad v0.17.0 (Tauri) and v0.17.1 (web)
- OpenCode binary v1.15.13

shantur:
Hi @raz123,

That's correct.
OpenCode is changing the SDK and we want to be compatible with the latest. 
Please upgrade to the latest opencode

shantur:
Try latest CodeNomad with opencode 1.17+
re-open if needed

PR #549, opened 2026-06-12
fix(ui): SSE connection status indicators for mobile network drops

JDis03:
## Problem

Mobile users experience SSE disconnections (wifi drops, Cloudflare idle timeouts) but get no visual feedback. The per-instance status dots remain green even when the SSE transport is down, because instance status updates travel over the broken SSE pipe.

## Solution

This PR implements **immediate connection status feedback** by propagating SSE transport state to all instance status indicators:

### Changes

1. **server-events.ts**: Add `onDisconnect()` handler
   - Mirrors existing `onOpen()` pattern
   - Fire `disconnectHandlers` in `scheduleReconnect()` when connection drops

2. **sse-manager.ts**: Register SSE lifecycle handlers
   - `onDisconnect`: set ALL instances to `connecting` (amber dot)
   - `onOpen`: clear `connecting` status (green dot restored by subsequent events)
   - Log transitions for debug visibility

3. **AGENTS.md**: Add PR Review Principles
   - Check regressions first, look for better implementations
   - Be the PR gatekeeper, ruthless code quality
   - Test before responding, UI/server version parity

## Verification

- ✅ TypeScript compilation clean (pre-existing SDK errors unrelated)
- ✅ Vite build successful
- ✅ Mobile testing: SSE disconnect → immediate amber dots
- ✅ Reconnect → green dots restored

## Edge Cases Handled

- **Transient drop**: amber → green on reconnect (no false disconnect modal)
- **Instance dies during outage**: amber → green → `disconnected` event → red
- **0 instances**: Map empty, loop is no-op
- **Rapid reconnect cycles**: idempotent (setting `connecting` on `connecting` is no-op)

## Related

- Complements PR #519 (pong retry with timeout) - merged upstream
- Addresses mobile UX gap: workspace-level SSE state now visible to all users (not just debug overlay)


pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

AGENTS.md
```diff
@@ -15,6 +15,14 @@
 - Prefer composable primitives (signals, hooks, utilities) over deep inheritance or implicit global state.
 - When adding platform integrations (SSE, IPC, SDK), isolate them in thin adapters that surface typed events/actions.
 
+## PR Review Principles
```
shantur: @JDis03 - please keep this as separate

shantur:
@CodeNomadBot

Check again

shantur:
@JDis03 can you please check the new changes

JDis03:
Reviewed `7d3da501` — LGTM 🚀

**Architecture improvement:** Separating `connectionStatus` (real instance state) from `transportStatus` (SSE transport) is strictly better than our original approach. The original mutated the status Map on transport drop, which could destroy genuine `disconnected`/`error` instance states. The new `deriveDisplayConnectionStatus()` pure function keeps both concerns independent.

**Checklist:**
- ✅ `deriveDisplayConnectionStatus` logic correct, 5 tests covering all cases
- ✅ `mapDesktopEventTransportStatus` Tauri parity correct, tested
- ✅ `onTransportStatus` mirrors `onOpen` pattern — consistent API surface
- ✅ `emitTransportStatus("disconnected")` fires in `scheduleReconnect` before retry timer — correct ordering
- ✅ `transportStatus` defaults to `"connecting"` — safe initial state
- ✅ Tests use `node:test` — zero extra dependencies

No blocking findings. Ready to merge.

shantur:
@JDis03 please do try running too and confirm

JDis03:
## Tested locally on fork (OpenCode 1.15.3)

**Working:**
- ✅ SSE transport disconnect/reconnect detected and logged
- ✅ Status dots reflect transport state (green → amber on heartbeat timeout, green on reconnect)
- ✅ Instance status preserved — real `disconnected`/`error` states are NOT overwritten by transport drops
- ✅ Heartbeat (15s timeout) detects SSE inactivity when `EventSource.onerror` doesn't fire (mobile wifi drops) and triggers `disconnected` transport state
- ✅ No log spam — `connectionStatus` only logs on actual state changes, not on every SSE event
- ✅ Auto-retry on reconnect works (existing behavior)

**One question for you:** What's the intended UX for the status dot when transport has been down for a prolonged period?
- Currently stays amber (`connecting`) indefinitely while auto-reconnect attempts run
- Would a timeout → red (e.g. after 30-60s without reconnect) be desirable?

**No regressions found.** Everything looks solid on this end.

JDis03:
**One more thing about mobile detection:**

`EventSource.onerror` doesn't fire immediately on mobile wifi drops — it can take 45-60s for the TCP keepalive to expire. On desktop and emulated mobile it's fine, but on real mobile devices (tested on Android/Chrome) the disconnect is silent until the TCP timeout.

To address this, I tested a **client-side heartbeat timer (15s)** locally. Every SSE event resets the timer; if no activity is detected for 15s, it forces `scheduleReconnect()` which triggers the disconnect transport state and updates the status dot. This is the only reliable way to detect mobile wifi drops without depending on `EventSource.onerror` or `navigator.onLine`.

Happy to open a separate PR for the heartbeat if you'd like to keep this PR focused on the infrastructure change.

shantur:
> `EventSource.onerror` doesn't fire immediately on mobile wifi drops — it can take 45-60s for the TCP keepalive to expire. On desktop and emulated mobile it's fine, but on real mobile devices (tested on Android/Chrome) the disconnect is silent until the TCP timeout.

IMO, that's acceptable. It's not critical if the status updates a bit later. What do you think?

JDis03:
@shantur — sorry to bother, quick question about something I'm stuck with on my fork.

I'm running CodeNomad server 0.16.0 with OpenCode 1.15.3. When I restart the server (PM2), sessions often don't load in the UI even though `opencode session list` shows them fine. I looked into it and it seems related to issues #542 (`response.data` vs `response.items` format change in OpenCode) and #543 (corrupted `time_archived`). 

I'd love to contribute a fix but I want to make sure I'm not duplicating work — is anyone already working on this? Or is there a known workaround I'm missing?

Thanks in advance 🙏

shantur:
@JDis03 - That's known issue. Use latest opencode version 1.17.0+
It's some incompatibility introduced and CodeNomad now only supports newer versions of OpenCode

shantur:
We would also need @pascalandr to test the Tauri native desktop-events to ensure its working there too.

JDis03:
@shantur — agree, 45-60s is reasonable. I tested with a 15s heartbeat locally and it worked, but I think it's better as a separate follow-up if someone needs faster feedback. The PR as-is already covers the important infrastructure.

Thanks for the tip on OpenCode 1.17.0+ — I'll update my fork 👍

pascalandr:
> We would also need @pascalandr to test the Tauri native desktop-events to ensure its working there too.

Validated Tauri native desktop-events on Windows.

Initial issue found during testing: after enabling SSE logs and reloading DevTools, the native stream was alive and receiving heartbeats, but the UI stayed orange because the connected transport status was not replayed after reload.

Fix applied: native batch open path now also emits connected status.

Retest passed:
- Event transport: rust-native (desktop_event_transport)
- SSE transport status changed {status: 'connected'}
- Dots stay green after reload
- Backend kill still turns dots orange/connecting

pascalandr (approved):

PR #555, opened 2026-06-15
fix(ui): handle SDK version mismatch in session list API

### LINKED ISSUE #554: [Bug]: session list fails when SDK resolves to version <1.16.0 — client.v2 not found
## Bug

When `@opencode-ai/sdk` resolves to version <1.16.0 (e.g. 1.2.6 due to monorepo dependency conflicts), the UI session list silently fails because `client.v2.session.list()` doesn't exist — `v2` is not a property on `OpencodeClient` in older SDK versions.

The actual SDK API exposes `client.session.list()` directly.

## Root cause

`packages/ui/src/stores/session-api.ts:142`:
```ts
return requestData<V2SessionsResponse>(client.v2.session.list(options), "v2.session.list")
```

This assumes `client.v2` exists, but older SDK versions expose session directly on the client.

Additionally, `getV2SessionItems` assumes `response.data` format but OpenCode returns `{ items: [...] }` starting from v1.15.13+.

## Fix

I have a working fix locally:
1. Use `client.session.list()` instead of `client.v2.session.list()`
2. Make `getV2SessionItems` handle multiple response formats: `items`, `data`, or array directly
3. Same for `getV2NextCursor` — check both `cursor.next` and `next`

Happy to open a PR if this is not already being worked on.

JDis03:
## Problem

`client.v2.session.list()` is called in `session-api.ts:142`, but when `@opencode-ai/sdk` resolves to version <1.16.0 (e.g. v1.2.6 due to monorepo dependency constraints), the `v2` property doesn't exist on `OpencodeClient`. The session list silently fails → UI shows "Start a conversation" despite having sessions in the database.

Additionally, OpenCode >=1.15.13 returns `{ items: [...], cursor: {...} }` but the code expects `response.data`.

## Changes

- `fetchV2Sessions`: Use `client.session.list()` instead of `client.v2.session.list()` with SDK version-agnostic options mapping
- `getV2SessionItems`: Handle all response formats — direct array, `{ items: [...] }`, or `{ data: [...] }`
- `getV2NextCursor`: Check both `cursor.next` and `next`
- Types: Use `SessionListResponse` (`{ 200: Session[] }`) instead of `V2SessionsResponse` for SDK compatibility

## Verification

- ✅ Tested with SDK v1.2.6 (installed) and OpenCode 1.15.3
- ✅ Tested with SDK v1.2.6 and OpenCode 1.17.7
- ✅ Sessions load correctly in both cases
- ✅ No regressions in session creation, deletion, or message loading

Fixes #554

### DIFF
--- a/packages/ui/src/stores/session-api.ts
+++ b/packages/ui/src/stores/session-api.ts
@@ -7,7 +7,7 @@ import {
   type SessionStatus,
 } from "../types/session"
 import type { Message } from "../types/message"
-import type { SessionV2Info, V2SessionsResponse } from "@opencode-ai/sdk/v2/client"
+import type { Session as SessionV2Info, SessionListResponse as V2SessionsResponse } from "@opencode-ai/sdk/v2/client"
 
 import { instances } from "./instances"
 import { preferences, setAgentModelPreference } from "./preferences"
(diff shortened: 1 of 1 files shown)


shantur:
@JDis03 - Please use latest OpenCode v1.17.0+
We don't intend to support older versions.

PR #556, opened 2026-06-16
fix(ui): keep streaming bash output pinned

shantur:
## Summary
- notify the existing virtual follow-list path after streaming bash ANSI output updates
- restore the inner bash output scroll immediately after direct pre updates
- apply the same coalesced notification pattern to final ANSI output

## Validation


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

Check again

PR #561, opened 2026-06-18
feat(yolo): move permission auto-accept (Yolo) to the server

heunghingwan:
## Summary

Yolo (permission auto-accept) currently lives entirely in the UI. Each browser keeps its own toggle in `localStorage` and auto-replies to permission requests over a 4-hop path (`OpenCode → server SSE → UI → server proxy → OpenCode`). The server — which already sits on the event stream that carries every `permission.asked` — does none of the work.

This PR makes the **server authoritative**: it owns the toggle state, resolves family-root inheritance, and performs the auto-reply in-process via loopback using the same `"once"` semantics the UI used to send. The UI becomes a pure view: it toggles via REST and mirrors state from a `yolo.stateChanged` SSE event.

## Why

- **Correctness**: the server already consumes the instance SSE stream (`InstanceEventBridge`); auto-accepting there is the natural choke point instead of bouncing to the UI and back.
- **Multi-client**: previously each browser had independent `localStorage` state and never synced. Toggles now broadcast to all connected clients in real time.
- **Headless**: Yolo keeps auto-accepting even when no UI is connected (useful for long autonomous runs).
- **Latency**: drops from 4 hops to a single in-process loopback call.

## What changed

**Server (new, authoritative)**
- `permissions/auto-accept-store.ts` — in-memory state keyed by family root. Faithful port of `resolvePermissionAutoAcceptFamilyRoot`: fork/`revert` sessions root at themselves; enabling any member enables the whole family.
- `permissions/auto-accept-manager.ts` — subscribes to `instance.event`, builds the session tree from `session.created/updated/deleted` (`properties.info`), intercepts `permission.v2.asked` / `permission.asked`, dedupes in-flight replies, emits `yolo.stateChanged` / `yolo.autoAccepted`, clears per-instance state on `workspace.stopped/error`.
- `permissions/opencode-replier.ts` — default replier calling OpenCode directly (`getInstancePort` + auth header), mirroring `background-processes/manager.ts`.
- `server/routes/yolo.ts` — `GET/POST /workspaces/:id/yolo/sessions/:sid[/toggle]`, following existing route conventions.
- `api-types.ts` / `events/bus.ts` — `YoloStateResponse` + the two new event types registered in `onEvent` so they flow over `/api/events`.

**UI (pure view)**
- `permission-auto-accept.ts` — removed `localStorage`, persistence, and drain logic; now a runtime (non-persisted) projection. `resolvePermissionAutoAcceptFamilyRoot` is **retained as a display aid** so the badge still lights up for child/sub-sessions of an enabled family (preserving the inheritance UX).
- `instances.ts` — toggle calls REST (optimistic, reconciled on success, reverted on failure); subscribes to `yolo.stateChanged`; `ensureYoloStateSynced` backfills the active session's state on first connect (deduped per session, reset on SSE reconnect so it re-syncs after a server restart).
- `session-events.ts` — removed the three `drainAutoAcceptPermissionsForInstance` hooks (the server drains now).
- `api-client.ts` — `getYoloState` / `toggleYolo`.

## Behavior parity

| Aspect | Before | After |
|---|---|---|
| Reply semantics | `"once"` | `"once"` (unchanged) |
| Inheritance | whole family (root + non-fork descendants) | **same** |
| Fork isolation | `revert` session is its own root | **same** |
| Persistence | UI `localStorage` | none (intentional, see Notes) |
| Multi-client sync | ❌ independent per browser | ✅ real-time via SSE |
| Works with UI closed | ❌ | ✅ |
| Auto-reply path | 4 hops | 1 in-process hop |

## Testing

32 unit tests added (`node:test`), all passing. Server + UI typecheck clean.

- **Store (16)**: inheritance (parent/child/sibling), fork isolation, cyclic parent chains, late parent discovery, `revert` re-rooting, per-instance independence, tree maintenance.
- **Manager (13)**: real `properties.info` event shapes, v2 vs legacy reply, in-flight dedup + retry-after-resolve, `yolo.stateChanged` emission, `session.deleted` keeps toggle, `workspace.stopped` clears state, `stop()` unsubscribes.
- **UI (3)**: retained `resolvePermissionAutoAcceptFamilyRoot` display-projection tests.

## Notes for reviewers

- **No persistence is intentional** for this milestone — server restart resets all Yolo state (matches the agreed scope). The UI mirror self-heals via SSE reconnect + `ensureYoloStateSynced`. Persistence can be layered on later (e.g. into `~/.config/codenomad/config.json`) without touching the manager.
- **Family-root resolution lives in two places on purpose**: the server resolves it to decide whether to auto-reply; the UI resolves the same pure function to render the badge instantly (synchronous memo). Both are faithful ports; no network round-trip is added for display.
- **`properties.info` nesting**: OpenCode wraps session records under `properties.info` for `session.*` events (permission events are flat). The manager handles both and the tests use the real nested shape to guard against regressions.
- `getYoloState` / `yolo.autoAccepted` are wired but `yolo.autoAccepted` is not yet consumed by the UI — it's available on the wire for future observability (e.g. an audit log / toast).

## Risk / rollback

The change is additive on the server and the UI gracefully degrades to the SSE mirror. If the server lacks the new routes (mixed-version), the UI's optimistic toggle still flips locally and `toggleYolo` failures are logged + reverted, so no hard breakage.


shantur:
Hi @heunghingwan ,

Thanks for the PR. 
This is the right approach, we will review it

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

heunghingwan:
## Review fixes pushed

All blocking findings from the gatekeeper review plus additional issues found during code review have been addressed in `f425e842`. Branch rebased onto latest `dev`.

### P0 — v2 reply used the deprecated OpenCode contract
`opencode-replier` sent v2 replies to `/session/{sessionID}/permissions/{permissionID}` with body `{ response }` — the deprecated `Permission.respond` route. Switched to the current `Permission2.reply` contract: `/api/session/{sessionID}/permission/request/{requestID}/reply` with body `{ reply }`.

### P1 — enabling Yolo no longer drains already-pending permissions (behavior regression)
The previous UI called `drainAutoAcceptPermissions` on toggle. The server-side manager only intercepted *new* events, so permissions that arrived before the toggle were silently skipped. Added a pending-permission set: ask/update events add to it, replied events remove from it, and `toggle(on)` drains all pending entries for the same family root.

### P1 — family-root toggle lost when session ancestry arrives late
Toggle state was keyed by the resolved root at toggle time. If a child was toggled before its parent was known, the toggle silently disappeared when the parent arrived. Added `migrateEnabledRoots()` on `upsertSession` to re-resolve and migrate enabled roots after any session tree change.

### P1 — stale Yolo badge after server clears state on workspace stop
Server cleared toggle state without emitting events; UI `removeInstance` did not clear the Yolo mirror or `syncedYoloSessions` cache. Added `clearPermissionAutoAcceptForInstance()` and `clearSyncedYoloSessionsForInstance()` wired into `removeInstance`.

### Additional fixes
- **`permission.updated` duplicate reply** — now only processed for permissions already in the pending set
- **Optimistic toggle revert race** — captures `wasEnabled` before toggle instead of naively flipping current state on failure

### Validation
42 permission-related tests pass (35 server + 7 UI). 206/208 full-repo tests pass (2 pre-existing failures: Chinese-locale git message, solid-toast browser dependency). Server + UI + Electron typecheck clean.

pascalandr:
@CodeNomadBot
same review

heunghingwan:
## Second round of fixes pushed

All blocking findings from the bot re-review have been addressed. Two additional commits: `f425e842` (core fixes) and `0f33ee5c` (SDK client migration + remaining items).

### SDK client replaces handwritten routes

The opencode-replier previously hand-assembled OpenCode HTTP URLs. This was wrong **twice**: first using the deprecated `Permission.respond` contract (`/session/{sid}/permissions/{pid}` + `{response}`), then a stale 1.16.0 route with an extra `/request/` segment not present in the pinned 1.17.8 contract (`/api/session/{sid}/permission/request/{rid}/reply` vs the correct `/api/session/{sid}/permission/{rid}/reply`).

The replier now uses the auto-generated SDK client (`@opencode-ai/sdk/v2/client`) via a new `createInstanceClient` factory (`workspaces/instance-client.ts`). Routes and body shapes always match the installed SDK version — no more hand-maintained URL strings. The factory also passes the workspace's root directory so OpenCode can scope permission replies correctly (matching the old UI path that included the `workspace` query param).

Other server modules that hand-build `http://127.0.0.1:{port}/...` URLs (`background-processes/manager.ts`, `workspaces/instance-events.ts`) can adopt this factory for the same version-correct routing.

### tsconfig: `moduleResolution` Node → Bundler

The SDK declares subpath exports (e.g. `@opencode-ai/sdk/v2/client`) via the standard `exports` field in `package.json`. TypeScript's classic `"Node"` resolution predates this field and cannot resolve subpath imports — it only reads `main`. `"Bundler"` supports `exports` and is a **strict superset** of `"Node"`: every existing import resolves identically.

**No runtime impact.** At runtime the server uses `tsx`, which delegates to Node's native module resolution (Node 18+ already supports `exports`). The tsconfig setting only governs type-checking.

Additional benefits:
- Server can now import any modern package with subpath `exports` (previously impossible without reaching into `/dist` by hand)
- Aligns with the UI package, which already uses bundler resolution for the same SDK subpaths

### Behavioral fixes

**Pending permission drain on enable.** The previous UI auto-accepted already-pending permissions when Yolo was toggled. The server-side manager only caught new events, so permissions that arrived before the toggle were silently skipped. Added a pending-permission set: `permission.asked`/`v2.asked` events add to it, `permission.replied`/`v2.replied` events remove from it, and `toggle(on)` drains all pending entries for the same family root.

**Family-root migration on late ancestry.** Toggle state was keyed by the resolved root at toggle time. If a child was toggled before its parent was known, the toggle silently disappeared when the parent arrived. Added `migrateEnabledRoots()` on `upsertSession` to re-resolve and migrate enabled roots after any session tree change. The test that codified the old (buggy) behavior has been updated.

**`permission.updated` duplicate reply.** After a successful auto-reply the permission is removed from the pending set; `permission.updated` for already-replied permissions is now skipped instead of triggering a second (failing) reply attempt.

**State cleanup on workspace stop AND error.** The server clears Yolo state on both `workspace.stopped` and `workspace.error`. The UI now clears its Yolo mirror + `syncedYoloSessions` cache in both cases (previously only `workspace.stopped` triggered cleanup via `removeInstance`).

**Optimistic toggle revert race.** The UI's `togglePermissionAutoAcceptForSession` now captures `wasEnabled` before the optimistic toggle and reverts to that exact value on REST failure, instead of naively flipping the current signal (which could be wrong if a `yolo.stateChanged` SSE event arrived in between).

**yoloManager lifecycle.** `yoloManager.stop()` moved from the `sseClients` set (semantically per-SSE-connection) into the http-server `stop()` method alongside `closeSseClients()` and `app.close()`.

### Validation

38 permission-related tests pass. Server + UI + Electron typecheck clean. Branch rebased onto latest `dev`.

pascalandr:
@CodeNomadBot
same review

heunghingwan:
Root `package-lock.json` updated — `@opencode-ai/sdk` is already hoisted in the root `node_modules` (UI dependency), so the lockfile change is a single line adding it to the server workspace entry. No new packages installed.

Regarding the typecheck failures in the bot worktree: those are environment issues (`Cannot find type definition file for node`, missing `solid-js`/`@opencode-ai/sdk` resolution) — not PR-specific. All three typechecks pass cleanly in a properly installed workspace:

```
npm run typecheck --workspace @neuralnomads/codenomad   ✅
npm run typecheck --workspace @codenomad/ui             ✅
npm run typecheck --workspace @neuralnomads/codenomad-electron-app  ✅
```

pascalandr:
Re-review after `3af691c1`, with extra focus on the historical YOLO optimization from #424: auto-accept should not depend on showing/waiting for the permission UI.

**Blocking finding**

- **P1: pending permissions are not drained when late session ancestry makes them part of an enabled YOLO family.** In `packages/server/src/permissions/auto-accept-manager.ts:126-159`, `handlePermissionRequest()` adds the permission to `pending` and checks `this.store.isEnabled(instanceId, sessionId)` immediately. If the permission event for a child/subagent arrives before the server has ingested that child’s `session.created` / `session.updated` ancestry, `isEnabled()` resolves the child as its own root and returns false. Later, `ingestSession()` updates the tree via `this.store.upsertSession(...)`, but it never calls `drainPending()`, so the already-pending permission is not retried even though it now resolves into the enabled family root. This regresses the #497 family-root behavior and can also undo the #424 improvement by leaving a YOLO-covered permission visible/pending until some unrelated event or toggle happens.

  Minimal fix: after `upsertSession()` in `ingestSession`, drain pending permissions for the ingested session, e.g. `this.drainPending(instanceId, session.id)`, with a test for “root enabled, child permission arrives before child ancestry, child session update arrives, permission is auto-accepted.”

**Important follow-up**

- `yolo.autoAccepted` is emitted by the server but is not consumed by the UI. The only UI listener is `yolo.stateChanged` in `packages/ui/src/stores/instances.ts:71-77`, while queue cleanup still depends on the later OpenCode `permission.replied` path in `packages/ui/src/stores/session-events.ts:740-748`. If that replied event is delayed or missed, the permission can remain in the UI queue even though the server already auto-accepted it. Given #424 specifically optimized away waiting on permission UI display/render timing, it would be safer to handle `yolo.autoAccepted` directly by marking the permission replied and removing it from the queue/message store.

**Previously blocking items now look resolved**

- The v2 reply path now uses the generated SDK client instead of handwritten routes.
- `workspace.error` now clears the UI YOLO mirror/cache, matching server-side cleanup.
- The root `package-lock.json` now reflects the new server `@opencode-ai/sdk` dependency.
- Existing targeted permission/Yolo tests pass locally.

**Validation**

- `npm exec tsx -- --test packages/ui/src/stores/permission-auto-accept.test.ts packages/ui/src/stores/permission-replies.test.ts packages/server/src/permissions/auto-accept-manager.test.ts packages/server/src/permissions/auto-accept-store.test.ts`: passed, 41/41.
- `npm run typecheck --workspace @neuralnomads/codenomad`: failed in this local worktree on existing dependency/type-resolution issues (`fuzzysort`, `commander`), not clearly PR-specific.

**Verdict**

Do not merge yet. The direction is right, but the server must re-drain pending permissions when late session ancestry changes their effective YOLO family, otherwise the server-side move can still leave YOLO-covered permissions stuck in the UI.

heunghingwan:
## Third round of fixes — regressions from #424/#497 + hardening

Two regressions from the original YOLO design were identified and fixed, plus all non-blocking concerns from the deep review have been addressed.

### Regressions fixed

**Re-drain pending on session ancestry arrival (#497 regression)**

The original UI re-drained the permission queue whenever `session.updated` revealed a parentId — ensuring that a child permission arriving before its ancestry was known would be auto-accepted once the parent showed up. The server-side manager called `upsertSession` → `migrateEnabledRoots` but never `drainPending`. Now `ingestSession` calls `drainPending(instanceId, session.id)` after every session tree mutation, closing the race.

Test: "re-drains pending when late session ancestry joins an enabled family".

**UI consumes `yolo.autoAccepted` (#424 regression)**

The #424 optimization ensured auto-accept cleanup happens at the queue level, not tied to UI render or external event timing. With the server doing the auto-reply, the UI depended solely on the OpenCode `permission.replied` SSE event arriving through the server proxy — a fragile multi-hop path. If that event was delayed or missed, the permission lingered in the UI queue. Now the UI listens for `yolo.autoAccepted` and immediately marks the permission replied + removes it from the queue and v2 message store.

### Hardening from deep review

| Fix | Detail |
|-----|--------|
| `permission.updated` source inference | Source is now inferred from the pending entry rather than defaulting to `legacy`, so a v2 permission updated via `permission.updated` preserves its original `v2` source for the correct SDK reply path |
| Replier failure cap | After `MAX_REPLY_ATTEMPTS` (3) failures, the permission is removed from `pending` + `replyAttempts`, freeing memory. The permission isn't permanently blocked — a new `permission.asked` re-adds it with a fresh counter |
| `session.deleted` clears pending | Pending permissions for a deleted session are now removed, preventing memory retention for abandoned sessions |
| `ensureYoloStateSynced` skips `"info"` | The info pseudo-session no longer triggers a pointless REST call |
| `clearInstance` cleans `inFlight` | Prefix-scans and removes entries for the stopped instance |
| Empty `pending` Maps pruned | `removePending` / `removePendingForSession` delete the instance-level Map when it becomes empty |
| HTTP timeout on loopback calls | `createInstanceClient` applies a 10s `AbortSignal.timeout` to all SDK calls, preventing hung connections from leaving `inFlight` entries indefinitely |

### Stale documentation updated

`.opencode/skills/codenomad-architecture-guide/references/sdk-integration-patterns.md` and `feature-traces.md` updated to reflect the server-side YOLO architecture (previously still described the removed `drainAutoAcceptPermissions()` mechanism).

### Validation

45 permission-related tests pass (11 new tests this round: ancestry re-drain, permission.updated source inference × 2, replier failure × 2, workspace.error cleanup, session.deleted clears pending, plus the existing drain/replied/clearInstance suite). Server + UI + Electron typecheck clean.

pascalandr:
Re-reviewed latest head `1853bfec`.

No blocking findings from this pass.

The two regressions called out in the previous review look addressed:

- `packages/server/src/permissions/auto-accept-manager.ts:139-155` now calls `drainPending(instanceId, session.id)` after session tree updates, so a permission that arrived before child ancestry was known is re-drained once that child joins an enabled family.
- `packages/ui/src/stores/instances.ts:79-91` now handles `yolo.autoAccepted` directly by marking the permission replied and removing it from both the queue and v2 message store, preserving the #424 queue-level cleanup behavior instead of relying only on the later OpenCode `permission.replied` event.

Additional hardening also looks reasonable from static review: source inference for `permission.updated`, pending cleanup on `session.deleted`, per-instance cleanup for `inFlight`/attempts, and loopback timeout.

Validation run locally:

- `npm exec tsx -- --test packages/ui/src/stores/permission-auto-accept.test.ts packages/ui/src/stores/permission-replies.test.ts packages/server/src/permissions/auto-accept-manager.test.ts packages/server/src/permissions/auto-accept-store.test.ts`: passed, 48/48.
- `npm run typecheck --workspace @neuralnomads/codenomad`: still fails in this worktree on dependency/type-resolution issues around `fuzzysort` and `commander`; I am not treating that as PR-specific because it matches the earlier local environment issue.

I do not see a remaining YOLO functional blocker in the latest code,

@shantur this is a huge quantity of code just to get server-authoritative YOLO +1583/-146  !

@heunghingwan I will merge the PR if testing goes fine.

pascalandr (approved):
thanks @heunghingwan !

PR #563, opened 2026-06-19
feat(ui): Add tool level expansion settings.

shantur:
## Summary
- add a central tool-call registry for renderer lookup and configurable expansion defaults
- add Appearance presets for transcript detail: Minimal, Balanced, Detailed, and Everything
- allow per-item customization for thinking, known tool calls, and Other tools with responsive settings layout
- persist toolCallExpansionDefaults while keeping legacy expansion preferences synchronized for command compatibility

## Validation


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur:
Addressed the tool-title registry duplication review in follow-up commit f78307d5.

- Removed the separate tool-title.ts / TITLE_RENDERERS map entirely.
- Normal tool headers now use the already-resolved renderer from the central tool registry.
- Kept the task renderer's legacy metadata.summary fallback, but it now uses state titles and generic input details instead of a second renderer map.
- Removed the remaining todoread branches from tool-call utilities.

Clarification: todoread is not a deprecated tool here; it is not an available tool in this registry/UI, so the stale references were removed rather than kept as a deprecated alias.

Validation: npm run typecheck --workspace @codenomad/ui

shantur:
@CodeNomadBot

Now?

PR #565, opened 2026-06-19
fix(ui): scope project session list requests

shantur:
## Summary
- update @opencode-ai/sdk to 1.17.8
- load session lists through client.session.list with scope=project
- use a single high-limit project request because this endpoint does not support cursor pagination

## Validation
- npm run typecheck --workspace packages/ui

## Notes
- Existing untracked .opencode/package-lock.json was left untouched.


shantur:
@CodeNomadBot

check for regressions
Be the PR gatekeeper

shantur:
@CodeNomadBot

check again

shantur:
Thanks for the detailed review. I fixed the API cleanup issue in 8ff9ed1a by removing the ignored `limit` option from `fetchSessions`, deleting the stale `getSessionFetchLimit` call path, and updating the close-session refresh caller to pass only `reset`.

On the remaining pagination concern: I do not think we should implement offset pagination with `start` for this endpoint. The SDK surface exposes `start?: number`, but for `client.session.list` that parameter is a timestamp/start-time filter, not a page offset. Using it as `0, 200, 400...` would produce incorrect query semantics rather than reliable incremental loading.

The reason this PR uses `client.session.list({ scope: "project", limit: 1000 })` is that it is the SDK endpoint that serializes `scope=project` and includes worktree sessions. The cursor-paginated `client.v2.session.list` path serializes `/api/session`, but the generated SDK does not support `scope` there, which is why those requests did not include project-scoped worktree sessions.

So the tradeoff in this PR is intentional: restore project/worktree session visibility via the endpoint that supports `scope=project`, while avoiding fake pagination with a timestamp filter. If we need lossless loading for projects above 1,000 sessions, I think the correct follow-up is an OpenCode API/SDK change that supports both project scope and real pagination (cursor or documented offset) on the same endpoint, rather than using `start` as pagination in the UI.

shantur:
@CodeNomadBot

Again

omercnet:
This scope change is amazing! Thank you! Saves a lot of troubling finding missing sessions

pascalandr:
@Shantur
This is not a real fix; it merely introduces another bug. Replacing pagination with a single request capped at 1,000 sessions causes sessions to disappear from the UI when subagents push a project beyond that limit, even though the sessions remain intact in OpenCode’s database.

PR #572, opened 2026-06-30
fix(ui): remove scope=project from session list requests

JDis03:
## Problem

PR #565 (commit e29c3a08) added `scope=project` to session list requests to avoid pagination limits. With OpenCode 1.17.x, the `/session` endpoint ignores the `directory` filter when `scope=project` is present and returns **all** sessions from the global SQLite database instead.

This makes the same session appear in every project tab. Deleting a session from one tab deletes it from the shared DB, affecting all of them.

## Reproduction

```
# Pick any opencode workspace port
PORT=4096
curl -s -u "codenomad:$PASSWORD" \
  "http://127.0.0.1:$PORT/session?directory=/home/user/project-a&scope=project" \
  | jq 'length'
# Returns sessions from project-b too — should only return project-a's

curl -s -u "codenomad:$PASSWORD" \
  "http://127.0.0.1:$PORT/session?directory=/home/user/project-a" \
  | jq 'length'
# Correct: only project-a's sessions
```

Related: anomalyco/opencode#33113

## Fix

Drop `scope: "project"` from `buildProjectSessionListOptions()`. The `directory` parameter alone correctly filters sessions to the requested folder on OpenCode 1.17.x. `scope: "project"` can be reintroduced once the backend respects `directory` alongside it.

## Validation

- `directory=/test` → 1 session (correct)
- `directory=/idm` → 1 session (correct)
- Same session no longer appears in multiple project tabs
- Deleting a session in one project no longer affects others

### DIFF
--- a/packages/ui/src/stores/session-api.ts
+++ b/packages/ui/src/stores/session-api.ts
@@ -53,10 +53,17 @@ import { clearCacheForSession } from "../lib/global-cache"
 import { getLogger } from "../lib/logger"
 import { requestData } from "../lib/opencode-api"
 import { getRootClient } from "./opencode-client"
-import { getWorktreeSlugForSession, migrateLegacyWorktreeMapToSessionMetadata, pruneStaleLegacyWorktreeMapEntries, removeLegacyParentSessionMapping, setWorktreeSlugForParentSession } from "./worktrees"
+import {
+  getWorktreeSlugForSession,
+  getWorktrees,
+  migrateLegacyWorktreeMapToSessionMetadata,
+  pruneStaleLegacyWorktreeMapEntries,
+  removeLegacyParentSessionMapping,
+  setWorktreeSlugForParentSession,
+} from "./worktrees"
 import { getOpenCodeWorkspaceIdForSession } from "./opencode-workspaces"
 import { hydrateSessionMetadataWithClient } from "./session-metadata"
-import { PROJECT_SESSION_LIST_LIMIT, buildProjectSessionListOptions } from "./session-list-options"
+import { PROJECT_SESSION_LIST_LIMIT, buildProjectSessionListOptions, filterProjectScopedSessions } from "./session-list-options"
 
 const log = getLogger("api")
 
@@ -142,7 +149,11 @@ async function fetchV2Sessions(instanceId: string, options: V2SessionListOptions
   const client = getRootClient(instanceId)
   const listOptions = buildProjectSessionListOptions(options)
   const data = await requestData<SessionListResponse>(client.session.list(listOptions), "session.list")
-  return { data }
+  const allowedDirectories = [options.directory, ...getWorktrees(instanceId).map((worktree) => worktree.directory)]
+
+  return {
+    data: filterProjectScopedSessions(data, allowedDirectories),
+  }
 }
 
 function getV2SessionItems(response: ProjectSessionListResponse): SDKSession[] {
@@ -208,7 +219,7 @@ async function fetchSessions(instanceId: string, options?: { reset?: boolean }):
   try {
     const sessionListOptions = instance.folder ? { directory: instance.folder } : {}
 
-    log.info("session.list", { instanceId, limit: PROJECT_SESSION_LIST_LIMIT, directory: sessionListOptions.directory })
+    log.info("session.list", { instanceId, limit: PROJECT_SESSION_LIST_LIMIT, directory: sessionListOptions.directory, scope: "project" })
     const response = await fetchV2Sessions(instanceId, sessionListOptions)
 
     let statusById: Record<string, any> = {}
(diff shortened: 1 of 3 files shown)


pascalandr:
@CodeNomadBot
check for regressions
Be the PR gatekeeper

pascalandr:
@CodeNomadBot
same review

shantur:
@pascalandr 
I think this needs to be analysed. I had a requirement to add scope=project to make the sessions available from worktrees.

JDis03:
## Analysis of the `scope=project` removal

@shantur — thanks for raising the worktree concern. I dug into our actual session data before pushing back and want to share what I found so we can decide this on evidence rather than speculation. I am happy to revise the approach if it conflicts with the worktree requirement.

### What `scope=project` actually does in OpenCode 1.17.x

Looking at `~/.local/share/opencode/opencode.db`, OpenCode groups sessions by `project_id`, and `project_id` is derived from `project.worktree`. The interesting mappings I see today:

| Directory | Project ID | Notes |
|---|---|---|
| `/home/dark/Project/idm` | `f6d9b8cf…` | Single folder in this project |
| `/home/dark/Project/codenomad` | `e3625924…` | Single folder |
| `/home/dark/Project/test` | `global` | Non-git folder |
| `/home/dark/Project/tiktok` | `global` | Non-git folder, same as `/test` |

Sessions in non-git folders are merged into a `"global"` project, so `/test` and `/tiktok` share the same `project_id = global`.

### What the user-visible bug was

In a fresh CodeNomad tab for `/test`, the sidebar showed `Hola greeting`, `Find Jellyfin IMDB lookup endpoint`, **and** `Automatizar publicaciones TikTok` — the last one is `/tiktok`, not `/test`. Both folders are `"global"` in OpenCode, so `scope=project` returned both.

When I added `directory=/home/dark/Project/test` as well, OpenCode 1.17.9 on this build **did not narrow the response**. The user behavior I confirmed earlier (sidebar mixing sessions across folder tabs) lines up with that.

### Worktree concern

I cannot reproduce a session-in-worktree scenario on this machine today, so I do not want to claim that `scope=project` is "safe to remove globally". The two paths I can think of:

1. **Sibling directories under the same git repo (the `bc5afa1…` project):** SQLite shows `memory-system` and the now-removed `hackDark` both had `project_id = bc5afa1…`. This is the only place `scope=project` would surface additional sessions in the same project. They are not literal git worktrees though — `hackDark` was just a normal sibling that OpenCode matched onto the same project row. If the desired behavior is "show all sessions in the same git repo across sibling folders", we have lost that with this PR.
2. **Actual git worktree sessions** (`.codenomad/worktrees/foo`) — I do not have a reproducible case here. The behavior the previous PR #565 added would have supported them, but I do not know if they ever appeared correctly under `scope=project` either, given the broader backend bug.

### Concrete questions before merge

@shantur, can you confirm one of:

1. Which exact `directory` value does the missing worktree session have in OpenCode DB? (`x-opencode-directory` forwarded by CodeNomad, or the main repo path?)
2. When the worktree session is missing, does adding `scope=project` to the current request actually return it, or does it still ignore `directory`?
3. If the answer to (2) is "yes it does return it today on a fresh OpenCode 1.17.x", then a follow-up OpenCode bug to support both `directory` filtering AND worktree-session inclusion is the right path — and this PR is the right temporary fix until that lands.

Without (2) being true, keeping `scope=project` does not actually solve the original issue #518 either; it only papers over one form of cross-folder leakage with another.

### My read

- This PR fixes the reproducer I have (the one @pascalandr and I saw with `Hola greeting`/`TikTok` showing in `/test`).
- The cost is potentially losing cross-sibling-folder visibility for `project_id`-grouped sessions, which may or may not include real worktree sessions depending on OpenCode version.
- I think the right next step is OpenCode upstreaming support for `{directory, scope=project}` together, then re-adding `scope=project` here once that lands. Until then, removing it is strictly less buggy than keeping it on 1.17.x.

Happy to revise wording, add a test for the sibling-folder scenario, or split the change into two PRs if that helps. Just want to make sure the decision is based on the actual session data and not the original PR #565 description.

pascalandr:
@JDis03 To answer your analysis more directly: I agree with the distinction you made.

I pushed a comit to  try to preserve both requirements:
- keep `scope=project` so OpenCode still returns project/worktree sessions
- then filter the returned sessions client-side to the current root directory plus known CodeNomad worktree directories

So this avoids the blanket removal of `scope=project`, while still preventing unrelated `/test` / `/tiktok`-style leakage when the backend ignores `directory`.

This should be a safer temporary client-side workaround until OpenCode supports `{ directory, scope=project }` correctly upstream.

Validation passed: focused session pagination tests, UI typecheck, UI build, and Tauri release build/manual exe launch.

pascalandr:
@CodeNomadBot


check for regressions
Be the PR gatekeeper

pascalandr (approved):

PR #577, opened 2026-07-09
feat(ui): focus messages after prompt submission

pascalandr:
## Summary

Move keyboard focus from the prompt to the active message stream immediately after a desktop prompt submission.

This makes conversation navigation keys such as Page Up and Page Down available without requiring a click. Starting to type printable text returns focus to the prompt for a follow-up message.

## Changes

- Focus the current session's message stream before waiting for the message, command, or shell request.
- Carry an explicit focus intent through first-prompt session creation so the newly mounted session focuses its message stream instead of the detached draft view.
- Do not overwrite focus if the user changes controls or opens a modal during first-session activation.
- Restore the active prompt when a submission fails.
- Make the message stream programmatically focusable without adding it to the normal Tab order.
- Keep type-to-focus active on hybrid touch/mouse devices and preserve `!` shell-mode activation from the message stream.

## Unchanged behavior

- Touch-only submissions keep their existing focus behavior.
- Existing global Escape handling is unchanged.

## Validation

- Focused prompt and scrolling tests pass: `24/24`.
- UI typecheck passes.
- UI production build passes.
- `git diff --check` passes.
- Independent gatekeeper re-review found no remaining blockers.


pascalandr:
Gatekeeper review of head `0f824a6c` against current `dev`.

The focused stream itself works once focus is retained: `tabIndex={-1}` makes it programmatically focusable without adding a Tab stop, and native Page Up, Page Down, arrows, Home/End, Space, and Shift+Space integrate correctly with the existing `VirtualFollowList` intent handling. Picker and shell-mode Escape priority is also preserved.

There are three blocking lifecycle issues in the actual focus transitions.

## Blocking findings

1. **P1: Escape focus is immediately undone by the global Escape shortcut in idle sessions.**

`handlePromptKeyDown()` focuses `.message-stream` at `packages/ui/src/components/prompt-input.tsx:761-765`, but it does not consume the event. The same keydown then reaches the global shortcut listener at `packages/ui/src/lib/hooks/use-app-lifecycle.ts:131-135`.

When the active session is idle, the Escape shortcut calls `blurInput()` (`packages/ui/src/lib/shortcuts/escape.ts:61-63`), which blurs `document.activeElement`. At that point the active element is the stream this PR just focused, so focus falls back to the document body.

Reproduction:

1. Open an idle session and focus the prompt.
2. Press Escape with no picker open and outside shell mode.
3. The stream receives focus in the textarea handler.
4. The global handler immediately blurs it.

The advertised Escape behavior therefore works while the session is busy (first Escape enters abort debounce) but not in the normal idle case. The prompt handler needs to coordinate with the global shortcut instead of focusing before a later global blur.

2. **P1: the first prompt focuses the detached draft stream instead of the newly created session.**

The first-message handlers create and activate a session before awaiting the send (`packages/ui/src/components/instance/instance-shell2.tsx:955-982`). Activation replaces the no-session draft subtree at `instance-shell2.tsx:1222-1283` with a new `SessionView`.

When the old draft `PromptInput.handleSend()` resumes, `focusMessageStream()` resolves through its old `wrapperRef` (`packages/ui/src/components/prompt-input.tsx:456-468,561-563`). That root is detached or belongs to the obsolete draft view, so it cannot focus the new active session's stream. The newly mounted `SessionView` separately schedules focus to its prompt at `packages/ui/src/components/session/session-view.tsx:238-282`.

Reproduction:

1. Open an instance with no active session.
2. Submit the first prompt, slash command, or shell command.
3. After the new session is activated, press Page Up/Page Down.

Focus remains in the new prompt instead of the new conversation. The accepted-submission focus must be owned by the parent that knows which session became active, or explicitly locate the active pane after the transition.

3. **P1: delayed submission completion can steal focus from a follow-up prompt or another control.**

`handleSend()` awaits the message/command/shell request at `prompt-input.tsx:529-547` and then unconditionally focuses the stream at `:561-563`. It does not check whether the user has already started another prompt or moved focus elsewhere while the request was pending.

Reproduction:

1. Submit over a delayed connection.
2. Start typing the next prompt before `promptAsync` resolves, or click another same-session control.
3. When the first request completes, focus jumps to the stream.

This directly conflicts with the follow-up workflow motivating this change. Transfer focus at a deterministic submission/acceptance point, and do not override focus if the user has moved on since the submission began.

## Important non-blocking findings

- Hybrid touch/mouse devices can receive post-submit stream focus but have no type-to-focus recovery. Submission uses `isTouchOnlyPointer()` (`prompt-input.tsx:333-336,561`), while the global typing listener is disabled by the broader `isCoarsePointer()` check (`:328-342`). On a device with a coarse primary pointer and an attached fine pointer, typing after submission will not return to the prompt. The two capability gates should use consistent semantics.
- The programmatically focused stream is an unnamed generic container (`packages/ui/src/components/virtual-follow-list.tsx:693-703`). A region role and localized accessible name would make the focus transition understandable to assistive technology.

## Validation

- Focused prompt, virtual-follow, and submit-pin tests: passed, `21/21`.
- UI typecheck: passed.
- UI production build: passed; only existing chunk-size warnings.
- `git diff --check upstream-review/dev...HEAD`: passed.
- GitHub Linux and Windows builds passed; remaining matrix jobs were still pending at review time.
- There is currently no DOM-level regression test covering the Escape/global-handler interaction or draft-to-session replacement.

## Verdict

**Request changes. Do not merge yet.** The existing-session submit path can focus a real stream, but idle Escape immediately loses that focus, first-session submission targets a detached view, and async completion can overwrite newer user focus.

pascalandr:
Gatekeeper re-review of head `d590c6f3` after the fixes for https://github.com/NeuralNomadsAI/CodeNomad/pull/577#issuecomment-4948539317.

## Resolved findings

- The conflicting prompt-level Escape behavior was removed. The existing global Escape lifecycle remains unchanged.
- Existing-session focus now moves to the stream before the asynchronous request begins, so request completion cannot steal focus from a follow-up prompt or another control.
- First-prompt submission now carries an explicit focus intent into the newly mounted `SessionView`; it no longer targets the detached draft stream.
- First-session activation rechecks active-session state, modal state, and whether another element claimed focus before moving focus.
- Failed first submissions restore the active newly created session's prompt.
- Touch-only submissions keep their previous behavior, while hybrid touch/mouse devices retain type-to-focus recovery.
- Typing `!` from the focused stream still enters shell mode.

The message stream remains programmatically focusable without entering the normal Tab order, and the existing virtual-follow keyboard intent path handles Page Up, Page Down, arrows, Home/End, Space, and Shift+Space correctly.

## Validation

- Focused prompt/focus/scrolling tests: passed, `24/24`.
- UI typecheck: passed.
- UI production build: passed; only existing chunk-size warnings.
- `git diff --check`: passed.
- Independent final re-review: no remaining concrete blockers.
- The first GitHub matrix run for this head was cancelled by the PR-description update; its replacement run is queued.

## Verdict

**No blocking findings on `d590c6f3`. Technically approved, pending the replacement GitHub matrix.**

pascalandr:
@shantur LGTM

shantur (approved):

PR #492, opened 2026-05-21
build: publish deb and portable tar.gz Linux artifacts

shantur:
## Summary
- Publish one supported installer on Linux: a Tauri .deb package for x64.
- Publish one portable fallback: an Electron .tar.gz archive for x64.
- Remove the temporary Flatpak path and stop publishing Linux AppImage, RPM, and zip variants.
- Remove obsolete Linux assets from reused releases before uploading replacements.
- Closes #487, Closes #488

## Artifact validation
- Extract the Electron archive and verify its executable, bundled server resources, app.asar renderer entry, referenced assets, and shared-library closure.
- Verify Tauri Debian metadata, bundled resources, desktop entry, shared libraries, and package installation in Ubuntu 24.04.
- Synchronize Tauri package versions before every platform build so generated metadata follows release inputs.

## Validation
- Parsed the workflow YAML and Electron package JSON.
- Verified the local @electron/asar APIs used by CI.
- Ran the Tauri version synchronization script.
- Ran git diff --check.
- Reviewed the final diff against current dev; full Linux packaging and Docker installation are delegated to PR CI.

## Compatibility
The Tauri deb is built and installation-tested on Ubuntu 24.04. Older Debian-based distributions are not yet guaranteed.

PR #579, opened 2026-07-11
feat(speech): separate STT/TTS provider configuration

heunghingwan:
## Summary

Voice input (speech-to-text / STT) and voice output (text-to-speech / TTS) currently share a single API key, base URL, and provider configuration. Users cannot use different providers for each direction — for example, Groq for transcription (STT) and OpenAI for synthesis (TTS).

This PR adds a `separateProviders` toggle to the speech settings. When **off** (default), behavior is unchanged — existing configs work as-is. When **on**, STT and TTS each get their own `apiKey`, `baseUrl`, and `model` fields, allowing independent OpenAI-compatible endpoints per direction.

### Server
- **`api-types.ts`**: Added `separateProviders`, `sttConfigured`, `ttsConfigured`, `sttBaseUrl`, `ttsBaseUrl` to `SpeechCapabilitiesResponse`
- **`speech/service.ts`**: Extended Zod schema with `separateProviders` + nested `stt`/`tts` sub-objects. Split `createProvider()` into direction-aware `createSttProvider()`/`createTtsProvider()` with `resolveSttSettings()`/`resolveTtsSettings()` resolvers that fall back to shared values when per-direction fields are absent
- **`settings/public-config.ts`**: Extended `sanitizeServerOwner` to strip per-direction `stt.apiKey`/`tts.apiKey` and set `stt.hasApiKey`/`tts.hasApiKey` booleans (same pattern as the shared `apiKey`)
- **`OpenAICompatibleSpeechProvider`**: Completely unchanged — it receives the same flat `NormalizedSpeechSettings` it always has

### UI
- **`stores/preferences.tsx`**: Extended `SpeechSettings` type with `separateProviders`, `stt`, `tts` sub-objects. Updated `normalizeSpeechSettings()` and `updateSpeechSettings()` to handle per-direction patches
- **`components/settings/speech-settings-card.tsx`**: Added toggle at the top of the card. When enabled, shows two sections (Input/STT and Output/TTS) with their own API Key, Base URL, and Model fields. Playback Mode and TTS Format remain shared. Also includes "Test input" button alongside existing "Test playback"
- **`lib/audio-utils.ts`**: Extracted shared `blobToBase64`, `createMediaRecorder`, `stopTracks` utilities (used by both `usePromptVoiceInput` and `useTranscriptionTest`)
- **`lib/hooks/use-transcription-test.ts`**: New hook for testing STT transcription from the settings card
- **`components/prompt-input/usePromptVoiceInput.ts`**: Uses `sttConfigured` instead of combined `configured`
- **`stores/conversation-speech.ts`**, **`lib/hooks/use-speech.ts`**, **`speech-settings-card.tsx`**: TTS consumers use `ttsConfigured` instead of combined `configured`
- **`styles/components/settings-screen.css`**: Added `settings-card-section-header`/`settings-card-section-title` styles
- **i18n**: Added new keys per locale across all 9 locales (en, es, ja, zh-Hans, fr, de, he, ne, ru)

## Backward Compatibility

- `separateProviders` defaults to `false` — existing configs work unchanged
- The `configured` field remains in capabilities response as `(sttConfigured || ttsConfigured)` so any consumer not yet updated still works
- `NormalizedSpeechSettings` (the provider-facing flat interface) is unchanged

## Testing

- 13 unit tests covering direction-aware resolution (shared mode, separate mode, partial config, fallback chains, model overrides) and config sanitization
- All tests pass: `npx tsx --test packages/server/src/speech/service.test.ts packages/server/src/settings/public-config.test.ts`
- Server typecheck: clean
- UI typecheck: clean
- UI build: succeeds (no duplicate i18n keys)


pascalandr:
Gatekeeper review of latest head `9ea41836`. I reviewed this as a security- and regression-sensitive configuration change, including the server resolution path, secret sanitization, settings persistence, and the new microphone test lifecycle.

## Blocking findings

1. **P0: a shared API key can be sent to a different directional provider.**

`packages/server/src/speech/service.ts:168-197` resolves `apiKey` and `baseUrl` independently. In separate mode, configuring a directional URL without a directional key combines that URL with the shared or environment key. The inverse also happens: a directional key without a directional URL is sent to the shared/environment/default endpoint.

I reproduced the TTS case locally with a shared OpenAI key and a third-party directional URL. The actual request was:

```json
{
  "url": "https://third-party.example/v1/audio/speech",
  "authorization": "Bearer sk-shared-openai"
}
```

The credential is attached in `packages/server/src/speech/providers/openai-compatible.ts:143-155`. This is a credential-disclosure risk, not only a configuration error.

Treat endpoint and credential as one atomic connection tuple:

- if neither directional value is overridden, inherit both shared values;
- if only one is overridden, reject the configuration;
- if both are overridden, use only the directional pair.

The current test at `packages/server/src/speech/service.test.ts:90-104` codifies the unsafe shared-key fallback without verifying the destination receiving that key.

2. **P1: clearing a directional Base URL does not clear the stored override.**

`packages/ui/src/components/settings/speech-settings-card.tsx:229-247` submits an empty string. `packages/ui/src/stores/preferences.tsx:328-341` normalizes that to `undefined`, and `JSON.stringify` omits it in `packages/ui/src/lib/api-client.ts:410-414`.

The server merge patch deletes a field only when it receives explicit `null` (`packages/server/src/settings/merge-patch.ts:23-26`). The old URL therefore remains stored and reappears after save. Empty directional model/Base URL values have the same inability to delete an override cleanly; in fallback cases the UI can instead persist the current shared value as a permanent directional override.

3. **P1: microphone recording can start after the Settings component has been closed.**

In `packages/ui/src/lib/hooks/use-transcription-test.ts:42-80`, cleanup can run while `getUserMedia()` is pending and `mediaStream` is still null. When the promise resolves, the continuation creates a recorder and starts the microphone without checking whether the component was disposed or whether the request is stale. The `shouldFinalize` flag only guards finalization; it does not guard asynchronous acquisition.

4. **P1: repeated clicks during the permission prompt can create multiple live streams.**

The hook remains in `idle` until `getUserMedia()` resolves (`use-transcription-test.ts:46-80`), so the button remains enabled at `speech-settings-card.tsx:301-305`. Multiple clicks can start multiple acquisitions. Each completion overwrites the shared recorder/stream references, leaving earlier streams no longer controllable by the UI.

Add a `requesting` state or request-generation token, and immediately stop streams returned for stale requests.

5. **P1: microphone tracks stay live during the whole transcription request.**

`use-transcription-test.ts:98-130` moves the stream into a local variable and stops its tracks only after the network request finishes. Once detached, component cleanup cannot reach it. A hung transcription request can therefore leave microphone capture active indefinitely. Stop the tracks immediately when recording stops, before encoding and uploading.

## Important findings

6. **`configured = sttConfigured || ttsConfigured` is not actually backward-compatible.**

At `packages/server/src/speech/service.ts:93-112`, a legacy client can see `configured: true` and `supportsTts: true` when only STT has credentials, then enable an operation that fails at runtime. Previously, `configured: true` meant the shared credential backed both advertised operations. Preserving that old invariant likely requires `configured = sttConfigured && ttsConfigured`; new clients can use the directional flags for partial availability.

7. **Switching modes can silently discard hidden unsaved edits.**

The toggle only changes `separateProviders` (`speech-settings-card.tsx:380-386`), while dirty detection and save serialize only the currently selected mode (`:178-260`). For example, editing the shared key/URL, switching to separate mode, and saving discards the hidden shared edits when drafts are reset at line 262. The inverse drops unsaved directional edits.

8. **Directional secret sanitization should fail closed.**

`packages/server/src/settings/public-config.ts:23-34` deletes `stt.apiKey`/`tts.apiKey` only when the value is a non-empty primitive string. Malformed values can pass through a sanitizer whose security purpose is to prevent secret exposure. Delete every `apiKey` property unconditionally, then derive `hasApiKey` separately; ideally validate settings on write.

9. **“Test input” tests persisted settings, not the visible drafts.**

`use-transcription-test.ts:113-118` sends only audio and uses server-persisted configuration. It cannot test a newly entered endpoint/key/model before saving, despite appearing beside that form. It also introduces a substantial microphone lifecycle surface unrelated to the core provider-separation change. I recommend splitting this feature into a follow-up PR unless it is made draft-aware and its state machine is hardened.

## Scope and maintainability

- `packages/ui/src/components/settings/speech-settings-card.tsx` is now approximately 621 lines.
- `packages/ui/src/stores/preferences.tsx` is approximately 1,162 lines, above the repository's target limit.
- The core separate-provider behavior and the new STT recording/test feature should be considered separately; splitting the latter would reduce review and regression surface.

## Validation

- `npm exec tsx -- --test packages/server/src/speech/service.test.ts packages/server/src/settings/public-config.test.ts`: passed, 13/13.
- Current GitHub build checks are green.
- Local server typecheck remains blocked by the existing workspace resolution errors around `fuzzysort` and `commander`; I did not classify those as PR-specific.
- The cross-provider credential disclosure described above was reproduced locally against the actual request path.

## Verdict

Do not merge as-is. The feature direction is useful, but endpoint/key fallback currently creates a concrete credential disclosure path, directional overrides cannot be reliably cleared, and the new microphone test has multiple lifecycle leaks. These need to be resolved before this configuration can be considered safe.

pascalandr:
Gatekeeper re-review of latest head `8344ed0c` after the review-fix commit.

Several previous findings are improved: shared credentials are no longer sent to a directional URL, directional fields now use explicit `null` deletion, the legacy `configured` invariant is restored in separate mode, stale microphone acquisitions are generation-guarded, and tracks are released before transcription.

However, the latest fix still has blocking issues, including regressions introduced by the fix itself.

## Blocking findings

1. **P0: credential/endpoint pairing is still unsafe in the inverse direction.**

`packages/server/src/speech/service.ts:173-179` and `:192-198` prevent a shared key from being sent to a directional URL, but a directional key without a directional URL is still combined with the shared/environment/default endpoint.

I reproduced the TTS case locally with a directional third-party key and no directional URL. The actual request was:

```json
{
  "url": "https://api.openai.com/v1/audio/speech",
  "authorization": "Bearer sk-third-party"
}
```

This is the inverse of the original credential leak. Endpoint and credential still are not treated as an atomic tuple.

The test at `packages/server/src/speech/service.test.ts:106-120` does not cover this case: it supplies directional keys but only checks configured flags. It never verifies which URL receives the key.

Required behavior:

- neither directional key nor URL: inherit the complete shared tuple;
- both directional key and URL: use the complete directional tuple;
- exactly one directional value: reject/report the incomplete configuration instead of combining providers.

2. **P1: Test Input cannot stop an active recording.**

`packages/ui/src/components/settings/speech-settings-card.tsx:301-305` disables the button whenever `testTranscription.state() !== "idle"`. That includes the `recording` state.

The hook requires a second `toggle()` while recording to call `stopRecording()` (`packages/ui/src/lib/hooks/use-transcription-test.ts:153-156`), but the only UI control capable of sending that toggle is disabled. The button changes its label/icon to Stop but cannot be clicked, so recording continues until the component is destroyed.

The control should remain enabled while `recording` and be disabled only while `requesting` or `transcribing`.

3. **P1: one permission or recorder-start failure permanently breaks future Test Input attempts.**

The error path at `packages/ui/src/lib/hooks/use-transcription-test.ts:95-98` calls `cleanup()`. `cleanup()` sets `disposed = true` at lines `32-34`, but `disposed` is never reset for a normal retry.

After one denied permission or recorder startup failure:

- the UI returns to idle;
- a later attempt can successfully obtain a stream;
- line `71` sees `disposed === true`, stops the stream, and returns;
- state remains `requesting`, leaving the button disabled indefinitely.

Separate normal attempt cleanup from component disposal. Only the `onCleanup` lifecycle path should permanently set `disposed = true`.

4. **P1: switching provider mode still silently discards unsaved edits.**

The toggle handler at `packages/ui/src/components/settings/speech-settings-card.tsx:379-391` resets the entire draft from persisted `serverSettings()` and clears every key-edit/clear flag.

This does not preserve edits or warn the user; it merely moves the loss earlier. Editing a shared key, URL, model, voice, playback mode, or format and then enabling separate mode immediately destroys those unsaved values. The inverse loses directional edits.

Preserve independent draft sets across the toggle, or confirm before discarding dirty values. Rebuilding from persisted settings without confirmation is still silent data loss.

## Previous findings that look resolved

- Shared API keys are no longer sent to explicitly directional URLs.
- `configured` now remains false for partial separate-mode configuration, preserving legacy-client behavior.
- Empty directional Base URL/model values now become `null` merge-patch deletions.
- Pending `getUserMedia()` results are generation-checked and stale streams are stopped.
- Microphone tracks are stopped before encoding/uploading.

## Remaining non-blocking concerns

- `packages/server/src/settings/public-config.ts:23-34` still removes directional `apiKey` only when it is a non-empty primitive string. A security sanitizer should delete the property unconditionally and derive `hasApiKey` separately.
- Test Input still exercises persisted server settings rather than the visible unsaved drafts, so it cannot validate a new provider configuration before saving.
- The speech settings component remains approximately 627 lines, and `preferences.tsx` remains above 1,100 lines.

## Validation

- `npm exec tsx -- --test packages/server/src/speech/service.test.ts packages/server/src/settings/public-config.test.ts`: passed, 14/14.
- The latest GitHub matrix is still partially pending at the time of review.
- The inverse cross-provider credential disclosure was reproduced locally against the actual request path.

## Verdict

Do not merge yet. The original credential leak was fixed only in one direction, while the fix commit introduced a recording control that cannot stop, a permanent failure state after one permission error, and immediate silent draft loss on mode switch.

pascalandr:
Gatekeeper re-review of latest head `a70cf473`.

The four blockers from the previous pass are addressed in isolation: the server no longer combines a truly incomplete directional pair, recording can be stopped, permission failures are retryable, and toggling modes no longer immediately resets local drafts.

The end-to-end settings path still has blocking regressions, however. The new server tests exercise manually constructed config objects, but not what the UI and `updateSpeechSettings()` actually persist.

## Blocking findings

1. **P0: the UI reconstructs a “complete” pair and can still send a directional key to the wrong provider.**

`createDirectionalDraft()` pre-populates each directional URL from the shared URL (`packages/ui/src/components/settings/speech-settings-card.tsx:32-38`). Once a directional key is entered, `handleSave()` persists that displayed URL as an explicit directional URL (`:235-249`).

Reproduction:

1. Shared URL is `https://api.openai.com/v1`.
2. Enable separate providers.
3. Enter a Groq key in the STT key field, but do not touch the pre-populated STT URL.
4. Save and transcribe.

The persisted pair is the Groq key plus the OpenAI URL. The resolver at `packages/server/src/speech/service.ts:173-190` now sees two directional values and correctly, but unsafely, treats them as an intentional complete pair. The key is sent to the shared provider endpoint.

This bypasses the atomic-pair protection added in the latest commits. Inherited display values must not silently become directional overrides. Keep explicit/inherited URL state separate, or require the user to explicitly provide/confirm the directional endpoint when setting a directional key.

2. **P1: every shared-mode save permanently deletes all directional credentials and configuration.**

The shared branch unconditionally sends:

```ts
stt: { apiKey: null, baseUrl: null, model: null },
tts: { apiKey: null, baseUrl: null, model: null },
```

at `packages/ui/src/components/settings/speech-settings-card.tsx:254-267`.

This is not limited to a separate→shared transition. If dormant directional config exists while shared mode is active, changing only voice, playback format, a shared model, or the shared URL deletes both directional keys, URLs, and models. Switching back to separate mode cannot recover the redacted secrets.

It also contradicts the original mode-toggle behavior documented in this PR: disabling separate mode was supposed to preserve directional config in storage and merely ignore it at runtime. Do not attach destructive credential deletion to unrelated shared-mode saves. If deletion is desired, expose it as an explicit, confirmed action.

3. **P1: clearing one directional key leaves its foreign endpoint behind, disables the direction, and then claims it uses the shared key.**

At `speech-settings-card.tsx:238-249`, `clearSttApiKey()` / `clearTtsApiKey()` count as `has*Key`, so clearing a key sends `apiKey: null` while preserving the directional URL/model.

For an STT config using Groq:

```yaml
stt:
  apiKey: sk-groq
  baseUrl: https://api.groq.com/openai/v1
```

clicking Clear produces:

```yaml
stt:
  baseUrl: https://api.groq.com/openai/v1
```

The service correctly refuses to send the shared key to that foreign URL (`service.ts:173-190`), so STT becomes unconfigured. After save, the UI sees no directional key but does see a shared key and renders “Using shared API key” (`speech-settings-card.tsx:457-475`), even though runtime explicitly refuses that fallback.

Clearing a directional credential must either clear the complete directional tuple or leave an explicit incomplete state with accurate UI and a repair path.

4. **P1: keyless directional URL/model edits can be silently ignored because the store rebuilds a full normalized patch.**

The card attempts not to persist directional URL/model fields without a directional key (`speech-settings-card.tsx:238-249`). But `updateSpeechSettings()` starts from all current normalized settings and spreads that complete object into the PATCH (`packages/ui/src/stores/preferences.tsx:679-705`). Normalization always materializes fallback directional models (`:326-343`) and retains existing directional URLs.

Consequences:

- switching to separate mode with no directional keys still persists directional models;
- an existing keyless foreign URL is re-emitted;
- changing or blanking that stale URL is omitted by the card and then restored by the store;
- the stale URL continues to make that direction unconfigured.

The update function should construct a minimal merge patch from explicit caller intent, not normalize the full redacted object and write it back. Add tests for the actual card→store→merge-patch path; resolver-only tests cannot catch these regressions.

## Important remaining findings

- Inputs and the mode toggle remain enabled while saving. `handleSave()` snapshots drafts at `speech-settings-card.tsx:229-233`, then unconditionally replaces them after the async save at `:269-277`; edits made during that window are silently lost while the UI reports Saved.
- Speech dirty state is still local to the Save button and is not registered with the settings discard guard, so closing/navigating away can silently lose unsaved credentials and endpoints.
- `packages/server/src/settings/public-config.ts:23-34` still deletes directional `apiKey` only when it is a non-empty primitive string rather than sanitizing the property fail-closed.
- Test Input still tests persisted server settings, not the visible drafts.

## Validation

- `npm exec tsx -- --test packages/server/src/speech/service.test.ts packages/server/src/settings/public-config.test.ts`: passed, 17/17.
- GitHub build matrix was still partially pending at review time.

The recording lifecycle corrections look sound in this pass. The remaining blockers are now primarily in configuration semantics and persistence, where the current unit tests do not cover the emitted merge patches.

## Verdict

Do not merge yet. The backend resolver is safer, but the UI can synthesize the unsafe key/endpoint pair it was meant to reject, shared-mode saves destroy unrelated credentials, and directional clear/edit operations persist misleading or unrecoverable states.

pascalandr:
## Corrected gatekeeper re-review of `0b7f00de`

I re-audited the complete 29-commit PR history, all prior gatekeeper comments, the original design document, and the corresponding behavior on `dev`. This replaces the earlier version of this comment, which overclassified several inherited or design-ambiguous behaviors as PR-specific blockers.

The latest commits materially improve the branch: endpoint/key resolution now rejects incomplete directional tuples, the UI no longer pre-populates a shared URL beside a new directional key, directional clearing removes the complete tuple, settings updates are minimal merge patches, and the microphone lifecycle regressions are fixed. Focused tests, UI typecheck/build, `diff --check`, and the GitHub build matrix pass.

## Blocking finding

1. **P1: switching from separate providers back to shared mode destroys both directional configurations, contrary to the approved design.**

The original design is explicit:

> **ON → OFF:** Save `separateProviders: false`. Per-direction values stay in config but are ignored.

The current shared-mode save instead deletes both complete tuples when `wasSeparate` is true (`packages/ui/src/components/settings/speech-settings-card.tsx:247-259`):

```ts
...(wasSeparate
  ? {
      stt: { apiKey: null, baseUrl: null, model: null },
      tts: { apiKey: null, baseUrl: null, model: null },
    }
  : {})
```

Because the keys are redacted, this is irreversible from the UI. A user can configure two providers, temporarily switch to shared mode, save, and permanently lose both directional credentials and endpoints. The fix is to persist only `separateProviders: false` and leave `stt` / `tts` untouched, as specified.

Please add a regression test for this exact transition that starts with stored directional tuples, applies the card-generated shared-mode update, and verifies both tuples remain stored.

## Important remaining findings

- A keyless foreign directional URL can be persisted while the field note says “Using shared API key.” Runtime intentionally refuses that fallback, so the note and accepted form state disagree (`speech-settings-card.tsx:451-511`, `packages/server/src/speech/service.ts:174-220`). Validate complete directional tuples or report the incomplete state accurately.
- The directional sanitizer repeats the existing shared sanitizer's fail-open behavior: `packages/server/src/settings/public-config.ts:15-32` deletes `apiKey` only when it is a non-empty string. Since this PR adds two new secret-bearing properties, it would be safer to delete every `apiKey` property unconditionally and derive `hasApiKey` separately.
- Test Input still exercises persisted settings rather than the visible drafts. That is potentially surprising beside editable provider fields, but it is not required by the original provider-separation design.

## non blocking findings

- Requiring key re-entry whenever an endpoint changes is not an established invariant in CodeNomad: the existing shared provider on `dev` also retains its redacted key when its Base URL changes. Treating the directional equivalent as a new P0 was not justified without a broader credential-policy change.
- Draft synchronization after external settings updates and edits made during an in-flight save are existing behaviors in the `dev` speech card. The PR extends the affected form but did not introduce those mechanisms; they should be handled as separate settings-form hardening work.
- Persisting displayed fallback models may pin them, but the original design explicitly calls for pre-populating directional values from shared values when enabling separate mode. Without a clearer product requirement, that is not a merge-blocking regression.

## Validation

- Focused server/UI tests: passed, `39/39`.
- UI typecheck: passed.
- UI production build: passed; only existing chunk-size warnings.
- GitHub build matrix: all required jobs passed.

## Verdict

**Request changes. Do not merge yet.** The remaining gate is narrow: preserve the directional configuration on the separate → shared transition, exactly as required by the original design. After that fix and a transition-level regression test, this PR should receive a final focused re-review rather than another open-ended audit.

pascalandr:
## Gatekeeper re-review of `f7d60c48`

This is the focused follow-up to https://github.com/NeuralNomadsAI/CodeNomad/pull/579#issuecomment-4948342983. I reviewed the six commits since `0b7f00de`, traced the card-generated patches through RFC 7396 persistence and the runtime resolver, reran the focused test suite and UI typecheck, and checked the completed GitHub matrix.

## Resolved former blocker

The separate → shared transition now preserves both dormant directional configurations. The shared branch no longer includes `stt` or `tts` cleanup objects (`packages/ui/src/components/settings/speech-settings-card.tsx:264-275`), and `buildSpeechPatch` emits only explicitly supplied fields. Applying that patch changes `separateProviders` to `false` while leaving both stored tuples intact, as required by the design.

## Blocking findings

1. **P1: changing only a directional URL silently and irreversibly deletes its stored credential.**

`buildDirSave` treats a changed URL beside a redacted stored key as an implicit key-clear operation (`packages/ui/src/components/settings/speech-settings-card.tsx:245-250`):

```ts
...(newKey
  ? { apiKey: newKey }
  : baseUrlChanged && storedDir.hasApiKey
    ? { apiKey: null }
    : {}),
```

Reproduction:

1. Store `stt: { apiKey: "stt-secret", baseUrl: "https://old/v1" }`.
2. Open the card; the key is redacted and `stt.hasApiKey` is true.
3. Change only the STT URL to `https://new/v1` and save.
4. The card sends `stt: { apiKey: null, baseUrl: "https://new/v1", ... }`.
5. Merge-patch deletes the secret, and the runtime then rejects the remaining foreign URL without a directional key (`packages/server/src/speech/service.ts:173-190`).

Requiring a replacement key when an endpoint changes is a defensible credential policy. Silently executing the existing key's destructive clear operation is not. Block save with a clear validation message until a replacement key is entered, or require explicit confirmation before deletion.

2. **P1: a newly entered shared key is discarded if the user enables separate mode before saving.**

The latest code intentionally carries unsaved shared URL and model edits into the separate-mode update (`speech-settings-card.tsx:257-259`), but omits both `current.apiKey` and `clearStoredApiKey`. Starting without a shared key, entering one, toggling separate providers, and saving therefore persists the URL/models and mode but silently loses the entered key. With no complete directional tuples, both directions remain unconfigured. The same path ignores a pending explicit shared-key clear.

The mode-switch save needs to preserve all edited shared fields consistently, including a newly entered or explicitly cleared shared key.

## Required regression coverage

The transition-level test explicitly requested in the prior review is still absent. `packages/ui/src/stores/preferences.test.ts:105-119` calls `buildSpeechPatch` directly; it does not render `SpeechSettingsCard`, invoke `handleSave`, start with stored directional tuples, or inspect the card-generated update. The old directional-cleanup helper test also remains even though cleanup is no longer the approved transition.

Please add card-level or extracted pure payload-builder tests that cover at least:

- stored STT/TTS tuples survive separate → shared;
- redacted stored key plus URL edit cannot silently delete the key;
- a newly entered shared key survives shared → separate.

## Non-blocking follow-ups

- A keyless foreign directional URL can still be saved while the note says the shared key is used, although runtime rejects that fallback.
- Directional public-config sanitization still deletes `apiKey` only when it is a non-empty string (`packages/server/src/settings/public-config.ts:23-33`). Delete the property unconditionally after deriving `hasApiKey` to avoid leaking malformed secret values. The shared sanitizer has the same inherited weakness.
- Inputs remain editable while save is in flight; only the checkbox handler is guarded. This save/reset race is inherited from `dev`, so it is not a PR-specific merge blocker despite the latest commit's claim that inputs are disabled.

## Validation

- Focused server/UI tests: passed, `39/39`.
- UI typecheck: passed.
- `git diff --check origin/dev...f7d60c48`: passed.
- GitHub build matrix: all required jobs passed.
- Two independent scenario audits agreed that the former blocker is fixed and that the URL-edit path causes destructive credential loss.

## Verdict

**Request changes. Do not merge `f7d60c48` yet.** The original mode-transition blocker is fixed, but the latest credential-safety implementation introduces a silent destructive operation, shared-key edits are lost across the opposite transition, and the requested card-level regression test is still missing.

pascalandr (dismissed):
Gatekeeper review: changes requested.

1. HIGH - directional credentials can be redirected by a URL-only settings patch (packages/server/src/server/routes/settings.ts:38-43, packages/server/src/speech/service.ts:168-220). Store a directional key+URL, PATCH only speech.stt.baseUrl or speech.tts.baseUrl, then call transcription/synthesis: RFC merge-patch preserves the old hidden key and the provider sends it to the new host. The UI guard is not a server security boundary. Enforce key/URL coupling when persisting server settings: changing a directional destination must require a replacement key in the same patch or atomically clear the stored key. Add route-level regression coverage.

2. HIGH - new clients disable speech against older remote servers (packages/ui/src/components/prompt-input/usePromptVoiceInput.ts:50, packages/ui/src/lib/hooks/use-speech.ts:108,139, packages/ui/src/stores/conversation-speech.ts:103,294). Older servers return only configured; the new required sttConfigured/	tsConfigured fields are undefined and every directional gate becomes false. Remote profiles allow client/server version skew. Normalize missing directional flags from legacy configured at the wire boundary and test a legacy capability payload.

3. MEDIUM - changing a directional URL while the direction inherits the shared key persists a broken configuration (packages/ui/src/components/settings/speech-settings-card.tsx:234-249, packages/server/src/speech/service.ts:168-220). The validation checks only saved.stt.hasApiKey/saved.tts.hasApiKey; with a shared stored key it allows a foreign directional URL without a directional key, which the server then marks unconfigured. Require a directional key whenever an inherited credential's destination changes.

4. MEDIUM - the UI action labelled as clearing only the input/output key also deletes the saved URL and model (packages/ui/src/lib/speech-dir-save.ts:8, packages/ui/src/components/settings/speech-settings-card.tsx:255-267). Preserve URL/model when clearing a key, or rename and explicitly confirm a full provider reset. Current tests codify the destructive behavior rather than the user-facing contract.

5. MEDIUM - the new transcription test can leave microphone tracks live on MediaRecorder errors (packages/ui/src/lib/hooks/use-transcription-test.ts:81-91,115-149). There is no recorder error listener, and automatic stop/finalization does not stop or clear mediaStream. Handle recorder errors and guarantee track cleanup on every terminal path. Also keep the stop control enabled while recording: it is currently disabled whenever a settings save is in flight (speech-settings-card.tsx:321-326).

6. MEDIUM - directional secret sanitization removes piKey only when it is a non-empty string (packages/server/src/settings/public-config.ts:23-34). Malformed manually edited or API-patched values remain in the public config response. Always delete the key property regardless of type and derive hasApiKey from a valid value.

The full build matrix is green, but current tests are mostly resolver/helper-level and do not exercise settings-route merge patches, mixed-version wire responses, mounted mode transitions, or recorder error lifecycle. The branch is two commits behind current dev but merges mechanically without changed-path overlap; that is not a blocker.

pascalandr (dismissed):
Corrected gatekeeper review after auditing the complete PR history and prior review decisions.

1. HIGH - mixed-version remote servers lose all speech controls. Older servers return configured but do not return sttConfigured or ttsConfigured. The new UI requires those fields directly in usePromptVoiceInput.ts, use-transcription-test.ts, use-speech.ts, conversation-speech.ts, and speech-settings-card.tsx. At runtime they are undefined, so a configured older remote server loses STT and TTS. Remote profiles permit client/server version skew. Normalize missing directional flags from legacy configured at the response boundary while preserving explicit false values from new servers, and add a legacy capability payload regression test.

2. HIGH - MediaRecorder terminal error paths can leave microphone tracks active. use-transcription-test.ts listens for dataavailable and stop but not error. An asynchronous recorder error is followed by stop; finalizeRecording clears the recorder but neither stops nor clears mediaStream. The hook can return to idle while the microphone remains live, and a subsequent start can overwrite the only stream reference. Guarantee track cleanup on every terminal recorder path and add lifecycle tests. Related: speech-settings-card.tsx disables the Test Input button whenever a settings save is running, even while recording, while Save remains clickable; a slow save removes the only direct stop control. Keep stop available during active recording or prevent save from entering that state.

The earlier endpoint/key, tuple-clear, keyless URL, malformed-sanitizer, and save-race observations were already analyzed in this PR history as accepted policy, intentional tuple semantics, or non-blocking inherited hardening. They are not gates in this corrected review.

CI is green on head 10989d89. The branch integrates mechanically with current dev; upstream divergence is not a blocker.

pascalandr (commented):
Non-blocking follow-ups retained from the broader audit:

- Server-side endpoint/key coupling: URL-change protection currently lives in the UI, while direct settings patches retain a stored key. This matches the pre-existing shared-provider policy, so it is not a PR-specific gate, but enforcing the invariant at the persistence boundary would provide stronger defense in depth.

- Inherited shared-key UX: separate mode can persist a foreign directional URL without a directional key; runtime safely refuses the fallback, but the resulting direction is unconfigured and the UI should explain or prevent that incomplete state.

- Clear-key semantics: clearing a directional key intentionally clears URL and model as an atomic tuple. This avoids an orphaned foreign endpoint, but the current action label sounds narrower than the destructive behavior. Consider renaming it to reset provider settings or adding explicit confirmation.

- Fail-closed sanitization: public-config sanitization removes apiKey only when it is a non-empty string. This repeats inherited shared-provider behavior and is not a merge blocker, but all secret properties should eventually be deleted regardless of malformed input type, with hasApiKey derived separately.

These are recommendations, not conditions for resolving the current CHANGES_REQUESTED review. The two actual gates remain mixed-version capability fallback and complete MediaRecorder/track cleanup.

pascalandr (changes requested):
## Gatekeeper re-review of 10989d89

I re-audited the complete PR history, prior gatekeeper comments, subsequent fixes, current runtime behavior, and the completed CI matrix. The earlier broad review overclassified several inherited or intentionally accepted behaviors; this consolidated review separates the two actual gates from useful non-blocking follow-ups.

## Blocking findings

1. **P1: a new client disables all speech controls against an older configured remote server.**

Older servers expose configured but not sttConfigured or ttsConfigured. The new UI requires the directional fields directly in packages/ui/src/components/prompt-input/usePromptVoiceInput.ts:50, packages/ui/src/lib/hooks/use-speech.ts:108 and 139, packages/ui/src/stores/conversation-speech.ts:103 and 294, and packages/ui/src/lib/hooks/use-transcription-test.ts:29. At runtime those missing fields are undefined, so prompt transcription, message playback, conversation speech, and settings tests are disabled even though the older server supports them.

Remote profiles permit client/server version skew. Normalize absent directional fields from the legacy configured value at the response boundary, while preserving explicit false values returned by new servers. Add a regression test using the legacy capability payload.

2. **P1: MediaRecorder terminal errors can return the test to idle while leaving the microphone active.**

packages/ui/src/lib/hooks/use-transcription-test.ts:81-91 listens for dataavailable and stop but not error. An asynchronous recorder failure is followed by stop; finalizeRecording at lines 115-149 clears mediaRecorder but neither stops nor clears mediaStream. The hook can therefore return to idle with a live track, and another start can overwrite the only reference to that stream.

Guarantee track cleanup on every terminal recorder path and add lifecycle coverage for recorder errors and automatic stop. Relatedly, packages/ui/src/components/settings/speech-settings-card.tsx:321-326 disables Test Input whenever a settings save runs, including while recording, while Save remains available. Keep the stop action enabled during recording or prevent save from creating that state.

## Important non-blocking follow-ups

- **Server-side endpoint/key coupling:** directional URL-change protection currently lives in the UI, while a direct settings patch retains the stored key. This matches the pre-existing shared-provider policy and is therefore not a PR-specific gate, but enforcing the invariant at persistence would provide stronger defense in depth.

- **Inherited shared-key UX:** separate mode can persist a foreign directional URL without a directional key. Runtime safely refuses to send the shared key, but the resulting direction is unconfigured and the UI should explain or prevent that incomplete state.

- **Clear-key wording:** clearing a directional key intentionally clears its URL and model as an atomic tuple. That avoids an orphaned foreign endpoint, but the action label sounds narrower than the destructive behavior. Consider naming it Reset provider settings or adding confirmation.

- **Fail-closed sanitization:** public-config sanitization removes apiKey only when it is a non-empty string. This repeats inherited shared-provider behavior and is not a merge blocker, but secret properties should eventually be removed regardless of malformed input type, with hasApiKey derived separately.

## Validation

- Current head: 10989d89152950ee0b47fec885da500d6a724a00.
- Full GitHub build matrix: passed.
- Existing focused server and UI tests: passed according to the PR history.
- git diff --check: passed.
- Integration with current dev: mechanically clean with no changed-path overlap; branch divergence is not a blocker.

## Verdict

**Request changes. Do not merge yet.** Resolve the mixed-version capability fallback and complete microphone lifecycle cleanup. The four follow-ups above are recommendations, not conditions for approval.

pascalandr (changes requested):
## Gatekeeper re-review of 5d6da7c0

This review follows up on the two gates from the previous consolidated review and audits only the seven corrective commits added since 10989d89.

## Resolved

- New-client/old-server speech compatibility is behaviorally fixed. Missing sttConfigured and ttsConfigured values inherit the legacy configured value, explicit false values are preserved, and unsupported separate-provider controls are gated.
- MediaRecorder error handling now releases the recorder and microphone stream. The original direct error-path microphone leak is fixed.

## Blocking findings

1. **P1: a new server clears stored speech credentials when an older client saves unrelated settings.**

packages/server/src/server/routes/settings.ts:22-40 treats the presence of speech.baseUrl, speech.stt.baseUrl, or speech.tts.baseUrl as proof that the URL changed and injects apiKey: null whenever the patch omits a key. Clients from before this PR always send the complete normalized speech object, including an unchanged baseUrl, while the stored key is intentionally absent from the sanitized response. Saving only voice, model, format, or playback mode from such a client therefore deletes the stored key and disables speech.

The persistence boundary must compare the submitted URL with the currently stored URL before clearing the paired key. Add route-level coverage for an old-client full speech patch containing an unchanged URL and no key, for both shared and directional configuration.

2. **P1: saving still disables the only Stop control during an active recording.**

packages/ui/src/components/settings/speech-settings-card.tsx:306-310 disables Test Input whenever isSaving is true, while lines 364-370 leave Save available during recording. A user can start recording and then save; while the settings request or capability refresh is pending, the microphone continues recording with no in-card stop action. This was explicitly part of the previous MediaRecorder gate and remains unresolved.

Keep Test Input enabled when its state is recording, or disable Save while transcription is requesting/recording/transcribing. Add a control-state regression test.

3. **Required CI is red on Linux.**

The branch is three commits behind dev. The Linux job uses the post-#492 validation that requires one Electron tar.gz, while this head still builds zip/AppImage, producing: Expected one Electron Linux tar.gz, found 0. Merge or rebase current dev and rerun the complete matrix.

## Non-blocking follow-ups

- packages/ui/src/lib/speech-capabilities.test.ts duplicates the production normalizer instead of importing or exercising the store boundary, so it can pass if production normalization regresses.
- The new settings recorder has no lifecycle tests for event ordering, automatic stop, repeated recordings, or unmount. Coordinate with PR #580 so the settings recorder does not retain the same subsequent-recording empty-audio behavior being fixed there.

## Validation

- 31 focused server speech/settings tests passed.
- 29 focused UI speech/preferences tests passed.
- Server and UI typechecks passed.
- git diff --check passed.
- macOS, Windows, and Tauri jobs passed; Electron Linux failed as described above.

## Verdict

**Request changes.** The previous mixed-version capability blocker is resolved, but credential preservation for old clients and the recording/save stop-control race must be fixed, then the branch must be synchronized with dev for a green required matrix.

pascalandr (changes requested):
## Gatekeeper follow-up of 69aac47c


## Resolved

- The owner-scoped server settings route now compares submitted URLs against stored URLs, so old-client full speech objects no longer clear credentials when the URL is unchanged.
- Test Input remains enabled while actively recording during a settings save, preserving the Stop action.
- The branch is synchronized with dev and the complete GitHub build matrix is green, including Electron Linux.

## Remaining blocker

**P1: the full-document config PATCH still bypasses speech credential pairing.**

packages/server/src/server/routes/settings.ts:55-58 still calls enforceSpeechCredentialPairing(request.body ?? {}) without the current config. More importantly, a full-document payload is shaped as server.speech, while the helper only inspects a top-level speech property. A direct PATCH to /api/storage/config with server.speech.baseUrl changed and no apiKey therefore reaches mergePatchDoc unchanged, preserving the old stored key and redirecting it to the new endpoint.

Reproduction against the production helper returns the input unchanged:

    body    = { server: { speech: { baseUrl: newUrl } } }
    current = { server: { speech: { baseUrl: oldUrl, apiKey: secret } } }

The full-document route must apply the invariant to its server owner patch using deps.settings.getOwner(config, server), just as the owner-scoped route now does. Add route-level persistence tests for both PATCH /api/storage/config and PATCH /api/storage/config/server; the current tests invoke only the helper with owner-shaped payloads and cannot catch this bypass.

## Validation

- 31 focused server speech/settings tests passed.
- 29 focused UI speech/preferences tests passed.
- Server and UI typechecks passed.
- Full GitHub matrix passed.
- git diff --check passed.

## Verdict

**Request changes.** The previous user-facing and old-client regressions are fixed. Close the remaining full-document security bypass, add route-level coverage, and this should be ready for approval.

pascalandr (approved):
## Final gatekeeper review of 8fb7403b

No blocking findings remain.

## Verified fixes

- New-client/old-server capability fallback preserves configured speech controls and explicit directional false values.
- Old-client/new-server saves preserve stored credentials when shared or directional URLs are unchanged.
- Shared and directional URL changes clear an omitted paired key at the server persistence boundary.
- Both PATCH /api/storage/config/server and the full-document PATCH /api/storage/config now apply credential pairing against the current server owner.
- MediaRecorder error paths release microphone resources, and the Stop action remains available during an active recording even while settings are saving.
- The branch is synchronized with dev and the complete required build matrix is green.

## Validation

- 33 focused server speech/settings tests passed.
- 29 focused UI speech/preferences tests passed.
- Server and UI typechecks passed.
- git diff --check passed.
- Electron Linux/macOS/Windows and Tauri Linux/macOS/macOS ARM64/Windows passed.

## Non-blocking residual

The two new doc-level cases exercise the production pairing helper with document-shaped data rather than Fastify route injection. The route wiring is straightforward and manually verified, but a future route-level persistence test would provide stronger regression coverage.

## Verdict

**Approved.** The previous security, mixed-version, recorder lifecycle, and CI gates are resolved.

pascalandr:
Thanks for your patience and tenacity @heunghingwan !

PR #580, opened 2026-07-11
fix(ui): cancel pending microphone acquisition on early release (#550)

### LINKED ISSUE #550: Speech-to-text fails on subsequent recordings — MediaRecorder sends near-empty audio chunks
## Description

Speech-to-text (STT) transcription works correctly on the **first** recording attempt, but consistently **fails on all subsequent attempts** within the same session. This affects the Linux Electron desktop app (v0.15.0) but does not reproduce on mobile browsers.

## Steps to reproduce

1. Open CodeNomad on Linux
2. Click the microphone button and speak → transcription succeeds
3. Click the microphone button a second time and speak → transcription fails with no visible error
4. Subsequent attempts also fail
5. Refresh the app (Ctrl+R) → first attempt works again → subsequent ones fail again

## Observed behavior

Server debug logs (`--log-level debug`) show:

| Attempt | Audio payload size | Result |
|---------|-------------------|--------|
| 1st | **61,154 bytes** | 200 OK (965ms) |
| 2nd | **110 bytes** | 400 "The audio file could not be decoded or its format is not supported" |
| 3rd | **110 bytes** | 400 same error |

The 110-byte payload is just the WebM container header with no actual audio data.

Relevant server logs:
```
[INFO] speech.transcribe mimeType=audio/webm;codecs=opus bytes=61154 → status=200
[INFO] speech.transcribe mimeType=audio/webm;codecs=opus bytes=110
[WARN] speech.transcribe verbose_json failed; retrying default format
  err=BadRequestError: 400 The audio file could not be decoded or its format is not supported.
→ status=502
```

## Root cause

The `MediaRecorder` (or its underlying `MediaStream`) is reused between recordings without being re-created. On Chromium/Linux, `stop()` → `start()` on the same `MediaRecorder` instance does not properly reset the audio capture pipeline, resulting in near-empty audio chunks (~110 bytes) on all attempts after the first.

The same code path works correctly on mobile (Android/iOS) because their WebRTC/MediaRecorder implementations handle the lifecycle differently.

## Expected behavior

Each microphone session should produce a valid audio recording, regardless of how many times the user starts/stops recording.

## Suggested fix

Re-create a fresh `MediaStream` and `MediaRecorder` instance for each recording session rather than reusing the previous instance. Ensure proper cleanup (`stream.getTracks().forEach(t => t.stop())`) of the previous stream before creating a new one.

## Environment

- **CodeNomad version**: 0.15.0 (Electron)
- **OS**: Arch Linux, PipeWire 1.6.6
- **Browser/Engine**: Electron/Chromium
- **Microphone**: Roland UA-22 (USB audio interface), also reproduced with USB webcam mic
- **STT provider**: OpenAI whisper-1 via openai-compatible adapter

heunghingwan:
## Problem

The voice input button uses push-to-talk: press to start recording, release to stop. On the **second and subsequent recordings**, the button fails to enter the recording state (doesn't turn red) and jumps directly to "transcribing", producing empty or header-only audio (~110 bytes) that the STT provider rejects with HTTP 400.

This affects **all platforms** (Linux/Electron, Windows/Chrome), not just Linux as originally reported in #550.

## Root Cause

`startRecording()` is async — it awaits `getUserMedia()`. But `beginVoicePress()` in `prompt-input.tsx` fires it with `void` (fire-and-forget):

```typescript
void voiceInput.startRecording()   // not awaited
```

When the user releases the button before `getUserMedia()` resolves (common on second+ recordings where mic permission is already granted and acquisition is fast), `stopRecording()` finds no active recorder (`mediaRecorder === null`, `state() === "idle"`) and **silently returns**. The pending `getUserMedia()` promise later resolves and starts an **orphan recording** with no matching stop.

The user perceives: button doesn't turn red → presses again → the orphan recorder is stopped → jumps to "transcribing". On fast taps, the orphan recording captures zero audio frames, producing a header-only WebM blob (~110 bytes).

### Why it only manifests on second+ recordings

On the **first** recording, `getUserMedia()` may show a permission prompt or have slower initial device setup, giving the user time to hold the button. On **subsequent** recordings, the device is already warm and `getUserMedia()` resolves faster, but the user's tap duration remains the same — releasing before acquisition completes.

### What this is NOT

- **Not a PipeWire/Linux-specific issue** — reproduced on Windows Chrome (server on Linux, browser on Windows)
- **Not a MediaRecorder encoder bug** — tested with Chrome's fake audio device: `start()` vs `start(timeslice)` produce identical results across consecutive recordings
- **Not a recorder reuse issue** — the code already creates fresh `MediaStream` + `MediaRecorder` per recording

## Fix

Add a `requestGeneration` counter (mirroring the pattern from `use-transcription-test.ts` in PR #579):

- `startRecording()` increments the generation before awaiting `getUserMedia()`
- After the await resolves, if the generation is stale, the stream is immediately stopped and discarded
- `stopRecording()` / `cancelRecording()` increment the generation when called while `state() === "idle"` (pending acquisition), immediately invalidating any in-flight request
- All event listeners (`dataavailable`, `stop`) and `finalizeRecording()` check the generation before mutating state
- `cleanupMedia()` and `onCleanup()` increment the generation to cover unmount and error paths

Additionally, `stopTracks(stream)` is called **before** the transcription network round-trip (instead of after), releasing the audio device sooner.

## Verification

- `tsc --noEmit` passes (packages/ui)
- Code review: no regressions, no stream leaks, no double-free paths found
- Reproduction evidence: confirmed the 110-byte header-only blob signature matches the orphan-recording failure mode

Closes #550

### DIFF
--- a/packages/ui/src/components/prompt-input/recording-controller.test.ts
+++ b/packages/ui/src/components/prompt-input/recording-controller.test.ts
@@ -1,0 +1,275 @@
+import assert from "node:assert/strict"
+import { describe, it } from "node:test"
+
+import { createRecordingController } from "./recording-controller.ts"
+import type { RecordingControllerDeps, RecordingControllerCallbacks, StreamHandle } from "./recording-controller.ts"
+
+function createMockStream(): StreamHandle & { tracksStopped: boolean } {
+  return {
+    getTracks: () => [{ stop: () => {} }],
+    tracksStopped: false,
+  }
+}
+
+interface Harness {
+  deps: RecordingControllerDeps
+  callbacks: RecordingControllerCallbacks & {
+    states: string[]
+    transcripts: string[]
+    errors: { key: string; detail?: string }[]
+  }
+  recorder: { startCalled: boolean; stopCalled: boolean }
+}
+
+function createHarness(options?: {
+  getUserMedia?: RecordingControllerDeps["getUserMedia"]
+  setupRecorder?: RecordingControllerDeps["setupRecorder"]
+  transcribe?: RecordingControllerDeps["transcribe"]
+}): Harness {
+  const recorder = { startCalled: false, stopCalled: false }
+
+  const cbState = {
+    states: [] as string[],
+    transcripts: [] as string[],
+    errors: [] as { key: string; detail?: string }[],
+  }
+
+  const callbacks = {
+    ...cbState,
+    onStateChange: (s: string) => cbState.states.push(s),
+    onElapsedChange: () => {},
+    onTranscript: (text: string) => cbState.transcripts.push(text),
+    onError: (key: string, detail?: string) => cbState.errors.push({ key, detail }),
+  }
+
+  const deps: RecordingControllerDeps = {
+    isElectron: () => false,
+    requestMicAccess: async () => undefined,
+    getUserMedia: options?.getUserMedia ?? (async () => createMockStream()),
+    setupRecorder:
+      options?.setupRecorder ??
+      ((_stream, handlers) => ({
+        start: () => {
+          recorder.startCalled = true
+          handlers.onDataAvailable(new Blob(["audio"], { type: "audio/webm" }))
+        },
+        stop: () => {
+          recorder.stopCalled = true
+          queueMicrotask(() => handlers.onStop())
+        },
+        mimeType: "audio/webm;codecs=opus",
+      })),
+    stopTracks: (stream) => {
+      if (stream && "tracksStopped" in stream) {
+        ;(stream as unknown as { tracksStopped: boolean }).tracksStopped = true
+      }
+    },
+    blobToBase64: async () => "base64data",
+    transcribe: options?.transcribe ?? (async () => ({ text: "hello world" })),
+    setInterval: () => 0,
+    clearInterval: () => {},
+  }
+
+  return { deps, callbacks, recorder }
+}
+
+function flushPromises() {
+  return new Promise((resolve) => setTimeout(resolve, 0))
+}
+
+describe("recording controller", () => {
+  it("starts recording after getUserMedia resolves", async () => {
+    let resolveGum!: (s: StreamHandle) => void
+    const h = createHarness({
+      getUserMedia: () => new Promise((r) => { resolveGum = r }),
+    })
+
+    const c = createRecordingController(h.deps, h.callbacks)
+
+    const p = c.startRecording()
+    assert.equal(c.getState(), "idle")
+
+    resolveGum(createMockStream())
+    await p
+
+    assert.equal(c.getState(), "recording")
+    assert.equal(h.recorder.startCalled, true)
+    assert.deepEqual(h.callbacks.states, ["recording"])
+  })
+
+  it("cancels pending acquisition when stopped before getUserMedia resolves", async () => {
+    let resolveGum!: (s: StreamHandle) => void
+    const h = createHarness({
+      getUserMedia: () => new Promise((r) => { resolveGum = r }),
+    })
+
+    const c = createRecordingController(h.deps, h.callbacks)
+
+    const p = c.startRecording()
+    c.stopRecording()
+    resolveGum(createMockStream())
+    await p
+    await flushPromises()
+
+    assert.equal(c.getState(), "idle")
+    assert.equal(h.recorder.startCalled, false)
+  })
+
+  it("stops the returned stream when cancelled before resolution", async () => {
+    let resolveGum!: (s: StreamHandle) => void
+    const h = createHarness({
+      getUserMedia: () => new Promise((r) => { resolveGum = r }),
+    })
+
+    const c = createRecordingController(h.deps, h.callbacks)
+    const p = c.startRecording()
+    c.stopRecording()
+
+    const stream = createMockStream()
+    resolveGum(stream)
+    await p
+    await flushPromises()
+
+    assert.equal(stream.tracksStopped, true)
+  })
+
+  it("transcribes and reports transcript on normal stop", async () => {
+    const h = createHarness()
+    const c = createRecordingController(h.deps, h.callbacks)
+
+    await c.startRecording()
+    assert.equal(c.getState(), "recording")
+
+    c.stopRecording()
+    await flushPromises()
+
+    assert.deepEqual(h.callbacks.transcripts, ["hello world"])
+    assert.equal(c.getState(), "idle")
+  })
+
+  it("does not transcribe when no audio chunks collected", async () => {
+    const h = createHarness({
+      setupRecorder: (_stream, handlers) => ({
+        start: () => {},
+        stop: () => { queueMicrotask(() => handlers.onStop()) },
+        mimeType: "audio/webm",
+      }),
+    })
+    const c = createRecordingController(h.deps, h.callbacks)
+
+    await c.startRecording()
+    c.stopRecording()
+    await flushPromises()
+
+    assert.deepEqual(h.callbacks.transcripts, [])
+    assert.equal(c.getState(), "idle")
+  })
+
+  it("allows next recording after cancelled start", async () => {
+    let resolveGum1!: (s: StreamHandle) => void
+    let firstCall = true
+    const h = createHarness({
+      getUserMedia: () => {
+        if (firstCall) {
+          firstCall = false
+          return new Promise((r) => { resolveGum1 = r })
+        }
+        return Promise.resolve(createMockStream())
+      },
+    })
+
+    const c = createRecordingController(h.deps, h.callbacks)
+    const p = c.startRecording()
+    c.stopRecording()
+
+    resolveGum1(createMockStream())
+    await p
+    await flushPromises()
+
+    assert.equal(c.getState(), "idle")
+    assert.equal(h.recorder.startCalled, false)
+
+    await c.startRecording()
+    assert.equal(c.getState(), "recording")
+    assert.equal(h.recorder.startCalled, true)
+  })
+
+  it("prevents start while transcribing", async () => {
+    let resolveTranscribe!: () => void
+    const h = createHarness({
+      transcribe: () => new Promise((r) => { resolveTranscribe = () => r({ text: "hello" }) }),
+    })
+
+    const c = createRecordingController(h.deps, h.callbacks)
+
+    await c.startRecording()
+    c.stopRecording()
+    await flushPromises()
+
+    assert.equal(c.getState(), "transcribing")
+
+    await c.startRecording()
+    assert.equal(c.getState(), "transcribing")
+
+    resolveTranscribe()
+    await flushPromises()
+    assert.equal(c.getState(), "idle")
+  })
+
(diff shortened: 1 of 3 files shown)


pascalandr:
Gatekeeper review of latest head `df9172c1`.

I reviewed this as a platform-specific MediaRecorder lifecycle fix, focusing on regressions, evidence for the proposed root cause, and integration with the other active speech PR.

## Blocking findings

1. **P1: the fix has not been validated on the affected Linux/Electron path.**

The primary behavioral change is `mediaRecorder.start(TIMESLICE_MS)` at `packages/ui/src/components/prompt-input/usePromptVoiceInput.ts:123`, but the PR does not demonstrate that current `dev` still reproduces the issue or that this patch fixes multiple consecutive recordings on Linux/Electron.

Issue #550 reports CodeNomad 0.15.0, while the current application and Electron version have changed. Green packaging checks do not execute MediaRecorder. Before merge, test the Linux/Electron artifact with at least three consecutive recordings and inspect the resulting payload/container contents, not only whether the request returns successfully.

2. **P1: the fixed 250-byte threshold can reject valid recordings and does not reliably detect empty audio.**

`packages/ui/src/components/prompt-input/usePromptVoiceInput.ts:8-9` and `:150-161` classify every blob below 250 bytes as no audio. The recorder can select WebM, MP4, Ogg, or a browser-default format at `:259-262`.

A format-independent byte count cannot establish whether encoded audio frames exist:

- a very short valid recording can be below the threshold;
- an empty MP4/Ogg container can exceed it;
- container metadata sizes vary by browser and codec.

Restrict the heuristic to the observed WebM failure signature, use format-aware validation, or remove this guard from the primary fix.

3. **P1: releasing the push-to-talk button while microphone acquisition is pending can start an orphan recording.**

`startRecording()` remains effectively idle while awaiting Electron permission and `getUserMedia()` at `usePromptVoiceInput.ts:84-106`. The caller releases the button through `packages/ui/src/components/prompt-input.tsx:784-787`, but `stopRecording()` returns because no recorder exists yet (`usePromptVoiceInput.ts:66-67`). When the pending promise resolves, the continuation still starts recording at `:106-123`, with no matching future stop.

The same problem occurs if the component is unmounted while acquisition is pending: cleanup cannot invalidate a stream that has not resolved yet. This race predates the PR, but the new timeslice makes an orphan recorder periodically append chunks while keeping the microphone active.

Add a `requesting` state or request-generation token, invalidate it on release/unmount, and immediately stop streams returned for stale requests.

4. **P1 integration risk: PR #579 introduces a second recorder path without this mitigation.**

PR #579 adds `packages/ui/src/lib/hooks/use-transcription-test.ts`, which still starts MediaRecorder without a timeslice, stops without `requestData()`, and uploads without checking for the observed header-only payload. A synthetic merge does not surface this semantically: prompt dictation gets the workaround while Settings → Test input retains the same failure mode.

Put the recorder start/finalization policy in the shared audio utility introduced by #579, or apply the mitigation consistently to both paths before both PRs merge.

## Important findings

5. **The claimed root cause is not established by the available evidence.**

The observed 61 KB first payload followed by 110-byte payloads establishes a client capture/encoding failure. It does not establish that omitting a timeslice specifically causes Chromium's encoder initialization failure. The earlier #550 investigation described timeslicing as a defensive mitigation to test because no matching Electron/Chromium report was found.

Until the affected Linux/Electron scenario is reproduced and verified, this should be described as a workaround rather than a proven root-cause fix.

6. **There is no regression coverage for the changed lifecycle.**

No test covers:

- `requestData()` followed by `stop()`;
- collection of periodic and final `dataavailable` events;
- the `<250` branch;
- release/unmount while permission is pending;
- track release before upload.

A fake MediaRecorder test can cover event ordering and cleanup even though the Chromium-specific encoder issue remains a manual platform test.

## What looks correct

- `requestData()` followed by `stop()` has valid specified event ordering.
- Collecting timeslice chunks and composing one Blob is valid MediaRecorder usage.
- Recordings shorter than the timeslice should still receive the final stop flush.
- Releasing microphone tracks before the network request is an improvement.
- The new i18n key is present in all nine registered locales.

## Validation

- Current GitHub build/package checks are green.
- No automated MediaRecorder test exists.
- No Linux/Electron verification of the reported repeated-recording scenario is included.

## Verdict

Do not merge yet. The `start(1000)` workaround is plausible, but this platform-specific fix needs validation on the affected artifact, the byte-size heuristic needs to be removed or made format-aware, pending microphone acquisition must be cancellable, and the recorder path introduced by #579 must not retain the same failure mode.

heunghingwan:
Thanks for the thorough review @pascalandr. The PR has been completely reworked based on your feedback and new evidence.

## Revised root cause

The original hypothesis (timeslice encoder flush) was **wrong**. The actual root cause is a **push-to-talk race condition** in `usePromptVoiceInput.ts`:

`beginVoicePress()` fires `startRecording()` with `void` (not awaited). When the user releases the button before `getUserMedia()` resolves, `stopRecording()` finds no active recorder and silently returns. The pending `getUserMedia()` later resolves into an **orphan recording** with no matching stop.

This explains:
- **P1-1**: Why it affects second+ recordings — on the first recording, `getUserMedia()` is slower (initial device setup), so the user holds long enough. On subsequent recordings, acquisition is faster but tap duration stays the same.
- **The original #550 report**: Not Linux/PipeWire-specific. A maintainer confirmed the same symptom on **Windows Chrome** (server on Linux, browser on Windows, audio processing entirely client-side on Windows).
- **The 110-byte blobs**: When the orphan recording is stopped almost immediately after starting, zero audio frames reach the encoder → header-only WebM container.

### What was removed

Based on your feedback, all speculative changes were removed:
- ~~`mediaRecorder.start(TIMESLICE_MS)`~~ — timeslice doesn't fix the race and produced 0-byte blobs for short recordings (worse than 110)
- ~~`requestData()` before `stop()`~~ — was a speculative companion to timeslice
- ~~`MIN_AUDIO_BYTES` guard~~ — format-blind (P1-2 was correct), and no longer needed since the race is prevented at the source

## P1-3 (orphan recording race) — fixed

This was the core finding. Added a `requestGeneration` counter (mirroring the pattern from `use-transcription-test.ts` in #579):

- `startRecording()` increments generation before `await getUserMedia()`
- After the await, if generation is stale, the stream is immediately `stopTracks`'d and discarded
- `stopRecording()` increments generation when `state() === "idle"` (pending acquisition), cancelling any in-flight request
- `cleanupMedia()` and `onCleanup()` also increment it

## P1-4 (PR #579 second recorder path)

`use-transcription-test.ts` from #579 already has the `requestGeneration` pattern and is unaffected. Since the timeslice approach was removed, there's nothing to share between the two paths. Both now use the same cancellation approach.

## P1-2 (byte-size heuristic)

Removed entirely as you suggested.

## Early `stopTracks()`

Kept the improvement of moving `stopTracks(stream)` before the transcription network round-trip.

## Cleanup

- Removed dead `cancelRecording()` function (zero callers)
- Removed redundant `disposed` flag (`cleanupMedia()` bumping `requestGeneration` already covers disposal)

## Tests (important-6)

Deferred for now. The project uses `node:test` with no jsdom/vitest infrastructure, and the hook imports `.tsx` modules (`useI18n`) that Node's native TypeScript stripper can't handle. All existing UI tests cover pure functions extracted from hooks, not the hooks themselves. Setting up hook testing would require either introducing vitest+jsdom or extracting the async state machine into a testable abstraction — both are larger efforts outside the scope of this fix.

pascalandr (changes requested):
## Gatekeeper re-review of da645429

The rework is substantially better and the speculative recorder changes are gone. One merge gate remains.

## Blocking finding

1. **P1: the replacement fix still has no behavioral validation.**

The new generation guard in packages/ui/src/components/prompt-input/usePromptVoiceInput.ts is logically sound: an early release invalidates the pending request, stale streams are stopped, stale recorder events cannot mutate state, and tracks are released before transcription. It also resolves the integration concern with PR #579 because both recorder paths now use the same cancellation pattern.

However, this head replaces the original root-cause hypothesis and the response explicitly defers tests. The green build/package matrix, typecheck, and bundle do not execute getUserMedia or MediaRecorder, and there is no comment showing that the new artifact was exercised. That leaves the actual #550 workflow unverified.

Before merge, provide either:

- an automated delayed-getUserMedia regression proving press -> release before resolution stops the returned stream, never creates/starts a recorder, never uploads audio, and allows the next recording; or
- a short manual result from the current da645429 artifact (Linux/Electron or the confirmed Windows/Chrome environment) covering early release plus at least three consecutive normal recordings, with no orphan microphone activity and no header-only upload.

This does not require introducing Vitest/jsdom if manual artifact verification is the proportionate option.

## Non-blocking observation

On Electron, release during requestMicrophoneAccess still proceeds to getUserMedia before the stale-generation check at lines 94-103. The resulting stream is stopped, so it cannot become an orphan recorder, but an additional generation check before getUserMedia would avoid acquiring the device after cancellation.

## Resolved from the previous review

- Removed the format-blind byte threshold.
- Removed speculative timeslicing and requestData behavior.
- Added stale request, event, finalization, cleanup, and unmount guards.
- Confirmed the #579 settings recorder already has generation protection.
- Moved track release ahead of the transcription round-trip.

## Validation

- npm run typecheck in packages/ui: passed.
- npm run build in packages/ui: passed.
- Complete required GitHub build matrix: green.

## Verdict

**Changes requested.** The code issue is addressed, but the behavioral gate from the previous review remains open until the current fix is exercised once on the affected workflow.

pascalandr (approved):
## Final gatekeeper review of d1f5c5cd

No blocking findings remain.

## Verified resolution

- The speculative timeslice, requestData, and format-blind byte-threshold changes remain removed.
- The push-to-talk lifecycle is now isolated in a dependency-injected recording controller while the Solid/browser hook remains a thin adapter.
- Releasing before getUserMedia resolves invalidates the request, prevents recorder creation/start, and stops the late stream.
- Cleanup and stale transcription results are generation-guarded.
- A subsequent recording works after a cancelled acquisition.
- The Electron permission path now checks cancellation before calling getUserMedia, resolving the previous non-blocking observation.
- Tracks are released before the transcription network round-trip.
- The recorder path added by PR #579 remains compatible with the same generation-guard approach.

## Regression coverage

The new 10-test controller suite covers normal start/stop, delayed acquisition cancellation, late stream release, no-chunk handling, recording after cancellation, transcribing lockout, transcription failure, cleanup during pending acquisition, and stale transcription completion.

## Validation

- npx tsx --test src/components/prompt-input/recording-controller.test.ts: 10/10 passed.
- npm run typecheck in packages/ui: passed.
- npm run build in packages/ui: passed.
- Electron Linux/macOS/Windows and Tauri Linux/macOS/macOS ARM64/Windows: passed.

## Non-blocking residual

The UI package still has no test script and the current packaging workflow does not execute the controller test automatically. The focused suite passes and closes this review gate, but wiring existing UI tests into CI would make the regression protection persistent.

## Verdict

**Approved.** The behavioral gate from the prior review is closed and the fix is ready to merge.

pascalandr:
Thanks @heunghingwan !

PR #581, opened 2026-07-11
fix(ui): persist environment variable removal

### LINKED ISSUE #575: [Bug]: Proxy environment variable for opencode can't be set
### App Variant

Electron

Windows 11 25H2 x64

### Issue Summary

http_proxy environment variable can be entered, but is corrupted when menu is reopened

### Steps to Reproduce

CodeNomad settings, OoenCode., environment variables.
For some reasons I have to set HTTP_PROXY=http://192.168.1.1:8118.
This works in the Windows shell.
Here, when I input the variable name, and the content, then open OpenCode in CodeNomad, Opencode can't connect.
When I reopen the settings in CodeNomad, the IP is corrupted to 'h'.

When I set http_proxy in "~\.config\codenomad\config.yaml", then start opencode and look at the envirenment variable setting, it is shown correctly.

### Expected Behavior

OpenNomad should save and set the environment variable exactly as entered.

### Logs & Screenshots

_No response_

### Additional Context

Image

pascalandr:
## Summary

- Send an explicit JSON Merge Patch deletion when removing an environment variable.
- Prevent deleted variables from remaining in config.yaml and being injected into future OpenCode instances.
- Maybe closes #575

## Root Cause

The previous removal flow submitted the remaining environment-variable record without the selected key. The settings API recursively applies JSON Merge Patch, where an omitted nested key means no change rather than deletion.

## Validation

- git diff --check


pascalandr:
Gatekeeper review for the latest PR state.

No merge-blocking findings.

What I checked:
- The diff is limited to the environment-variable removal path.
- The new null patch value matches the server's JSON Merge Patch semantics, so the selected nested key is removed instead of being retained.
- The existing secure-key retention behavior is unchanged; it does not inject a removed variable into future OpenCode instances.
- The latest PR build matrix completed successfully, including Electron Windows and Tauri Windows builds.
- Local validation passed: UI typecheck, UI production build, and git diff --check.

Residual risk:
- This focused UI store path has no automated regression test. The change is small and the persistence behavior was manually reproduced before the fix, but a future UI-store test would provide stronger protection.

pascalandr:
@shantur LGTM

PR #582, opened 2026-07-12
fix(ui): surface session loading failures

### LINKED ISSUE #543: [Bug]: Corrupted time_archived in SQLite causes HTTP 400 on session list — zero sessions shown

pascalandr:
## Summary

- preserve and display detailed OpenCode errors when the session list cannot be loaded instead of failing silently
- share error extraction and retry UI with per-session message loading while keeping each error scoped correctly
- continue instance hydration after list failures and ignore stale overlapping list responses
- show loading feedback during retries and localize the new states for every supported locale

Closes #543.

## Validation

- `bun test --conditions=browser packages/ui/src/stores/session-list-error.test.ts packages/ui/src/stores/session-tree.test.ts`
- `git diff --check upstream/dev...HEAD`


pascalandr:
## Gatekeeper review: REQUEST CHANGES

I reviewed the latest PR head `2f288290` against `dev` and found two blocking issues.

### Blocking findings

1. **Cold-start retry can permanently lose pending permission/question indicators.**

   `hydrateInstanceData()` now continues to synchronize permissions and questions after `fetchSessions()` fails. With no sessions in the store, `setSessionPendingPermission()` and `setSessionPendingQuestion()` cannot update anything. If the user later retries the session list successfully, the newly loaded sessions never reconcile the already-populated interruption queues, so affected sessions and instance indicators remain idle until another event or full rehydration.

   Expected fix: add one deterministic pending-state reconciliation step and invoke it after session-list success and queue synchronization.

2. **The new test does not run in the default Bun test environment.**

   `session-list-error.test.ts` imports the full `session-state.ts` graph, which reaches `solid-toast` and throws `Client-only API called on the server side` under `bun test packages/ui/src/stores/session-list-error.test.ts`. The PR command passes only because it adds `--conditions=browser`.

   Expected fix: move `getOpencodeErrorMessage()` coverage into a focused `opencode-api.test.ts` with no UI-store dependency, and test session-list state through a client-safe module or the project's browser-capable harness.

### Non-blocking improvements

- Clear per-instance request counters and list errors during instance teardown.
- Avoid storing localized fallback text so changing locale updates an already-visible fallback.
- Add controlled out-of-order request tests for older-success/newer-failure and older-failure/newer-success.
- Add coverage for cold-start failure followed by Retry.

### Implementation assessment

The shared `LoadErrorState` component and `opencode-api.ts` extraction boundary are appropriate. The latest-request guards and Retry loading state are improvements over the first revision. A single `reconcileSessionPendingState(instanceId)` helper is the simplest robust fix for hydration-order independence.

### Verification

- UI typecheck: passed
- Production UI build: passed
- Recursive session tree/pagination tests: passed
- `git diff --check`: passed
- Default execution of the new test: failed before assertions

I would not merge this state until both blocking findings are resolved.

pascalandr:
## Gatekeeper follow-up: PASS

Resolved both blocking findings in `822c4f44`:

- Session-list success now reconciles pending permission and question state from the interruption queues, including cold-start failure followed by Retry.
- Error extraction tests no longer import the client-only store graph and run under default Bun conditions.
- Added server-safe coverage for queue state arriving before sessions and for stale pending flags.
- Instance teardown now clears session-list error/loading/request state. Request IDs use a monotonic generation so late responses remain stale even if an instance ID is reused.

Validation:
- 12 focused tests passed
- UI typecheck passed
- Production UI build passed
- `git diff --check` passed
- Independent gatekeeper re-review: PASS

pascalandr:
@shantur LGTM

shantur:
@pascalandr - Needs to resolve conflicts

PR #583, opened 2026-07-12
fix(ui): refresh worktrees from OpenCode ready events

pascalandr:
## Problem

Worktrees created through OpenCode can leave CodeNomad's cached worktree inventory and workspace mapping stale. The previous approach in #516 polled on `session.idle` and attempted to infer which session created a newly observed worktree, which introduced race conditions and unsafe attribution.

## Fix

- route OpenCode's `worktree.ready` event through the existing UI SSE manager;
- reload the live Git worktree inventory when that event arrives;
- synchronize OpenCode workspace mappings only after the refreshed inventory is available;
- avoid idle polling and avoid auto-switching a session because the event identifies the new worktree directory but does not identify an originating session.

The server bridge already forwards arbitrary OpenCode events and preserves the global event `directory`, so no server change is required.

Supersedes #516 and implements the SSE approach suggested by @shantur  https://github.com/NeuralNomadsAI/CodeNomad/pull/516#issuecomment-4658638464.

## Validation

- `bun test --conditions browser packages/ui/src` (`126` passing)
- `git diff --check`


pascalandr:
Gatekeeper review for the latest PR state.

## Blocking findings

1. **Concurrent `worktree.ready` events can leave both caches stale.**

   `packages/ui/src/stores/worktrees.ts:109-127`
   `packages/ui/src/stores/opencode-workspaces.ts:34-73`

   Every event starts an independent `reloadWorktrees()` call. If worktrees A and B become ready close together, the request that observed `[A, B]` can finish first, followed by an older request that observed only `[A]`; the older response then overwrites the current inventory and B disappears until another refresh.

   The workspace side has a related race: `reloadOpenCodeWorkspaces()` deletes `workspaceSyncs` before starting a new synchronization, and the `finally` of an older synchronization can delete the entry belonging to a newer one.

   Serialize the complete worktree + workspace refresh per instance, or use generation checks that prevent older responses from committing. Add a regression test with two overlapping ready events whose requests resolve in reverse order.

2. **The asynchronous event handler has no rejection boundary.**

   `packages/ui/src/lib/sse-manager.ts:231-233,266`
   `packages/ui/src/stores/sessions.ts:106`

   `onWorktreeReady` is typed as returning `void`, but it is assigned an async function and its promise is discarded. The current request helpers absorb common HTTP failures, but an import/chunk failure or a future thrown error becomes an unhandled rejection.

   Type the callback as `void | Promise<void>` and explicitly capture/log failures at the dispatch boundary, or make the registered callback wrap the promise with a rejection handler.

## Test coverage

The new test confirms nominal ordering, but it does not cover the event dispatch or the concurrency race above. The project Bun suite itself is green, so this is a coverage gap rather than a harness failure.

## Scope notes

- This only refreshes for OpenCode-managed creation. A direct external `git worktree add` does not emit `worktree.ready`.
- Older selectable OpenCode binaries that do not emit this event retain the stale-inventory behavior.
- Avoiding session auto-switch remains correct because the event contains no originating session ID.

Validation reviewed: UI typecheck passes, the full UI suite reports 126 passing tests, and the OpenCode global event envelope is correctly flattened by the existing server bridge.

Verdict: **changes requested**. The per-instance concurrency race must be fixed before merge.

pascalandr:
Addressed the gatekeeper findings in `611a3986`.

- Serialized inventory reloads per instance and made each reload wait for any initial inventory hydration, preventing older responses from overwriting newer state.
- Serialized the complete `worktrees -> OpenCode workspaces` refresh for each instance, while keeping different instances independent.
- Removed the unsafe workspace-sync lock reset and made cleanup conditional on promise identity, so an older `finally` cannot delete a newer synchronization.
- Updated the SSE callback contract to `void | Promise<void>` and added explicit handling for both synchronous throws and asynchronous rejections.
- Added regression tests for overlapping ready events and queue recovery after a rejected refresh.

Validation after the changes:

- `bun test --conditions browser packages/ui/src`: 128 passing
- `npm run typecheck --workspace @codenomad/ui`: passing
- `git diff --check`: passing

pascalandr:
Gatekeeper re-review for the latest PR state (`611a3986`).

## Blocking finding

1. **Initial hydration can still overwrite a newer `reloadWorktrees()` result.**

   `packages/ui/src/stores/worktrees.ts:35-82,85-121`

   `reloadWorktrees()` only waits for the `worktreeLoads` promise that exists when line 90 is evaluated. `ensureWorktreesLoaded()` can start immediately after that lookup, creating a request that runs concurrently with the reload.

   Concrete scenario:

   - A ready event starts `reloadWorktrees()` while no initial load is registered yet.
   - UI hydration then starts `ensureWorktreesLoaded()` and its request observes only worktree A.
   - Worktree B appears before the reload request enumerates Git, so the reload observes `[A, B]` and commits it.
   - The older initial-load response finishes last and overwrites the inventory with `[A]`.

   The ready-event queue and OpenCode workspace synchronization can therefore complete successfully while the final UI inventory remains stale. Initial loads and reloads need one shared per-instance ordering/version mechanism so they cannot commit responses in reverse order.

## Follow-up coverage

The added tests correctly cover ready-event serialization and recovery after rejection, but they inject refresh callbacks. They do not exercise the production interaction between `ensureWorktreesLoaded()` and `reloadWorktrees()`, which is why this remaining race is not detected. Add a regression test that holds the initial request and reload request open, resolves the reload first, then verifies that the older initial result cannot replace it.

## Verified fixes

- Multiple ready events for one instance are serialized correctly.
- Different instances retain independent queues.
- `workspaceSyncs` cleanup is protected by promise identity.
- Queue Maps recover after rejection.
- `SSEManager` captures synchronous throws and asynchronous callback rejections.

## Validation

- `git diff --check upstream/dev...HEAD`: passing
- `bun test --conditions browser packages/ui/src/stores/worktree-ready.test.ts packages/ui/src/stores/opencode-workspaces.test.ts`: 7 passing
- UI typecheck: passing
- Server typecheck: passing

Verdict: **CHANGES REQUESTED**. The initial-load versus reload race must be fixed before merge.

pascalandr:
Addressed the remaining hydration/reload race in `dd9ac800`.

- Replaced the separate initial-load and reload locks with one per-instance `worktreeRequests` queue.
- Initial hydration and explicit reloads can no longer write concurrently or commit responses in reverse order.
- `ensure -> reload` now guarantees a trailing fresh enumeration.
- `reload -> ensure` waits for and reuses the reload result, with an initial-load fallback if the reload failed to populate state.
- Queue cleanup remains guarded by promise identity.
- Added a production-path regression test that holds initial hydration open, queues a reload, and verifies that the newer inventory is the final state.

Validation:

- `bun test --conditions browser packages/ui/src`: 129 passing
- `npm run typecheck --workspace @codenomad/ui`: passing
- `git diff --check`: passing

The follow-up gatekeeper pass found no remaining blocker and returned **READY**.

pascalandr:
Gatekeeper review for the latest PR state (`dd9ac800`).

## Findings

No blocking bug found. All previously reported blockers are resolved.

**Low, test coverage only:** `packages/ui/src/stores/worktree-ready.test.ts:97-144` exercises the real `ensure -> reload` path, but does not explicitly cover `reload -> ensure`, dispatch through `SSEManager`, or the production `workspaceSyncs` lock. Static review found those paths correct; this is a follow-up coverage opportunity, not a merge blocker.

## Verified behavior

- `worktreeRequests` serializes all inventory writes for one instance while keeping different instances independent.
- Both `ensure -> reload` and `reload -> ensure` preserve the newest inventory.
- `worktreeReadyRefreshes` preserves the complete `worktrees -> workspaces` ordering and recovers after rejection.
- Promise-identity cleanup prevents older tasks from deleting newer queue entries.
- `workspaceSyncs` waits for an existing synchronization before starting the requested fresh synchronization.
- `SSEManager` captures both synchronous handler throws and asynchronous rejections.
- The existing server bridge correctly flattens the OpenCode global event envelope and preserves its worktree directory.
- No new runtime import cycle was introduced.

## Residual scope

- The refresh remains limited to OpenCode versions and creation paths that emit `worktree.ready`.
- `reloadWorktrees` remains best effort and preserves its historical error handling.
- Missing experimental workspace APIs degrade safely without throwing.

## Validation

- `git diff --check upstream/dev...HEAD`: passing
- `bun test --conditions browser packages/ui/src`: 129 passing
- Targeted event/transport tests: passing
- UI typecheck: passing
- Server typecheck: passing
- Worktree clean

Verdict: **READY**. No remaining merge blocker.

pascalandr:
@shantur LGTM

PR #585, opened 2026-07-13
feat(settings): add OpenCode update action

pascalandr:
## Summary

- add an OpenCode update card to the OpenCode settings section
- show the installed version and a versioned update button only when npm advertises a newer release
- run the upgrade through the generated OpenCode SDK against a ready instance using the configured binary
- keep running instances uninterrupted; newly started instances use the installed version
- localize loading, unavailable-instance, success, and failure states across all supported locales

## Server behavior

- resolve the selected binary with the existing binary probe
- check opencode-ai/latest with a 10-second timeout and a five-minute cache
- require a ready workspace using that exact binary before enabling the upgrade
- invoke client.global.upgrade({ target }) rather than assembling an OpenCode API request manually
- return stable error codes without exposing server details

## Validation

- pm run typecheck in packages/server
- pm run typecheck in packages/ui
- ode --import tsx --test src/opencode-update/service.test.ts
- pm run build in packages/ui
- pm run build in packages/server
- git diff --check

image
image


pascalandr (commented):
## Gatekeeper review: CHANGES REQUIRED

### Findings

1. **High: arbitrary binary probing through GET is unsafe on Windows.** `packages/server/src/server/routes/opencode-update.ts:11-12,27-30` accepts an arbitrary `binary` query parameter and `packages/server/src/opencode-update/service.ts:39-42,86-90` executes it with `--version`. `.cmd` paths are interpolated into a `cmd.exe /c` command line by `packages/server/src/workspaces/spawn.ts:82-91`, so crafted quotes/metacharacters can become command injection. Because the session cookie is `SameSite=Lax`, a cross-site top-level GET can carry authentication. Resolve only persisted/validated binaries server-side and do not expose command execution through a parameterized GET.

2. **High: the normal Windows npm installation cannot show status or update.** The service probes the raw configured value, normally `opencode`, while workspace launch first resolves it with `where`. Reproduced locally: `opencode --version` returns `1.17.18`, but `probeBinaryVersion("opencode")` returns `spawnSync opencode ENOENT`; `where` resolves `opencode.cmd`. Status and upgrade must use the same canonical resolution as workspace startup.

3. **High: upgrade inherits an unsuitable 10-second timeout.** `createInstanceClient` applies `AbortSignal.timeout(10_000)` in `packages/server/src/workspaces/instance-client.ts:42-46`, and the upgrade uses that client at `packages/server/src/opencode-update/service.ts:145-149`. npm/Homebrew/Chocolatey/WSL upgrades routinely exceed ten seconds, potentially reporting failure during a still-running or partial installation. Use an upgrade-specific longer timeout and deterministic in-flight handling.

4. **Medium: success is not verified against the selected executable.** The SDK result is trusted and the UI immediately declares the target version installed. Re-probe the canonical selected binary after upgrade and only report success when its actual version matches the target; otherwise a custom/portable binary may leave a different global installation updated.

5. **Medium: registry failure hides the locally installed version.** Local probing and the npm request are combined into one failing status operation. Return the installed version independently and model update availability as unavailable when the registry cannot be reached.

6. **Medium: concurrent clients can launch overlapping package-manager upgrades.** Add a per-canonical-binary in-flight guard that joins or rejects duplicate requests.

7. **Medium: WSL binary matching lowercases the Linux path suffix.** `packages/server/src/workspaces/manager.ts:81-86` lowercases the entire UNC path on Windows, although the Linux path portion is case-sensitive. Preserve Linux path casing when matching WSL binaries.

8. **Low: tests cover only mocked service behavior.** Missing coverage includes route validation, Windows command resolution, timeout behavior, post-upgrade verification, concurrency, WSL identity, and UI states. The focused service test is also not wired into the current package/CI test scripts.

### Residual context

All current CI builds are green or completing successfully, but they compile the feature without exercising the runtime paths above.

**Verdict: CHANGES REQUIRED**

pascalandr (commented):
## Gatekeeper follow-up: PASS

Commit `30c81867` resolves the blocking findings from the previous review:

- removed request-controlled binary paths from both update endpoints;
- resolved the persisted binary with the same canonical lookup used by workspace startup;
- verified the Windows npm wrapper resolves from `opencode` to `opencode.cmd` and reports its version;
- extended only the upgrade SDK call to a ten-minute timeout;
- re-probed the configured executable after upgrade before reporting success;
- preserved the installed version when the npm registry check fails;
- coalesced concurrent upgrades per canonical binary;
- preserved Linux path case when comparing WSL UNC binaries;
- ignored stale upgrade completions after the configured binary changes;
- added route, service, concurrency, verification, registry-failure, Windows-path, and WSL-path tests.

Validation completed locally:

- 10 focused tests passed;
- server typecheck passed;
- UI typecheck passed;
- UI build passed;
- full server build passed;
- real Windows resolution/probe passed for `opencode.cmd`.

Residual non-blocking gaps: there is no dedicated component test for the stale-completion transition, and the focused tests are not yet wired into a dedicated CI job. The full server suite produced 153 passes and one unrelated existing failure caused by the local Git default branch name.

**Verdict: PASS**

shantur:
@pascalandr - Needs to resolve conflicts

pascalandr:
@Shantur LGTM

PR #586, opened 2026-07-13
perf(ui): virtualize nested session list

pascalandr:
## Summary
- flatten expanded session trees into a stable pre-order row projection
- render only the viewport and buffer with the existing Solid virtua integration
- preserve nested connectors, dynamic heights, search expansion, focus, active-session scrolling, and loading sentinel behavior

## Why
Large or deeply expanded session trees currently mount every visible row. Status updates then affect a large DOM tree even when most sessions are far outside the sidebar viewport. This PR bounds mounted rows while leaving selection, deletion, filtering, and store semantics unchanged.

## Validation
- UI TypeScript typecheck
- 47 focused session-tree, pagination, and virtual-scroll tests
- 1,000-session projection regression test
- production Vite build
- git diff whitespace check
- final focused regression review
- manual test


pascalandr (commented):
Gatekeeper verdict: APPROVE.

No blocking code findings. The virtualized projection preserves nested ordering, expansion and search semantics, stable Solid/virtua identity, dynamic measurement, focused-row mounting, active-index scrolling, and correct sentinel rooting. Focused projection tests and UI typechecking pass.

The required macOS check should be rerun before merging: its failure was an external connection reset while downloading the Electron darwin-arm64 archive, after the preceding build validations and macOS x64 packaging had succeeded.

Residual risks are limited to missing browser-level coverage for active scrolling, focus retention, expansion under virtualization, responsive/RTL rendering, and same-length reorder measurement. Touched oversized files remain session-list.tsx at about 917 lines and session-layout.css at about 572 lines.

pascalandr:
@Shantur LGTM

PR #588, opened 2026-07-14
feat(ui): support bracket math delimiters

jollyxenon:
## Why

LaTeX content commonly uses `\(...\)` and `\[...\]`, but CodeNomad only recognizes `$...$` and `$$...$$` through `marked-katex-extension`.

## What

- Add parser-native Marked rules for inline `\(...\)` and display `\[...\]` formulas.
- Preserve existing dollar math, code spans and fences, escaped delimiters, and unmatched text.
- Add focused regression tests and document the delimiter design.

### DIFF
--- a/docs/superpowers/specs/2026-07-14-bracket-math-delimiters-design.md
+++ b/docs/superpowers/specs/2026-07-14-bracket-math-delimiters-design.md
@@ -1,0 +1,38 @@
+# Bracket Math Delimiters Design
+
+## Goal
+
+Support inline `\(...\)` and display `\[...\]` math in CodeNomad Markdown while preserving existing `$...$` and `$$...$$` behavior. No unrelated Markdown changes.
+
+## Approach
+
+Keep `marked-katex-extension` as owner of dollar-delimited math. Add two small parser-native Marked extensions in `packages/ui/src/lib/markdown.ts`:
+
+- Inline rule: recognize `\(...\)` and render with KaTeX `displayMode: false`.
+- Block rule: recognize `\[...\]` and render with KaTeX `displayMode: true`.
+
+Both rules will use `katex.renderToString()` with the same error policy as the existing integration: `throwOnError: false` and `strict: "ignore"`. Parser-native rules avoid source rewriting, ignore fenced/code content through Marked's normal lexer flow, and preserve unmatched delimiters as plain Markdown text.
+
+## Integration
+
+Register the two rules beside the existing `markedKatex(...)` registration in `setupRenderer()`. Existing renderer setup, syntax highlighting, HTML handling, styles, and component APIs remain unchanged. No new runtime dependency is needed because KaTeX is already a direct UI dependency.
+
+## Edge Cases
+
+- `\\(` and `\\[` remain escaped literal text rather than opening math.
+- Inline math must close with `\)` before a line break.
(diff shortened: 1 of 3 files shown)


pascalandr (changes requested):
## Gatekeeper review of a0fb8554

## Blocking finding

1. **P1: display bracket math is not recognized when it follows prose without a blank line.**

The block extension at packages/ui/src/lib/markdown.ts:431-453 only tokenizes when the current block source begins with \\[. It has no block start() hint to interrupt an in-progress paragraph.

Reproduction: a line containing before, followed immediately by a line containing \\[x^2\\], followed by a line containing after.

Current output is effectively <p>before<br>[x^2]<br>after</p>. The same formula renders only when blank lines force a new block first.

This misses a normal LaTeX/LLM formatting pattern and makes the advertised \\[...\\] support depend on undocumented surrounding whitespace.

Add paragraph-interruption handling, for example a block start() hint that locates a valid unescaped line-start opener, or another parser-native approach that preserves code/fence precedence. Add a regression with prose immediately before and after the display block, including a multiline formula.

## Important observations

- packages/ui/src/lib/markdown.ts:21-24 still says all math rendering and delimiter rules are delegated to marked-katex-extension specifically to avoid hand-written tokenizer code. That is now false and should be updated to describe the split ownership.
- The new test assertions all pass, but npx tsx --test src/lib/markdown.test.ts did not terminate locally after reporting 8/8 passed; it remained alive until timeout. Please verify the documented test command exits cleanly in the supported Node environment.

## What looks correct

- Inline \\(...\\) rendering works and stops at line boundaries.
- Standalone and blank-line-separated display formulas render with displayMode.
- Code spans, fences, escaped delimiters, empty/unmatched delimiters, and existing dollar math remain protected by the focused cases.
- KaTeX uses the existing throwOnError: false and strict: ignore policy, with trust left disabled.

## Validation

- Focused assertions: 8/8 passed before the runner hang.
- npm run typecheck in packages/ui: passed.
- npm run build in packages/ui: passed.
- Complete required Electron/Tauri build matrix: green.

## Verdict

**Changes requested.** The inline path is ready, but the display tokenizer must handle the common no-blank-line paragraph boundary before merge.

pascalandr (approved):
## Gatekeeper re-review of 3aa36cd3

**PASS.** The previous blocking case is fixed: single-line and multiline \[...\] blocks now interrupt adjacent prose without requiring blank lines, while inline code, escaped/empty/unmatched delimiters, dollar math, blockquotes, and lists retain the expected behavior.

Validation:
- Focused Markdown assertions: 11/11 passed.
- Additional blockquote/list/code-span cases: passed.
- UI typecheck: passed.
- UI production build: passed.
- git diff --check: passed.
- Full required Electron/Tauri CI matrix: green.

Non-blocking follow-ups:
- The focused test process remains alive after all assertions under both project-supported Node 20 and local Node 25. The assertions are valid, but the test lifecycle should be fixed before this command is wired into CI.
- The comment above the inline extension's start() says it does not scan beyond the next line, while src.search() can locate an opener on a later line; the tokenizer itself still enforces the intended same-line closing rule.
- packages/ui/src/lib/markdown.ts is approximately 634 lines, above the repository's 500-line refactoring warning threshold.

**Verdict: approved.**

pascalandr:
Thanks @jollyxenon

PR #612, opened 2026-07-18
fix(ui): defer session virtualization for closed left panel

pascalandr:
## Summary
- do not mount SessionList while the temporary left drawer is floating and closed
- keep the session-list error state mutually exclusive with virtualized rows
- register focused visibility-policy tests in the PR workflow

## Root cause
SUID constructs temporary Drawer children while open is false. This occurs both after restarting with a persisted closed panel and when narrowing the window into mobile mode, which forces the left panel to become unpinned and closed. SessionSidebar then mounted the virtua Virtualizer in a detached staging document. virtua resolves ResizeObserver through ownerDocument.defaultView, which is null for that document.

The failure is timing-dependent: if session hydration publishes rows while that closed mobile Drawer is detached, the synchronous render exception escapes through setSessionPage and is caught by fetchSessions as if the successful API request had failed. Opening the panel later therefore reveals an empty list or the misleading Unable to load sessions error. If hydration finishes under a different drawer lifecycle, the bug does not appear.

Because the error UI and virtualized rows were both mounted, Retry cleared the error and immediately hit the same poisoned lifecycle again.

## Behavior
Session fetching and startup restore continue while the panel is closed. The virtualized DOM is created only after the panel is open or pinned. Genuine list errors dispose the rows; Retry can then mount a clean virtualizer after succeeding.

## Reproduction
1. Narrow the window until CodeNomad enters mobile mode and the left panel can no longer remain pinned.
2. Leave the sessions panel closed while sessions hydrate, or restart in that state.
3. Open the left panel.
4. Before this fix, the list may be empty or show Unable to load sessions with a ResizeObserver null error.

## Validation
- 19 focused session visibility, tree, and pagination tests
- UI TypeScript typecheck
- production Vite build
- full Windows Tauri release build
- NSIS installer bundle
- regression test included in PR CI

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -104,6 +104,7 @@ jobs:
       - name: Test changed runnable UI behavior
         run: >-
           node --import tsx --test
+          packages/ui/src/components/session-list-visibility.test.ts
           packages/ui/src/lib/hooks/use-app-session-capture.test.ts
           packages/ui/src/lib/trailing-resync.test.ts
           packages/ui/src/stores/abort-created-workspace-cleanup.test.ts
--- a/packages/ui/src/components/instance/shell/SessionSidebar.tsx
+++ b/packages/ui/src/components/instance/shell/SessionSidebar.tsx
@@ -18,6 +18,7 @@ import AgentSelector from "../../agent-selector"
 import ModelSelector from "../../model-selector"
 import ThinkingSelector from "../../thinking-selector"
 import { getLogger } from "../../../lib/logger"
+import { shouldMountSessionList } from "../../session-list-visibility"
 
 const log = getLogger("session")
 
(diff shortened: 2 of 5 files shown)


pascalandr (commented):
Gatekeeper verdict: REQUEST_CHANGES.

The closed-state guard prevents session hydration from mounting virtua while the temporary drawer is closed, but it does not by itself prove safety for the closed -> hydrated -> open transition. SUID can pre-resolve drawer children into its staged tree; SessionSidebar currently mounts the populated SessionList as soon as drawerState becomes floating-open, before portal attachment is guaranteed. That can reach the same ownerDocument.defaultView.ResizeObserver failure.

Please gate virtualization on actual DOM attachment and add coverage for the detached-versus-attached viewport policy. The current helper-only test verifies logical drawer state and error exclusivity but not portal attachment.

The error/row exclusivity, Retry remount policy, pinned behavior, mobile close transition, Solid reactivity, workflow registration, and all current CI checks otherwise pass. The empty-message metadata-gate hypothesis remains separate from this session-list crash.

pascalandr (commented):
Gatekeeper re-review: APPROVE.

The follow-up commit 6c0e7180 closes the staged-portal race identified in the previous review. Virtualized rows now wait for the viewport to be both connected and owned by a document with a live defaultView before virtua initializes ResizeObserver. RAF polling is limited to the unattached transition and cancelled on disposal.

The logical closed-drawer guard, actual DOM attachment gate, error/row exclusivity, Retry clean remount, pinned/open behavior, and mobile transition now compose correctly. The focused suite covers detached, disconnected, and attached viewport policies; 20 focused tests and UI typecheck pass. No blocking findings remain.

pascalandr:
@Shantur LGTM

PR #620, opened 2026-07-22
fix(auth): submit the displayed default username

### LINKED ISSUE #616: [Bug]: Remote login displays codenomad as a placeholder but submits an empty username
### App Variant

Tauri

Windows 11 Pro 10.0.26200 (build 26200)

### CodeNomad Version

0.18.0

### Issue Summary

The remote login page visually displays the default username (`codenomad`), but it is only placeholder text. The input's actual value is empty, so entering a password and pressing Continue does not submit the displayed username. This is especially misleading on a phone, where the field appears to have already been populated.

### Steps to Reproduce

1. In the Tauri app, enable **Allow connections from other IPs** and configure a server password.
2. From another device, open one of the HTTPS remote-access addresses (reproduced over Tailscale).
3. Observe that the username field appears to show `codenomad`.
4. Enter the configured password without manually retyping the apparently populated username.
5. Press **Continue**.
6. Authentication does not proceed because the username input's value is empty.
7. Manually type `codenomad` into the username field and submit again; login succeeds.

### Expected Behavior

The displayed default username should be the input's actual value, so the form submits `codenomad` unless the user intentionally changes it.

### Root Cause

In `packages/server/src/server/routes/auth-pages/login.html`, the username is rendered only as a placeholder:

```html
<input id="username" autocomplete="username" placeholder="{{DEFAULT_USERNAME}}" value="" />
```

The submit handler reads the value rather than the placeholder:

```js
const username = $("username").value.trim()
```

Consequently, the username is empty even though the page appears to display `codenomad`.

### Proposed Solution

Populate the actual value and disable mobile text transformations on credential fields:

```html
<input
  id="username"
  autocomplete="username"
  autocapitalize="none"
  autocorrect="off"
  spellcheck="false"
  placeholder="{{DEFAULT_USERNAME}}"
  value="{{DEFAULT_USERNAME}}"
/>

<input
  id="password"
  type="password"
  autocomplete="current-password"
  autocapitalize="none"
  autocorrect="off"
  spellcheck="false"
  value=""
/>
```

I applied this change locally to both the source template and installed server resource. Login was then verified successfully from a phone through the Tailscale address. No password hashing, cookie, or network behavior needs to change.

### Additional Context

This is distinct from #128, which concerned LAN authentication/session behavior over HTTP. In this case HTTPS connectivity and the password were valid; explicitly entering the username made the same login succeed.

pascalandr:
## Summary
- Populate the remote login username input with the configured default value instead of displaying it only as a placeholder.
- Disable mobile capitalization, autocorrection, and spellchecking for username and password inputs.
- Add a route-level regression test for the rendered username value and mobile credential attributes.

## Validation
- node --import tsx --test packages/server/src/server/routes/auth.test.ts
- node packages/server/scripts/copy-auth-pages.mjs

Closes #616

### DIFF
--- a/packages/server/src/server/routes/auth-pages/login.html
+++ b/packages/server/src/server/routes/auth-pages/login.html
@@ -95,7 +95,7 @@
 
       async function submit() {
         hideError()
-        const username = $("username").value.trim()
+        const username = $("username").value
         const password = $("password").value
         if (!username || !password) {
           showError("Username and password are required.")
--- a/packages/server/src/server/routes/auth-pages/login.html
+++ b/packages/server/src/server/routes/auth-pages/login.html
@@ -72,10 +72,10 @@
       <p>This CodeNomad server is protected. Enter your credentials to continue.</p>
 
       <label for="username">Username</label>
-      <input id="username" autocomplete="username" placeholder="{{DEFAULT_USERNAME}}" value="" />
+      <input id="username" autocomplete="username" autocapitalize="none" autocorrect="off" spellcheck="false" placeholder="{{DEFAULT_USERNAME}}" value="{{DEFAULT_USERNAME}}" />
 
       <label for="password">Password</label>
-      <input id="password" type="password" autocomplete="current-password" value="" />
+      <input id="password" type="password" autocomplete="current-password" autocapitalize="none" autocorrect="off" spellcheck="false" value="" />
 
       <button id="submit" type="button">Continue</button>
       <div id="error" class="error" style="display: none"></div>
--- a/packages/server/src/server/routes/auth.test.ts
+++ b/packages/server/src/server/routes/auth.test.ts
@@ -4,18 +4,20 @@ import Fastify from "fastify"
 import type { AuthManager } from "../../auth/manager"
 import { registerAuthRoutes } from "./auth"
 
-it("renders the default username as the login field value", async () => {
+it("renders the escaped default username literally as the login field value", async () => {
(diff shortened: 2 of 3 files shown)


pascalandr (commented):
Gatekeeper review - round 2

Round 1 finding is resolved locally.

Finding (medium): login.html trims the populated username before submission, while configured usernames are preserved and compared verbatim. A username with leading or trailing whitespace therefore renders but cannot authenticate.

Required fix: submit the exact input value and add whitespace regression coverage.

Status: changes requested; another gatekeeper pass will follow.

pascalandr (commented):
Gatekeeper review - round 1

Finding (medium): auth.ts used the escaped username as a String.replace replacement string, so configured usernames containing replacement tokens such as $& could corrupt the rendered login value or markup.

Required fix: use a replacement callback and add adversarial coverage for literal $ handling plus HTML attribute escaping.

Status: fix implemented locally; follow-up review required.

pascalandr (commented):
Gatekeeper review - round 3

Zero actionable findings.

Verified:
- configured usernames containing replacement tokens remain literal after HTML escaping
- leading and trailing username whitespace is rendered and submitted verbatim
- regression coverage fails against both previous behaviors
- mobile credential attributes remain correct

Residual risk: no real-device Mobile Safari/Chrome run. Targeted test and server typecheck pass.

pascalandr:
@shantur LGTM

PR #621, opened 2026-07-22
fix(permissions): ignore stale permission updates

### LINKED ISSUE #566: Permission request keeps reappearing after being granted, causing app freeze
## Bug Description

Even after granting permission for a folder access request, the "Permission required" message keeps reappearing. The previously granted consent is not being remembered/applied.

## Steps to Reproduce

1. Trigger a permission request for folder access in CodeNomad
2. Grant the permission (approve the request)
3. Observe that the "Permission required" message appears again despite the prior approval

## Expected Behavior

Once a permission for folder access has been granted, it should be remembered and the "Permission required" message should not reappear for the same scope/folder.

## Actual Behavior

- The "Permission required" message reappears even after the permission has already been granted.
- CodeNomad frequently freezes completely after this happens.
- The "Permission required" dialog cannot be dismissed.
- A full restart of CodeNomad is usually required to recover.

## Impact

- Blocks normal workflow
- Forces frequent restarts
- Permission state is not persisted correctly between requests

## Environment

- OS: Windows
- (Please add your CodeNomad version and OS details)

## Additional Context

Screenshots were provided by the reporter showing the recurring "Permission required" message. Unfortunately they could not be attached to this issue automatically.

/cc @NeuralNomadsAI

pascalandr:
## Summary
- Preserve the original V2 source when permission.updated changes an already-pending request, so approval continues through the V2 reply API instead of the legacy endpoint.
- Keep standalone permission.updated events as valid legacy requests for older supported OpenCode binaries.
- Continue suppressing delayed post-reply updates through the existing replied-request tombstone, preventing an approved request from reappearing.

## Reproduced failure
The regression test fails on dev before the fix: permission.v2.asked followed by permission.updated changes the request source to legacy, so approval calls the legacy endpoint and never reaches the V2 reply API.

## Compatibility coverage
- V2 request plus permission.updated replies through V2
- delayed permission.updated after a successful local reply does not re-enter the queue
- standalone legacy permission.updated remains visible and replies through legacy

## Scope
This PR intentionally does not add speculative timeout, reconnect, or modal-rendering changes. It only fixes the reproduced source-routing bug and preserves verified legacy behavior.

## Validation
- npm run typecheck
- 144 existing runnable UI tests passed
- browser integration tests passed, including all permission lifecycle regressions

Closes #566

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -127,6 +127,7 @@ jobs:
         run: >-
           node --conditions=browser --import tsx --test --test-force-exit
           packages/ui/src/stores/instances-restore-ownership.test.ts
+          packages/ui/src/stores/permission-lifecycle.test.ts
 
       - name: Test server
         run: node --import tsx --test "packages/server/src/**/*.test.ts"
--- a/packages/ui/src/stores/instances.ts
+++ b/packages/ui/src/stores/instances.ts
@@ -1391,12 +1391,12 @@ function recomputeActiveInterruption(instanceId: string): void {
   setActiveInterruptionForInstance(instanceId, computeActiveInterruption(instanceId))
 }
 
-function addPermissionToQueue(instanceId: string, permission: PermissionRequest, source: PermissionSource = "v2"): PermissionRequest | undefined {
+function addPermissionToQueue(instanceId: string, permission: PermissionRequest, source?: PermissionSource): PermissionRequest | undefined {
   let inserted = false
   let updated = false
   let previousPermission: PermissionRequest | undefined
   let queuedPermission = permission
-  permissionRegistry.setSource(instanceId, permission.id, source)
+  if (source) permissionRegistry.setSource(instanceId, permission.id, source)
 
   setPermissionQueues((prev) => {
     const next = new Map(prev)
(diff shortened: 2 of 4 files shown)


pascalandr (commented):
Gatekeeper review - round 1

Finding (P1): the new orphan permission.updated guard drops valid initial legacy permission requests. Supported older OpenCode binaries can emit permission.updated as the sole initial request event, so their sessions would remain blocked with no permission visible in CodeNomad.

Deterministic result: a standalone legacy permission.updated leaves both UI permission queues empty.

Required fix: accept standalone permission.updated as legacy, preserve the existing source when it updates an already-pending V2 request, and rely on replied-request tombstones to reject delayed post-reply updates.

Status: changes requested; another published gatekeeper round will follow.

pascalandr (commented):
Gatekeeper review - round 2

Zero actionable findings.

Verified:
- standalone permission.updated remains a valid legacy request
- permission.asked remains legacy
- permission.v2.asked followed by permission.updated preserves the V2 reply route
- delayed permission.updated after a successful reply does not re-enter either permission queue
- malformed permission IDs are ignored

Relevant UI tests, browser integration tests, typecheck, and UI build pass. Residual limitation: no direct reproduction on the reporter's Windows device.

pascalandr:
@shantur LGTM

PR #622, opened 2026-07-22
fix(ui): avoid mobile selection action overlap

### LINKED ISSUE #598: [UI] On mobile, Can't choose add quote and add comment
After I select text on mobile, mobile popup appear above codenomad selection popup options, so I can't select any of its options.

pascalandr:
## Summary
- place CodeNomad selection actions below the final selected line on touch-only devices
- keep desktop and hybrid fine-pointer positioning unchanged
- fall back to the message stream's top edge when bottom placement would be clipped

## Validation
- 146 UI Node tests
- npm run typecheck
- git diff --check

Closes #598

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -106,6 +106,7 @@ jobs:
           node --import tsx --test
           packages/ui/src/components/session-list-visibility.test.ts
           packages/ui/src/lib/hooks/use-app-session-capture.test.ts
+          packages/ui/src/lib/message-selection-position.test.ts
           packages/ui/src/lib/trailing-resync.test.ts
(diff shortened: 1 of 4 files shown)


pascalandr (commented):
Gatekeeper review, round 1: 1 finding.

High: touch placement initially did not account for the clipped message stream's bottom edge, so a selection near the bottom could move the action popover off-screen. Fixed in this PR by falling back to the stream's top edge when the popover cannot fit below the selection, with a regression test for that geometry.

pascalandr (commented):
Gatekeeper review, round 2: zero findings.

Verified the demonstrated Android native-toolbar collision, unchanged desktop coordinates, touch bottom-edge visibility, focused regression coverage, and PR CI registration.

pascalandr:
@shantur LGTM

PR #626, opened 2026-07-26
fix(ui): bound renderer memory and keep large sessions responsive

### LINKED ISSUE #553: [Bug]: WebKit renderer memory grows to 2.5 GB over 24h+ runtime — tool outputs fully rendered with no eviction
## Description

The WebKit WebContent renderer process (WKWebView) grows to 2.5 GB physical footprint after ~1.3 days of continuous runtime. This is a memory accumulation bug, not a crash-triggering leak — it degrades system performance over time as the process is paged out.

**Physical footprint peak observed: 3.1 GB**, stable at 2.5 GB before investigation.

```
Physical footprint:         2.5G
Physical footprint (peak):  3.1G
```

Live process analysis (vmmap PID 55616, macOS 26.5.1, Tauri variant):

| Region | Dirty | Swapped | **Physical Cost** | **% of 2.5G** |
|--------|-------|---------|------------------|---------------|
| WebKit Malloc (DOM, styles, layers) | 183 MB | 1,500 MB | **~1,683 MB** | **67%** |
| Graphics buffers (offscreen renders) | 79 MB | 609 MB | **~688 MB** | **27%** |
| V8/JS heap (DefaultMallocZone + MALLOC_SMALL) | 9 MB | 21 MB | **~30 MB** | **1%** |
| JS VM Gigacage + JIT | 4 MB | 16 MB | **~20 MB** | **1%** |
| Everything else | 4 MB | 54 MB | **~58 MB** | **4%** |

**2.7 million allocations** in the WebKit Malloc zone, totaling **5.5 GB allocated**.

## Environment

- **CodeNomad variant**: Tauri
- **OS**: macOS 26.5.1 (25F80)
- **Runtime**: ~1.3 days (launched 2026-06-13 01:39)
- **Observability**: Renderer PID 55616, no remote debugging port open (no `--inspect` flag)

## Root Cause

Three independent issues interact to produce the 2.5 GB:

### 1. Tool outputs rendered as full DOM with no size limits (P0)

Every tool call's output (bash, write, edit, read, etc.) is rendered as complete DOM nodes when expanded. Default state is **expanded** for all non-read tools.

`stores/preferences.tsx:150` — default preference:
```
toolOutputExpansion: "expanded"
```

`components/tool-call.tsx:638-641` — output section always expanded:
```
const outputSectionExpanded = () => {
    const override = outputSectionOverride()
    if (override !== null) return override
    return true  // <-- always expanded
}
```

`components/tool-call.tsx:609-622` — `defaultExpandedForTool` returns `true` for all non-read tools.

`components/tool-call/renderers/bash.tsx:60` — no size cap:
```
return [command, outputResult?.text].filter(Boolean).join("\n")
```

`components/tool-call/renderers/bash.tsx:91-98` — rendered as full innerHTML:
```
<Show when={finalAnsiHtml()} fallback={...}>
    {(html) => <div innerHTML={html()} />}
</Show>
```

Message-level virtual scrolling (`virtua`, 800px overscan) exists, but **within each visible message, every tool call's full output is a complete DOM subtree**. WebKit's DOM tree + render layers + compositing multiplies text size by ~3-5x. This is why 67% of the physical footprint lives in WebKit Malloc — it's the DOM tree and render layers, not JS heap.

### 2. Message store never evicts data (P0)

The message store (`stores/message-v2/instance-store.ts`) accumulates all messages, parts, part data strings, permissions, questions, todos, and scroll state with zero automatic eviction.

`instance-store.ts:578-579` — unbounded pending parts buffer:
```
function bufferPendingPart(entry: PendingPartEntry) {
    setState("pendingParts", entry.messageId, (list = []) => [...list, entry])
}
```

`instance-store.ts:714` — unbounded string concatenation during streaming:
```
part[input.field] = `${currentValue ?? ""}${input.delta}`
```

`instance-store.ts:1162` — `clearSession()` exists but is only called on explicit user deletion.
`instance-store.ts:1262` — `clearInstance()` only called on instance removal.
**There is NO automatic eviction — no LRU, no TTL, no memory-pressure trigger.**

### 3. Session state never cleaned on instance close (P1)

When an instance is stopped, `removeInstance()` (`stores/instances.ts:626-668`) cleans **7** stores but misses **15+** per-instance signals in `session-state.ts`:

```
function removeInstance(id: string) {
    setInstances(...);          // ✅
    removeLogContainer(id);     // ✅
    clearCommands(id);          // ✅
    clearPermissionQueue(id);   // ✅
    clearRepliedPermissions(id); // ✅
    clearQuestionQueue(id);     // ✅
    clearInstanceMetadata(id);  // ✅
    clearCacheForInstance(id);  // ✅
    messageStoreBus.unregisterInstance(id); // ✅
    clearInstanceDraftPrompts(id);         // ✅
    // ❌ sessions, activeSessionId, activeParentSessionId,
    //    agents, providers, messagesLoaded, messageLoadErrors,
    //    sessionInfoByInstance, threadTotalsByInstance,
    //    expandedSessionParents, sessionPagination, sessionSearch,
    //    instanceIndicatorCounts, loading, sessionThreadCache
}
```

The `sessions` signal (`stores/session-state.ts:37`) is `Map<string, Map<string, Session>>` keyed by instanceId — when an instance is closed, its `Map` entry is never deleted. All sessions for that instance persist forever.

## Steps to Reproduce

1. Launch CodeNomad (Tauri or Electron variant)
2. Create a workspace/session and run several agent conversations with tool calls (bash, file writes, diffs — the larger the output, the faster the growth)
3. Keep the app running for 24+ hours
4. Observe renderer memory via Activity Monitor or `vmmap`:
   ```bash
   vmmap <WebContent PID> -summary
   ```

The process will show 2-3 GB physical footprint dominated by WebKit Malloc and graphics memory.

## Expected Behavior

- Renderer memory should stay **below ~500 MB** under sustained use
- Sessions not actively viewed should not retain full DOM
- Closed instances should not retain any session data
- Old sessions should be evicted from the in-memory message store

### P0 — Truncate / virtualize tool output rendering

1. **Change default to collapsed** — Show collapsed tool outputs with a "Show output" button, especially for outputs >10 KB
2. **Add size caps on rendered output** — Cap bash and other tool output DOM at ~10,000 characters with a "Show full output" expand button
3. **Use `IntersectionObserver` for lazy DOM** — Only create tool output DOM nodes when the tool call is near the viewport

### P0 — Automatic message store eviction

1. **Auto-call `clearSession()`** when user navigates away from a session (or after inactivity timeout)
2. **Add per-instance and global memory caps** — e.g., 50 MB per instance, 200 MB global, with oldest-session-first eviction
3. **Add periodic memory pressure check** — poll `performance.memory?.usedJSHeapSize`, trigger eviction when over threshold

### P1 — Clean session state on instance close

1. Add `purgeInstanceSessions(instanceId)` to `stores/session-state.ts` that clears all 15+ per-instance signal entries when an instance is removed
2. Call it from `removeInstance()` in `stores/instances.ts`

### P2 — Streaming delta optimization

1. Buffer chunks in an array and join at end instead of repeated string concatenation (reduces O(n²) temporary allocations during large streaming responses)

## Related Files

- `packages/ui/src/stores/message-v2/instance-store.ts` — message store with no eviction
- `packages/ui/src/components/tool-call.tsx` — tool call expansion defaults
- `packages/ui/src/components/tool-call/renderers/bash.tsx` — no size cap on bash output
- `packages/ui/src/stores/session-state.ts` — signals never cleaned on instance close
- `packages/ui/src/stores/instances.ts` — `removeInstance()` missing session cleanup
- `packages/ui/src/stores/preferences.tsx` — default `toolOutputExpansion: "expanded"`

pascalandr:
## Problem addressed

This PR closes #553 and addresses the linked [Discord report](https://discord.com/channels/1391832426048651334/1458412028325793887/1527610433844805673).

- GitHub #553 reports the Tauri/WebKit renderer growing to 2.5-3.1 GB during long sessions, primarily from fully rendered tool output, graphics layers, and session data retained without a byte limit.
- The Discord report describes the Tauri UI becoming unclickable or crashing out of memory while 3-5 workspaces run concurrently with long sessions and many subagents; a fresh browser client remains responsive.

## What changes for users

- Tool output, diffs, diagnostics, search, and nested-task rendering now have explicit work and DOM limits. Normal rendered output is capped at 10,000 characters while full content remains available through explicit copy actions.
- The existing intentional five-session MRU per workspace remains. A second byte-aware 64 MiB budget applies globally across all workspaces and subagents, so multiple workspaces cannot each retain several arbitrarily large histories. Least-recently-used inactive sessions are evicted first; visible, loading, streaming, compacting, and input-blocked sessions are protected.
- Selecting an evicted session reloads its complete authoritative history from OpenCode. CodeNomad keeps its existing complete-session behavior for search, timeline, usage totals, revert, fork, subagent history, and scroll restoration.
- Obsolete root and subagent requests are cancelled when the user switches session, subagent, workspace, or runtime. Revision-conflict retries remain cancellable during their backoff window, preventing hidden full-history reloads.
- Closing a workspace purges its remaining session, pagination, search, selection, draft, attachment, and message-store state.
- Streaming, reload, deletion, revert, compaction, and workspace-reopen updates are fenced so stale asynchronous work cannot repopulate evicted or closed state.
- In-memory render caches, search scans, pending deltas, SSE frame parsing, diagnostics, and nested-task indexing are bounded independently of transcript residency.

## Scope

This PR does not add persistent or offline transcript caching. OpenCode remains the single authority for message history, and evicted sessions reload their complete history over HTTP.

Existing desktop restoration of workspace tabs, selected sessions, drafts, scroll positions, and layout remains unchanged. Clearing saved startup state does not delete project files or OpenCode conversations.

## Validation

Current head: `59b415c0`

- 202 targeted UI and restore/request-authority tests
- 118 Electron native client-state and lifecycle tests
- UI and Electron typechecks
- Production UI build
- Focused regression coverage for cross-workspace memory eviction, hidden root/subagent request cancellation, revision-retry cancellation, instance-state purge, bounded tool output, SSE runtime authority, and oversized delta recovery

A 24-hour WebKit memory soak and the Discord reporter's exact five-workspace concurrent stress case have not been run. Simultaneously active or streaming sessions are intentionally protected from eviction, so resident message memory may temporarily exceed 64 MiB while work is running; their rendered output remains bounded, and they become eviction candidates once idle. A single visible session also remains complete by design.

Closes #553


pascalandr (commented):
## Gatekeeper verdict: DO NOT SHIP

Reviewed commit `dd3b818e` against `upstream/dev`. The platform tests and builds pass, but the following correctness and memory-safety blockers remain.

1. **P1: normal workspace refreshes invalidate lifecycle fences.** `instances.ts:353-356,765-766,927-933` replaces every existing `Instance` object during SSE reconnect reconciliation. Identity checks in `session-api.ts:926`, `session-actions.ts:468,494`, and `session-message-cache.ts:134,290` then treat an unchanged workspace as removed/reopened. A successful server delete can skip local cleanup, and valid queued cache work can be dropped. Use a stable occurrence token rather than object identity, and change it only on actual remove/reopen.

2. **P1: an idle session can evict active optimistic work.** `session-memory.ts:23-29,52-64` checks only pending input when metadata says `idle`, ignoring normalized `sending` and `streaming` records. Send a prompt, switch away before the working-status event, then trigger memory pressure: the optimistic transcript is cleared while the request remains active. Protect all active normalized work regardless of metadata status.

3. **P1: orphan deltas are unbounded and recovery can lose them.** `delta-buffer.ts:27-32,108-115` re-holds failed deltas with no byte cap, entry cap, TTL, timer, or budget accounting. `session-api.ts:1259-1264,1307-1310` clears held deltas before the replacement HTTP load succeeds; failure/cancellation loses streamed output. The `||=` also stops evaluating later cleanup calls once true. Replace retained payloads with a bounded dirty-session marker plus a coalesced authoritative reload, or retain deltas until successful reconciliation.

4. **P1: same-ID reopen reuses stale HTTP and SSE authority.** `session-state.ts:157-170,1217-1253` deletes search authority so request IDs restart; an old search can match the replacement instance and write old sessions through `session-api.ts:595-646`. `sse-manager.ts:164-170` gates only on current ID membership, so buffered events from the old occurrence are accepted after reopen, while valid startup events are dropped before the instance list loads. Fence both transports with one stable instance-occurrence generation.

5. **P1: persistent cache metadata misrepresents truncated history and can restore stale snapshots.** `session-message-cache.ts:94-121` first truncates to an approximately 8 MiB suffix. `lib/session-message-cache.ts:212-260` receives only that suffix and records `startIndex: 0`, `totalCount: suffix.length`, and `complete: true`. If a delayed writer later skips because the store was evicted or `snapshotSession` returns null (`session-message-cache.ts:124-140`), the older IndexedDB snapshot is left eligible for restore. Preserve original total/start metadata and explicitly invalidate obsolete snapshots when a replacement cannot be written.

6. **P1: large structured tool output becomes inaccessible.** `tool-call/utils.ts:161-171` replaces oversized non-string output for rendering and returns `null` for copy. `tool-call/renderers/default.tsx:8-27` consequently removes the copy action, and long errors are truncated without recovery in `tool-call.tsx:483-489`. This contradicts the full-copy behavior described by the PR. Keep full content behind a lazy copy path without eagerly stringifying it.

7. **P2: cache fencing maps grow for process lifetime.** `session-message-cache.ts:296-325,345-359` retains generation/version keys for every invalidated or destroyed session. This is monotonic memory growth in the feature intended to bound memory. Delete settled keys or scope generations to an instance occurrence.

### Validation

- All application CI tests and cross-platform builds passed.
- `git diff --check` passed and the worktree is clean.
- The red `comment` check is unrelated: the artifact-comment workflow received a GitHub API `504` while paginating runs.
- Missing coverage: actual IndexedDB transactions/failures, idle-plus-sending eviction, orphan-delta bounds/recovery failure, same-ID reopen, and cache/HTTP reconciliation.

The authenticated account owns this PR, so GitHub does not permit a self-authored `REQUEST_CHANGES`; this COMMENT review carries the gatekeeper rejection.

pascalandr (commented):
Gatekeeper verdict: SHIP

Re-reviewed the updated head at c10cab65 after the correction loop. The runtime-authority, persistent-cache, bounded-memory/search, prompt recovery, SSE stream identity, and Tauri coalescing findings are resolved. The final independent gate found no release blocker.

Validation:
- UI: 435/435 tests passed
- Tauri: 85/85 tests passed
- UI, Electron, and server typechecks passed
- UI production build passed
- Server: 248 passed, 4 skipped; the only local failure is the unchanged Windows git-worktrees fixture that assumes the root branch is named main

pascalandr (commented):
## Gate verdict: REQUEST CHANGES

Reviewed latest head `59b415c01c0e5d19040b8bfe0963781b6ca08f35`, all eight PR commits, issue #553, and the complete diff from the true merge base/current `dev` base `1d4f77869e9d7d2c6820bc3f4515dddabab19758`.

### Merge blockers

1. **P1: complete-session search is now silently incomplete.** `packages/ui/src/lib/session-search.ts:34-39,180-214` searches only the newest 10,000 messages, 20,000 parts, 5,000,000 characters, and 1,000 matches. It then scans the retained window oldest-first, so early large parts can exhaust the character budget before newer messages. A fully hydrated long session can therefore report no match for content that is resident and visible, with no partial-results indication. This directly contradicts the PR's complete-session search claim. Scan the authoritative resident transcript incrementally with cancellation/yielding; bound stored/rendered matches and previews, not the searchable corpus, or expose continuation/truncation explicitly.

2. **P1: scroll restoration can commit against a partial post-eviction suffix.** `packages/ui/src/components/message-section.tsx:881-927` no longer waits for `props.loading`; that guard was explicitly removed by cleanup commit `59b415c0`. If an evicted session receives one or more SSE messages and is reopened, restoration runs while the complete HTTP hydration is still in flight. An older anchor is absent from the suffix, so fallback scrolls to bottom and sets `didRestoreScroll`; the subsequent full seed never reapplies the saved anchor. Restore only after loading is false.

3. **P1: reconnect does not reconcile permissions or questions missed during the SSE gap.** `packages/ui/src/stores/instances.ts:276-286` refreshes sessions and one active resident transcript, but never calls the existing authoritative `syncPendingPermissions` / `syncPendingQuestions` paths at `packages/ui/src/stores/instances.ts:501-619`. A request created during disconnection never appears; one answered elsewhere remains actionable and can keep its session protected indefinitely. Run both snapshot reconciliations during the trailing resync, fenced against both runtime replacement and newer accepted interruption events.

4. **P1: revision-conflict recovery has no global or per-session work bound.** After three conflicted full-history loads, `packages/ui/src/stores/session-api.ts:1265-1283` re-enters delta recovery. The active recovery at `packages/ui/src/stores/session-events.ts:487-509` is marked dirty, treats the nested load as successful, resets failures to zero, and immediately repeats. A live session whose revision changes during each HTTP parse can continuously download and parse its complete history. Return an explicit conflict result and use one bounded, backoff-enabled recovery scheduled after streaming settles, with global concurrency limits.

5. **P1: the Tauri event path still permits unbounded frame and delta allocations.** `packages/tauri-app/src-tauri/src/desktop_event_transport/stream.rs:161-189` uses `BufRead::read_line` and accumulates `data_lines` without a line/frame byte limit; a valid CR-only stream or a peer that withholds LF can grow one string indefinitely. `packages/tauri-app/src-tauri/src/desktop_event_transport.rs:448-462` also repeatedly copies coalesced deltas without a byte cap before the UI's 64 KiB delta limit is reached. This leaves the affected Tauri path outside the PR's SSE/delta-bound claim. Parse CR/LF/CRLF incrementally with strict line/frame limits and cap/drop-to-authoritative-reload before concatenation.

6. **P1: stream rotation does not rotate server-side auto-accept authority.** `packages/server/src/permissions/auto-accept-manager.ts:113-117` changes `streamId` but clears only queued hydration events. Old runtime session ancestry, pending permissions, reply attempts, and in-flight replies survive; completions at `packages/server/src/permissions/auto-accept-manager.ts:401-415` are not generation-fenced. After a workspace PID/runtime replacement, an old permission can be sent to the replacement instance, or a late old reply can publish `yolo.autoAccepted` and clear current UI state. Advance the instance generation on stream rotation, clear runtime-scoped pending state, rehydrate persisted root settings, and ignore late reply completions from older generations.

7. **P1: the common tool-output DOM cap is bypassed by always-mounted titles and task summaries.** Collapsed tools render the complete `state.title` / renderer title at `packages/ui/src/components/tool-call.tsx:950-999,1120-1124`. Legacy nested-task rows retain up to 200 entries but render each unbounded title/description at `packages/ui/src/components/tool-call/renderers/task.tsx:430-443,499-523`. A single multi-megabyte title, or 200 large summary descriptions, recreates the oversized DOM #553 is meant to prevent even with output collapsed. Apply a small common header/row text limit while preserving full data only behind explicit lazy copy/access.

### Non-blocking follow-ups

8. **P2: input-blocked payload memory is omitted from the 64 MiB accounting.** `packages/ui/src/stores/message-v2/instance-store.ts:372-396` measures session/message/usage data, then protects sessions with permission/question entries without including those canonical queue objects in the size. Large interruption payloads can exceed the advertised budget invisibly. Include each session-associated request once in retained-size accounting or cap those queues separately.

9. **P2: command and shell failures are all classified as delivery-ambiguous.** `packages/ui/src/stores/session-actions.ts:274-366` wraps even definite HTTP validation failures with `suppressPromptRecovery`; `packages/ui/src/components/prompt-input.tsx:491-509,541-570` has already cleared the text. A definite 4xx loses recoverable shell/command input. Preserve input for definite failures and retain a separate delivery-unknown record for genuinely ambiguous outcomes instead of silently discarding it.

10. **P2: the loopback client replaces caller cancellation and times only response headers.** The installed SDK invokes custom fetch as `fetch(Request)`, but `packages/server/src/workspaces/instance-client.ts:56-62` reads only `init.signal`, replaces the signal carried by the Request, and clears its timer when fetch returns headers. Caller cancellation is lost and a stalled body can hang indefinitely. Combine the Request signal with the timeout and keep timeout authority through body completion, or narrow the documented guarantee and preserve caller cancellation.

11. **P3: cleanup leaves earlier PR-artifact transcript data orphaned.** Current source has no stale transcript-cache imports or runtime hooks, and startup correctly restores UI/session metadata before authoritative HTTP hydration. However, `packages/ui/src/stores/client-state.ts:232-236` clears only native client state; users who ran earlier #626 artifacts can retain the deleted `codenomad-session-messages` IndexedDB database and unsafe marker. Keep a one-time deletion migration, then remove it after affected artifacts age out.

### Better implementation assessment

Removing the experimental persistent transcript cache was the right simplification: OpenCode remains authoritative and the normal desktop tab/draft/scroll restore path is intact. Whole-session eviction is also simpler and safer than partial transcript restoration. The remaining design is not merge-ready because bounds are fragmented across renderers, recovery is recursive instead of scheduled, and runtime authority is not one end-to-end generation across server, transport, and UI.

### Validation assessment

All current GitHub checks pass at `59b415c0`, including Linux/macOS/Windows Electron and Tauri builds, the UI/server/Rust test jobs, and the serialized Windows Tauri suite. `git diff --check` passes and the review worktree is clean. CI success is insufficient here: there is no WebKit soak, native oversized-frame test, evict/SSE/reopen scroll integration, reconnect interruption reconciliation, sustained revision-conflict recovery test, or auto-accept runtime-rotation test. The manually enumerated UI CI list at `.github/workflows/pr-build.yml:104-145` also omits the new `packages/ui/src/components/prompt-input/submitPrompt.test.ts`.

GitHub does not permit a self-authored `REQUEST_CHANGES` review on this PR, so this is published as a COMMENT review with an exact gate verdict of **REQUEST CHANGES**.

pascalandr (commented):
## Gatekeeper verdict: no findings

Reviewed the complete PR range from base \1d4f77869e9d7d2c6820bc3f4515dddabab19758\ through head \2c3dd0a8\ after the remediation passes.

No reproducible high- or medium-severity merge blockers remain in session authority, reconnect/revert handling, revision and delta recovery, memory eviction, full-session search/access, ambiguous prompt recovery, auto-accept rotation, or native desktop event ordering and bounds.

Validated locally with:
- 219 standard UI tests
- 178 browser-condition UI tests
- 72 affected server tests
- 99 locked Rust tests
- 118 Electron native tests
- UI, Electron, and server typechecks
- production UI build
- clean \git diff --check\

Residual risk: native webview reload behavior is covered at the UI/Rust unit boundary rather than by a packaged end-to-end Tauri test.

pascalandr (commented):
## Final trade-off review: REQUEST CHANGES

The memory work has measurable value: the packaged A/B run reduced the repeated-workload heap from 3.20-3.32 GB to 2.46-2.48 GB, released inactive-workspace panes, and removed the baseline Session not found failure. About 55% of the 10,985 added lines are regression coverage, but the final diff still ships roughly 4,850 lines across 75 production files.

Four correctness blockers remain at current head `e517b6a1`:

1. **P1: stale runtime responses return success.** Prompt, custom command, and shell dispatch return normally after the captured runtime is replaced (`session-actions.ts:237-246,345-353,379-388`), after the composer has cleared the draft.
2. **P1: permission approval can be based on a truncated diff.** The permission renderer uses the common 10,000-character prefix while approval buttons remain enabled (`diff-render.tsx:67-69,177-178`; `permission-block.tsx:74-83,103-123`).
3. **P1: same-runtime rehydration does not invalidate earlier transcript loads.** `clearReloadableInstanceState` resets flags but does not abort controllers or advance message epochs (`instances.ts:895-922`; `session-state.ts:61-79`; commit paths in `session-api.ts:1310-1448`).
4. **P1: unrelated metadata mutations suppress authoritative status recovery.** `fetchSessions` skips the entire session after a metadata mutation, including the independently fetched status snapshot (`session-api.ts:552-574`).

The 64 MiB budget also remains intentionally soft because five mounted MRU panes and active work are protected. That limitation is acceptable if described as controlled reclamation rather than an absolute process-memory bound.

GitHub does not allow a self-authored REQUEST_CHANGES review, so this COMMENT carries the gate verdict. I recommend fixing the four blockers before merge, then repeating the focused regression suite, typecheck, packaged smoke test, and final review.

pascalandr (commented):
## Gatekeeper verdict: SHIP

Re-reviewed the complete blocker-fix range from `e517b6a1` through current head `93fb8a74` directly, without delegated review. No high- or critical-severity blocker remains in the four rejected areas.

Resolved:

1. Prompt, custom-command, and shell responses that settle after runtime replacement now return delivery-ambiguous instead of success. Prompt dispatch exits immediately without verifying against the obsolete runtime, and its optimistic message is removed.
2. Permission diffs beyond the 10,000-character render limit now keep both approval buttons and Enter/A shortcuts blocked until copying the complete patch succeeds. Reject remains available.
3. Same-runtime rehydration now aborts every active transcript controller and advances its epoch, preventing pre-reset HTTP work from committing afterward.
4. A newer SSE metadata mutation no longer suppresses the independent authoritative status snapshot. Status is merged field-by-field while newer title/metadata/revert authority remains intact.

Validation at `93fb8a74`:

- focused blocker regressions: 49/49
- CI-equivalent runnable UI tests: 223/223
- CI-equivalent browser/authority tests: 186/186
- UI typecheck and production build: pass
- GitHub tests and locked Windows Tauri tests: pass
- Electron builds: Windows, Linux, macOS pass
- Tauri builds: Windows, Linux, macOS x64 and arm64 pass
- `git diff --check`: pass

Trade-off conclusion: the change is worth shipping. The packaged A/B result remains a measured 21-26% reduction on the identical repeated workload, with inactive-workspace reclamation reaching 0.81-0.86 GB and no baseline `Session not found` failure. Roughly 55% of the large diff is regression coverage rather than shipped logic.

Accepted residual risks: the 64 MiB transcript budget is intentionally soft for mounted MRU panes and active work; a single visible or actively streaming session can exceed it; and the exact 24-hour, five-workspace reporter soak has not been run. These are disclosed design limits rather than hidden correctness failures, and should remain follow-up observability/soak work.

GitHub cannot record an approval on a self-authored PR, so this COMMENT carries the final SHIP verdict.

PR #623, opened 2026-07-23
fix(files): search large workspaces beyond candidate cap

pascalandr:
## Summary
- count only fuzzy-qualified paths toward the 8000 workspace search candidate cap
- scope the bounded cache to the current query and entry type
- stop recursive directory-link cycles without hiding legitimate alias paths
- expand the @file picker to its parent width and horizontally scroll long paths

## Validation
- 6 focused filesystem search/cache tests pass
- server typecheck passes
- workspace UI/Electron typecheck passes
- UI production build passes
- gatekeeper round 3: zero findings
- full server suite: 241 pass, 3 skip; the unrelated existing Windows git-worktrees fixture fails to parse its mocked branch output

Closes #503

PR #628, opened 2026-07-28
fix(ui): keep relevant file path segments visible

### LINKED ISSUE #603: [Bug]: Cannot see relevant part of file names when path is long
### App Variant

Tauri

macOS 26.3.1

### Issue Summary

When trying to add a file via @, if the file path is very long it will have the final part truncated with `...` and you can't see the file name.

### Steps to Reproduce

1. Open a project with nested folders, like Java or Php
2. Type `@` and a file name that is deeply nested
3. The list of files is truncated and you can't see the actual file names

Image

### Expected Behavior

File name should always appear, including it's parent folder, and the starting folder.

Example from OpenCode CLI itself, much more usable

Image

### Logs & Screenshots

_No response_

### Additional Context

_No response_

pascalandr:
## Summary
- show the file or directory name first in each @ picker result
- show the immediate parent and complete root-aligned directory on separate lines
- preserve horizontal inspection for long directories and reset it when results change
- retain the complete path for assistive technology without changing selection values
- run the focused path-splitting regression test in PR CI

## Context
PR #623 replaced end truncation with horizontal scrolling. This follow-up keeps that behavior while moving the relevant filename and parent to the start of each result, matching the CLI-style expectation from the report.

## Validation
- workspace typecheck passes
- UI production build passes
- focused path test passes
- git diff check passes
- Gatekeeper round 3: zero findings

Closes #603

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -105,6 +105,7 @@ jobs:
         run: >-
           node --import tsx --test
           packages/ui/src/components/session-list-visibility.test.ts
+          packages/ui/src/components/unified-picker-path.test.ts
           packages/ui/src/lib/hooks/use-app-session-capture.test.ts
           packages/ui/src/lib/message-selection-position.test.ts
           packages/ui/src/lib/trailing-resync.test.ts
--- a/packages/ui/src/components/unified-picker-path.test.ts
+++ b/packages/ui/src/components/unified-picker-path.test.ts
@@ -2,23 +2,20 @@ import assert from "node:assert/strict"
 import test from "node:test"
 import { splitDisplayPath } from "./unified-picker-path"
 
-test("splits paths into file and abbreviated directory context", () => {
+test("splits paths into file and directory context", () => {
   assert.deepEqual(splitDisplayPath("packages/ui/src/components/unified-picker.tsx"), {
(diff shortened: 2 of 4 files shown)


pascalandr (commented):
Gatekeeper review, round 1: 2 findings. Long filenames could wrap across enough lines to consume the picker viewport, and the omitted middle path was unavailable to sighted keyboard/touch users because the complete value was only exposed through sr-only text/title. Both are being corrected while preserving the immediately visible filename, parent, and starting folder.

pascalandr (commented):
Gatekeeper review, round 2: 3 findings. Query changes preserved horizontal scroll position, Array.at would fail on older supported WKWebView releases, and an extreme filename could push the parent out of the initial viewport. All three are corrected in ab556674. The existing keyboard limitation of PR #623's horizontal panning was classified as pre-existing; no footer workaround was retained.

pascalandr (commented):
Gatekeeper review, round 3: zero findings. Verified filename/parent/root visibility, complete touch-scrollable directory context, full screen-reader path, unchanged selection values, horizontal reset, older WKWebView compatibility, focused CI coverage, workspace typecheck, UI production build, and diff hygiene.

pascalandr:
@shantur LGTM

PR #630, opened 2026-07-29
refactor(server): route background-process prompt through the instance client factory

heunghingwan:
## Summary

PR #561 introduced `createInstanceClient` so server modules would stop hand-assembling the OpenCode loopback URL and use the generated SDK client. Three call sites adopted it; the background-process completion prompt was the remaining holdout — it still hand-built `http://127.0.0.1:{port}/session/{id}/prompt_async` with manual auth/content-type wiring. This PR routes it through the factory + `client.session.promptAsync`, and consolidates the loopback host constant.

## Why

- **Consistency** — the last hand-built OpenCode-instance loopback URL (explicitly flagged in #561 as a future adoption target) now uses the same version-correct SDK path as the permission replier, yolo metadata, and the opencode updater.
- **Correctness** — the old call sent no directory at all; `notify.directory` was stored but never used (dead data). The factory now scopes the prompt to the session's own directory, which for a worktree/subdirectory session is the correct project context (a normal session coincides with the workspace root, so no change there).

## What changed

**Migration** — `background-processes/manager.ts`: `sendCompletionPrompt` → `createInstanceClient(...)` + `client.session.promptAsync({sessionID, parts:[{type:"text", text, synthetic:true}]}, {throwOnError:true})`. Synthetic `<system-message>` text + `synthetic:true` preserved. Side effect: the call now carries the factory's 10s loopback timeout (the old fetch had none; the failure is swallowed + warn-logged, so finalization is unaffected).

**Factory** — `workspaces/instance-client.ts`: new optional `directory` override in `InstanceClientOptions` (defaults to workspace root); adopts shared `LOOPBACK_HOST`. Fetch wrapper unchanged from `dev`.

**Consolidation** — new `workspaces/loopback.ts` exporting `LOOPBACK_HOST = "127.0.0.1"`, adopted by `instance-client.ts`, `instance-events.ts`, and the workspace manager's health + TCP probes (`manager.ts`). (Other `127.0.0.1` uses — remote-access proxy, sidecars — are different concerns, left alone.)

**Tests**
- `workspaces/instance-client.test.ts` (8 cases): null-when-no-port, loopback host/port targeting, auth header attach/omit, directory scoping present/absent, explicit directory override, timeout aborts stuck call.
- `background-processes/manager.test.ts` (2 cases, first test for this manager): drives the real lifecycle (spawn + exit) against a mocked transport; asserts the `prompt_async` POST URL/body/authorization/`x-opencode-directory` (= session dir, distinct from workspace root), and that a failed prompt is swallowed + warn-logged without aborting finalization.

## Behavior parity

| Aspect | Before | After |
|---|---|---|
| Route / body | `/session/{id}/prompt_async`, `{parts:[{type,text,synthetic}]}` | **same** (via SDK) |
| Auth header | manual | via factory |
| Directory scoping | none | session's `notify.directory` |
| Loopback timeout | none | 10s (factory default) |
| Error handling | throw on non-2xx → caught + warn-logged | `throwOnError` → caught + warn-logged (**same**) |

## Testing

10 new tests pass. The background-process integration test is stable 12/12 under `--test-concurrency=4` and is mutation-killed if the `directory` override or `throwOnError` is removed. Full server suite: 255 pass / 1 fail — the single failure is pre-existing (`git-clone.test.ts` Chinese-locale git message, confirmed failing identically on `dev`). Server typecheck clean.

## Risk / rollback

Additive refactor; the factory is production-proven by #561's consumers. Rollback = revert the single commit.

## Notes

- Per the file-length guideline: `packages/server/src/background-processes/manager.ts` is ~684 lines (above the 500 warn threshold, under the 800 limit); this PR shrank it by ~16 lines net.

### DIFF
--- a/packages/server/src/background-processes/manager.test.ts
+++ b/packages/server/src/background-processes/manager.test.ts
@@ -1,0 +1,153 @@
+import assert from "node:assert/strict"
+import { describe, it } from "node:test"
+import { promises as fs } from "node:fs"
+import path from "node:path"
+import os from "node:os"
+
+import { BackgroundProcessManager } from "./manager"
+import type { WorkspaceManager } from "../workspaces/manager"
+import type { EventBus } from "../events/bus"
+import type { Logger } from "../logger"
+
+const WORKSPACE_ID = "ws-test"
+const SESSION_ID = "sess-1"
+const INSTANCE_PORT = 9999
+const AUTH_HEADER = "Basic test-auth"
+const TERMINAL_TIMEOUT_MS = 3000
+
+interface CapturedRequest {
+  method: string
+  url: string
+  headers: Headers
+  body: string
+}
+
+/**
+ * Drives the real {@link BackgroundProcessManager} lifecycle (spawn a
+ * fast-exiting command with notify enabled) against a mocked transport, so the
+ * migrated `sendCompletionPrompt` path — factory + SDK client + `fetch` — is
+ * exercised end to end without touching production wiring.
+ *
+ * The workspace temp directory is intentionally left in place (under
+ * `os.tmpdir()`, OS-reaped): removing it from the test races the manager's
+ * asynchronous finalization writes, which intermittently fail with ENOENT.
+ */
+async function runCompletionPrompt(
+  fetchImpl: (input: Request, init: RequestInit | undefined) => Promise<Response>,
+): Promise<{ requests: CapturedRequest[]; warned: boolean; directory: string }> {
+  const requests: CapturedRequest[] = []
+  const originalFetch = globalThis.fetch
+  // Captured now but swapped in only inside the try below, so a failure during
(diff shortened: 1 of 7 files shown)


pascalandr (approved):
Reviewed exact head `784e73d493620dd0479da50e2ba3e00362229ed8` against `dev`.

No blocking findings. The SDK migration preserves the prompt route, body, authentication, directory scoping, error handling, and finalization behavior.

Validation:
- Server typecheck passed.
- Focused tests: 10 passed.
- Full server suite: 255 passed, 1 unrelated environment-dependent failure, 4 skipped.
- Windows Tauri rerun passed; the original cross-host failure was unrelated to this server-only diff.

Non-blocking: `manager.test.ts` intentionally leaves small `bp-test-*` temporary directories behind. These should be cleaned up when practical.

**Gatekeeper verdict: APPROVE.**

Required packaging checks should still complete before merge.

PR #632, opened 2026-08-02
fix(server): retry Windows reserved auto ports

### LINKED ISSUE #627: [Bug]: EACCES on port 9899 due to Windows WinNAT excluded port range
### App Variant

Tauri

Windows 11 Pro 25H2

### Issue Summary

CodeNomad fails to start on Windows with `CLI exited early: exit code: 1` when port 9899 falls inside a WinNAT excluded port range, causing `listen EACCES` instead of binding successfully. The issue is intermittent because the excluded range varies across boots.

### Steps to Reproduce

1. On Windows with Hyper-V / WSL / Docker enabled (WinNAT active)
2. Install CodeNomad 0.18.0 Tauri build
3. Launch from desktop shortcut
4. Observe: app fails with `CLI exited early: exit code: 1`

To confirm: run `netsh int ipv4 show excludedportrange protocol=tcp`. If the range `9893-9992` appears, both default ports (9898, 9899) are blocked. Setting `CLI_HTTPS_PORT=9891 CLI_HTTP_PORT=9892` (outside the range) makes CodeNomad start normally.

### Expected Behavior

CodeNomad starts successfully regardless of WinNAT's excluded port range, or at minimum surfaces an actionable error telling the user which port is reserved and how to fix it.

### Logs & Screenshots

[INFO] [app] Starting CodeNomad CLI server options={"httpsPort":9898,"httpPort":9899,...} [ERROR] [app] CLI server crashed err=Error: listen EACCES: permission denied 127.0.0.1:9899 at Server.setupListenHandle [as _listen2] (node:net:1918:21) [tauri-cli] cli process exited before ready: Some("CLI exited early: exit code: 1")

### Additional Context

**Root cause**: CodeNomad's default HTTP port (9899) and HTTPS port (9898) can fall within WinNAT's dynamically reserved excluded port range on Windows. Unlike `EADDRINUSE`, Windows returns `EACCES` on `listen()` for excluded ports — the port is system-reserved, not in use.

**Workaround**: Set user environment variables `CLI_HTTPS_PORT=9891` and `CLI_HTTP_PORT=9892`.

**Suggested fixes**:

- Move default ports outside typical WinNAT excluded ranges (e.g. below 9000)
- Allow port configuration in `config.yaml` for the Tauri desktop app
- Detect `EACCES` on `listen()` and surface an actionable message (e.g. "Port 9899 is reserved by Windows. Try setting CLI_HTTP_PORT or check `netsh int ipv4 show excludedportrange`")

pascalandr:
## Summary
- retry an ephemeral port when an automatic listener receives Windows `EACCES`
- keep explicitly configured ports strict
- keep non-Windows `EACCES` behavior unchanged

## Context
CodeNomad already treats 9898 and 9899 as preferred rather than mandatory ports. The fallback previously handled only `EADDRINUSE`, so a Windows port reservation could terminate Tauri startup before the operating system was asked for an available port.

The specific WinNAT attribution in the issue remains unverified, but the observable startup failure is fixed regardless of which Windows reservation produced `EACCES`.

## Validation
- focused listener retry test passes
- server typecheck passes
- server suite: 256 passed, 4 skipped, 1 pre-existing Windows-only `git-worktrees.test.ts` fixture failure
- autonomous gatekeeper review will be published as a PR comment after opening

Fixes #627

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -278,8 +278,8 @@ function resolveHost(input: string | undefined): string {
   return trimmed
 }
 
-function programHasArg(argv: string[], flag: string): boolean {
-  return argv.includes(flag)
+export function programHasArg(argv: string[], flag: string): boolean {
+  return argv.some((argument) => argument === flag || argument.startsWith(`${flag}=`))
 }
 
 async function main() {
--- a/packages/server/src/server/__tests__/listener-retry.test.ts
+++ b/packages/server/src/server/__tests__/listener-retry.test.ts
@@ -1,0 +1,11 @@
+import assert from "node:assert/strict"
+import test from "node:test"
+
+import { shouldRetryPreferredPort } from "../http-server"
+
+test("automatic listeners retry Windows reserved ports without masking explicit failures", () => {
+  assert.equal(shouldRetryPreferredPort({ code: "EADDRINUSE" }, true, "linux"), true)
+  assert.equal(shouldRetryPreferredPort({ code: "EACCES" }, true, "win32"), true)
+  assert.equal(shouldRetryPreferredPort({ code: "EACCES" }, true, "linux"), false)
+  assert.equal(shouldRetryPreferredPort({ code: "EACCES" }, false, "win32"), false)
+})
--- a/packages/server/src/server/__tests__/listener-retry.test.ts
+++ b/packages/server/src/server/__tests__/listener-retry.test.ts
@@ -1,6 +1,7 @@
 import assert from "node:assert/strict"
 import test from "node:test"
 
+import { programHasArg } from "../../index"
 import { shouldRetryPreferredPort } from "../http-server"
 
 test("automatic listeners retry Windows reserved ports without masking explicit failures", () => {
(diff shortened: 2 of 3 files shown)


pascalandr (commented):
## Autonomous gatekeeper review

Two independent AI gatekeepers reviewed the public PR after it was opened.

### Round 1

**Correctness review: one medium finding**

`--http-port=<n>` and `--https-port=<n>` were not recognized as explicit arguments because the existing detection only matched standalone flags. The new fallback could therefore have silently replaced an explicitly selected equals-form port with an ephemeral port.

**Requirements review: SHIP**

The observable Windows startup failure is addressed independently of the unverified WinNAT attribution, and `Fixes #627` is justified.

### Correction

Commit `04d8e75a` now recognizes both supported explicit forms:

- `--http-port 9899`
- `--http-port=9899`

It also rejects similarly prefixed unrelated arguments. Focused regression coverage was added.

### Round 2

- Correctness gatekeeper: **SHIP**
- Requirements gatekeeper: **SHIP**
- High findings: **0**
- Medium findings: **0**

### Validation

- focused listener/CLI tests pass
- server typecheck passes
- full server suite before the gatekeeper correction: 256 passed, 4 skipped, with the known unrelated Windows `git-worktrees.test.ts` fixture failure

### Residual risk

Low: retry selection is covered through the decision predicate rather than an injected Fastify `listen()` failure. The production call site uses that predicate directly.

**Final verdict: SHIP**

pascalandr:
@shantur LGTM

PR #637, opened 2026-08-05
feat(ui): manage visible provider models

### LINKED ISSUE #636: Feature Request: Model & Provider Management — selective model enable/disable
### Problem

CodeNomad currently has no model management UI or configuration. When a provider plugin (e.g. @omniroute/opencode-plugin) emits a large catalog — 600+ models from a live /v1/models endpoint — the model picker becomes completely unusable. There is no way to:

- Disable individual models from a provider
- Whitelist only the models actually used in a project
- Persist model visibility preferences across sessions

The OpenCode Windows desktop app does offer per-model enable/disable toggles in its settings UI, which partially mitigates this. However:

- CLI/TUI users (the primary CodeNomad interface) have no such UI
- Even in the desktop app, toggling off 570+ models one by one is tedious
- There is no config-file equivalent (e.g. provider.<id>.models.<model>.enabled: false)

### Current workaround (fragile)

A second curator plugin registered after the provider plugin can re-overwrite config.provider[<id>] with a filtered model set via the config hook. This requires:

1. Understanding OpenCode/CodeNomad's config-hook merge semantics
2. Maintaining a separate plugin file with a hardcoded allowlist
3. Knowing that plugin array order matters
4. Full provider entry replacement (nested mutation breaks the provider)

### Suggested solution

**Provider-level model management** — either via config or a dedicated UI:

**Config approach (simplest):**
When enabled: false (or the model is absent from the config while the provider emits it), the model is hidden from the picker. When no enabled filtering is configured, all models are shown (current behavior).

**UI approach (power-user):**
A /models manage command or settings panel that lists all models per provider with enable/disable toggles, persisting to the config file.

### Broader scope

This request extends to general provider management:
- Enable/disable entire providers
- Reorder providers in the picker
- Set default models per agent type
- Bulk enable/disable operations

### Environment

- CodeNomad / OpenCode version: 1.18.13
- Plugin: @omniroute/opencode-plugin v0.2.1
- Provider: OmniRoute (local instance at http://127.0.0.1:18080)
- OS: Windows 11

pascalandr:
## Summary
- add per-provider model visibility controls to Settings > Providers and the floating provider dialog
- persist exact hidden model IDs as CodeNomad presentation preferences
- keep large plugin catalogs manageable with search and Show all / Hide all actions

## Behavior
- all current and newly reported models remain visible by default
- unchecked models are hidden only from model picker lists
- active/default models remain usable even when hidden
- favorites remain persisted and return when their models are shown again
- the same management UI is available from Settings and the floating model-picker dialog

This intentionally mirrors OpenChamber/OpenCode Desktop client-side hiding. It does not rewrite OpenCode configuration or change model execution/default semantics.

## Validation
- npm run typecheck
- node --import tsx --test packages/ui/src/lib/model-visibility.test.ts
- Gatekeeper reviews: PASS

Closes #636


pascalandr (commented):
## Gatekeeper review — FAIL

Reviewed current head `de7e7a11e46cc878aded3841d3e1450dd60ecfb1` against `dev`.

### Findings
1. **High — blocking:** The new provider Refresh path calls `global.dispose()` before reloading providers (`provider-manager-modal.tsx:366-374`). An ordinary refresh can dispose OpenCode runtime instances and interrupt active work.
2. **Medium:** A hidden active model remains the controlled Combobox value but is absent from its options (`model-selector.tsx:83-88, 118-121, 158-162, 290-302`), risking inconsistent keyboard/screen-reader state.
3. **Medium:** Model-visibility persistence can appear successful when the server cannot write the YAML file because server write errors are logged but not returned (`preferences.tsx:653-696`; `yaml-doc-store.ts:101-108`).
4. **Low — test gap:** Pure visibility helpers are covered, but not the serialized writer, hidden-current Combobox integration, bulk actions, or floating-dialog focus restoration.

### Validation
- Focused model-visibility tests: 5 passed
- UI typecheck: passed
- UI production build: passed
- `git diff --check`: passed
- i18n keys present in all nine locales

Gatekeeper should be rerun after the blocking refresh behavior is corrected.

pascalandr (commented):
## Gatekeeper re-review — PASS

Reviewed current head `19e5d1318b393f26e390955a4975945bbacdc8c0` against `dev`.

No actionable PR-caused findings remain.

### Verified
- provider Refresh updates shared and modal catalogs without disposing active OpenCode instances
- hidden active models remain usable with collection-safe and provider/ID-aware accessible picker state
- visibility writes are serialized across providers
- settings persistence is single-write and atomic, with rollback, exact mode preservation, private new files, and symlink-chain retention
- SideCar runtime state rolls back on persistence failure; post-commit listener errors are contained
- all nine locales include the accessible label

### Validation
- focused regression tests: 15 passed; 2 POSIX-only tests skipped on Windows
- UI, Electron, and server typechecks: passed
- UI production build: passed
- `git diff --check`: passed

pascalandr:
@shantur LGTM

PR #639, opened 2026-08-06
fix(ui): resolve stuck/duplicated messages after mobile background

JDis03:
## Problem

On mobile, backgrounding CodeNomad while the AI is working suspends JS execution. The SSE connection eventually times out (~45s), and any events that arrived while suspended (a tool call finishing, a message completing) are lost. Returning to the app could leave the UI permanently stuck showing "running" even though the work had actually finished on the server — the only way out was aborting the request.

## Fix

**1. `useForegroundRefresh` hook** (`packages/ui/src/lib/hooks/use-foreground-refresh.ts`)

Subscribes to the SSE transport's `disconnected -> connected` transition (not `visibilitychange`) and re-fetches session status + force-reloads messages only when a reconnect follows a real disconnect. An earlier attempt triggered on `visibilitychange` after a fixed delay, but that force-reloaded even when the connection had stayed alive the whole time, clearing in-flight "sending" state and making it look like the AI never responded to a message the user had just sent.

Also fixes a latent bug in `server-events.ts`: `onPing` could fire for a stale SSE generation after a reconnect, sending a pong tied to the wrong connection. Guarded with the existing `connectGeneration` counter.

**2. Two `hydrateMessages` bugs** found while testing the hook against real mobile background/foreground cycles (`packages/ui/src/stores/message-v2/instance-store.ts`):

- **Unnecessary full re-render on every reload.** The force-reload path always bumped every message's revision regardless of whether the server returned identical content, invalidating render caches and re-rendering the entire visible session on every reconnect. Several reconnects in a row produced perceptible lag. Fixed by only bumping revision when a message's parts or status actually changed — reload time dropped from ~280-320ms to ~10ms for unchanged content.
- **Duplicated message bubble on a race.** If the user sends a message right as a reconnect happens, the optimistic "sending" bubble (client-side temp id) isn't in the REST snapshot yet and was silently dropped from the visible list — without being deleted from the store. When the real SSE echo for it later arrived, the code could no longer find it to swap cleanly, so it created a new record instead, and a subsequent reload could leave both the orphaned bubble and the new one visible at once. Fixed by preserving pending "sending" messages across a reload until they're actually confirmed.

## Validation

- `npm run typecheck` (workspace `@codenomad/ui`): clean
- `node --test src/stores/message-v2/instance-store.test.ts`: 4/4 pass (2 new tests covering the duplicate-message race)
- Full UI test suite run file-by-file: 57/66 pass; the 9 that don't fail identically on a pristine `upstream/dev` checkout with none of this PR's changes applied (pre-existing solid-toast SSR issue in the test environment, unrelated to this change)
- `npm run build` (workspace `@codenomad/ui`): succeeds
- Manually verified over ~1.7 hours of real mobile background/foreground cycles (14s glances up to two 40+ minute backgrounds) — refresh fires only on genuine reconnects, completes in ~200-450ms, no stuck "running" state and no duplicated messages on return

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -120,7 +120,9 @@ jobs:
           packages/ui/src/stores/client-state-codec.test.ts
           packages/ui/src/stores/client-state.test.ts
           packages/ui/src/stores/instances-restore-cancellation.test.ts
+          packages/ui/src/stores/message-v2/instance-store.test.ts
           packages/ui/src/stores/message-v2/message-hydration-authority.test.ts
+          packages/ui/src/stores/message-v2/message-status.test.ts
           packages/ui/src/stores/session-generation-recovery.test.ts
           packages/ui/src/stores/session-metadata.test.ts
           packages/ui/src/stores/session-pagination.test.ts
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -107,6 +107,7 @@ jobs:
           packages/ui/src/components/session-list-visibility.test.ts
           packages/ui/src/components/unified-picker-path.test.ts
           packages/ui/src/lib/hooks/use-app-session-capture.test.ts
+          packages/ui/src/lib/hooks/use-foreground-refresh.test.ts
           packages/ui/src/lib/launch-errors.test.ts
           packages/ui/src/lib/message-selection-position.test.ts
           packages/ui/src/lib/trailing-resync.test.ts
@@ -133,6 +134,8 @@ jobs:
           node --conditions=browser --import tsx --test --test-force-exit
           packages/ui/src/stores/instances-restore-ownership.test.ts
           packages/ui/src/stores/permission-lifecycle.test.ts
+          packages/ui/src/stores/session-request-authority.test.ts
+          packages/ui/src/stores/session-send-lifecycle.test.ts
 
       - name: Test server
         run: node --import tsx --test "packages/server/src/**/*.test.ts"
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -272,13 +272,14 @@ const App: Component = () => {
       //     re-fetches lazily on next activation instead of returning stale
       //     content (a non-forced load short-circuits once "loaded").
       const instanceIds = Array.from(instances().keys())
-      await Promise.all(
-        instanceIds.map((id) =>
-          fetchSessions(id).catch((error) =>
-            log.error("Foreground refresh: fetchSessions failed", { instanceId: id, error }),
-          ),
-        ),
-      )
+      const sessionListResults = await Promise.allSettled(instanceIds.map((id) => fetchSessions(id)))
+      const failedInstanceIds: string[] = []
+      sessionListResults.forEach((result, i) => {
+        if (result.status === "rejected") {
+          failedInstanceIds.push(instanceIds[i])
+          log.error("Foreground refresh: fetchSessions failed", { instanceId: instanceIds[i], error: result.reason })
+        }
+      })
 
       const activeInst = activeInstance()
       const activeSession = activeSessionIdForInstance()
@@ -293,9 +294,19 @@ const App: Component = () => {
         }
       }
 
+      let activeReloadFailed = false
       if (hasActive) {
         const statusBefore = getSessionStatus(activeInst!.id, activeSession!)
-        await loadMessages(activeInst!.id, activeSession!, { force: true })
+        try {
+          await loadMessages(activeInst!.id, activeSession!, { force: true })
+        } catch (error) {
+          activeReloadFailed = true
+          log.error("Foreground refresh: active session reload failed", {
+            instanceId: activeInst!.id,
+            sessionId: activeSession,
+            error,
+          })
+        }
         const statusAfter = getSessionStatus(activeInst!.id, activeSession!)
         log.info("Foreground refresh: active session reloaded", {
           instanceId: activeInst!.id,
@@ -303,6 +314,15 @@ const App: Component = () => {
           statusBefore,
           statusAfter,
         })
+      }
+
+      // Report failure so the hook keeps its dirty latch and retries on the
+      // next reconnect instead of treating a partial recovery as success.
+      if (failedInstanceIds.length > 0 || activeReloadFailed) {
+        throw new Error(
+          `Foreground refresh incomplete: ${failedInstanceIds.length} session-list fetch(es) failed` +
+            (activeReloadFailed ? ", active session reload failed" : ""),
+        )
       }
     },
   })
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -55,6 +55,7 @@ import {
   updateSessionModel,
 } from "./stores/sessions"
 import { useForegroundRefresh } from "./lib/hooks/use-foreground-refresh"
+import { messagesLoaded, invalidateSessionMessageLoad } from "./stores/session-state"
 
 import { hasWakeLockEligibleWork, getSessionStatus } from "./stores/session-status"
 import { openSettings } from "./stores/settings-screen"
@@ -261,19 +262,48 @@ const App: Component = () => {
 
   useForegroundRefresh({
     onRefresh: async () => {
-      const instance = activeInstance()
-      const sessionId = activeSessionIdForInstance()
-      if (!instance || !sessionId || sessionId === "info") return
-      const statusBefore = getSessionStatus(instance.id, sessionId)
-      log.info("Foreground refresh: fetching session state after SSE reconnect", {
-        instanceId: instance.id,
-        sessionId,
-        statusBefore,
-      })
-      await fetchSessions(instance.id)
-      await loadMessages(instance.id, sessionId, { force: true })
-      const statusAfter = getSessionStatus(instance.id, sessionId)
-      log.info("Foreground refresh: done", { statusBefore, statusAfter })
+      // The SSE transport is global: a reconnect can mean missed events for
+      // EVERY loaded workspace/session, not just the one on screen. So we:
+      //  1. Re-fetch the session list for every instance (status/titles).
+      //  2. Force-reload whatever session is active AFTER the fetch settles
+      //     (so a session switch during the await still lands on the right
+      //     one), since it's the one the user is looking at.
+      //  3. Invalidate the loaded flag for every other loaded session so it
+      //     re-fetches lazily on next activation instead of returning stale
+      //     content (a non-forced load short-circuits once "loaded").
+      const instanceIds = Array.from(instances().keys())
+      await Promise.all(
+        instanceIds.map((id) =>
+          fetchSessions(id).catch((error) =>
+            log.error("Foreground refresh: fetchSessions failed", { instanceId: id, error }),
+          ),
+        ),
+      )
+
+      const activeInst = activeInstance()
+      const activeSession = activeSessionIdForInstance()
+      const hasActive = Boolean(activeInst && activeSession && activeSession !== "info")
+
+      // Invalidate every loaded session except the active one (force-reloaded
+      // below). Snapshot the map first; invalidate mutates it via setState.
+      for (const [instId, sessionSet] of messagesLoaded().entries()) {
+        for (const sId of sessionSet) {
+          if (hasActive && instId === activeInst!.id && sId === activeSession) continue
+          invalidateSessionMessageLoad(instId, sId)
+        }
+      }
+
+      if (hasActive) {
+        const statusBefore = getSessionStatus(activeInst!.id, activeSession!)
+        await loadMessages(activeInst!.id, activeSession!, { force: true })
+        const statusAfter = getSessionStatus(activeInst!.id, activeSession!)
+        log.info("Foreground refresh: active session reloaded", {
+          instanceId: activeInst!.id,
+          sessionId: activeSession,
+          statusBefore,
+          statusAfter,
+        })
+      }
     },
   })
 
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -41,6 +41,7 @@ import {
   stopInstance,
   disconnectedInstance,
   acknowledgeDisconnectedInstance,
+  syncPendingRequests,
 } from "./stores/instances"
 import {
   getSessions,
@@ -78,6 +79,28 @@ import {
   selectSidecarTab,
 } from "./stores/app-tabs"
 const log = getLogger("actions")
+const FOREGROUND_REFRESH_TIMEOUT_MS = 10_000
+
+async function withForegroundRefreshTimeout<T>(
+  promise: Promise<T>,
+  label: string,
+  onTimeout?: () => void,
+): Promise<T> {
+  let timer: ReturnType<typeof setTimeout> | undefined
+  try {
+    return await Promise.race([
+      promise,
+      new Promise<never>((_, reject) => {
+        timer = setTimeout(() => {
+          onTimeout?.()
+          reject(new Error(`${label} timed out`))
+        }, FOREGROUND_REFRESH_TIMEOUT_MS)
+      }),
+    ])
+  } finally {
+    if (timer !== undefined) clearTimeout(timer)
+  }
+}
 
 const App: Component = () => {
   useAppSessionRestore()
@@ -271,8 +294,29 @@ const App: Component = () => {
       //  3. Invalidate the loaded flag for every other loaded session so it
       //     re-fetches lazily on next activation instead of returning stale
       //     content (a non-forced load short-circuits once "loaded").
-      const instanceIds = Array.from(instances().keys())
-      const sessionListResults = await Promise.allSettled(instanceIds.map((id) => fetchSessions(id)))
+      const instanceIds = Array.from(instances().values())
+        .filter((instance) => instance.status === "ready" && Boolean(instance.client))
+        .map((instance) => instance.id)
+      const sessionListResults = await Promise.allSettled(
+        instanceIds.map((id) => {
+          let invalidateSessions = () => {}
+          let invalidatePendingRequests = () => {}
+          return withForegroundRefreshTimeout(
+            Promise.all([
+              fetchSessions(id, {
+                strictStatus: true,
+                registerInvalidation: (invalidate) => { invalidateSessions = invalidate },
+              }),
+              syncPendingRequests(id, (invalidate) => { invalidatePendingRequests = invalidate }),
+            ]),
+            `Foreground refresh for ${id}`,
+            () => {
+              invalidateSessions()
+              invalidatePendingRequests()
+            },
+          )
+        }),
+      )
       const failedInstanceIds: string[] = []
       sessionListResults.forEach((result, i) => {
         if (result.status === "rejected") {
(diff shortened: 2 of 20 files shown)


pascalandr (changes requested):
Reviewed exact head `56c72f97c29801b22ba3ac9802ee1a7a054eb498` against `dev`.

Requesting changes for reconnect correctness:

1. **Blocking: different-ID snapshot confirmation leaves a permanent duplicate.** `hydrateMessages` preserves every unmatched optimistic ID after the server IDs (`packages/ui/src/stores/message-v2/instance-store.ts:482-489`). The optimistic ID is client-only and is not sent in `promptAsync`, so the normal confirmation uses a different server ID. If the refresh snapshot already contains that real ID, later SSE reconciliation sees the real record and skips the temp-ID replacement (`packages/ui/src/stores/session-events.ts:398-405`). This was directly reproduced as visible IDs `real,temp`. The new tests cover an absent snapshot entry and same-ID confirmation, but not this normal different-ID race.

2. **Blocking: reconnect recovery refreshes only the currently selected instance/session.** The global transport can miss events for every loaded workspace, but `packages/ui/src/App.tsx:262-274` refreshes only one active pair. Inactive loaded sessions remain stale, because activating them later performs a non-forced load that returns immediately once messages are marked loaded. A session switch while `fetchSessions` awaits has the same gap.

3. **Ongoing messages can be demoted to complete.** The forced refresh uses `loadMessages`, which maps REST messages to `status: "complete"` even when the assistant message has no end time and the session remains working. This can disable streaming-specific behavior until another qualifying SSE update arrives.

4. **The claimed unchanged-hydration cache preservation is incomplete.** Identical hydration still increments `messageInfoVersion`, which participates in the message-block cache signature, so unchanged snapshots continue invalidating those caches.

Validation completed:
- UI typecheck passed.
- UI build passed.
- Focused message hydration, request-authority, and transport tests passed.
- A dedicated different-ID reproduction confirmed the duplicate.
- `git diff --check` passed.

The Windows Tauri failure is unrelated to this UI-only diff; the exact Rust test passed locally and is a known timing flake.

**Gatekeeper verdict: REQUEST_CHANGES.**

JDis03:
Thanks for the thorough review — all four addressed in `97e4865a`.

**1. Different-id duplicate (blocking).** `hydrateMessages` now reconciles optimistic `sending` bubbles against the snapshot by **content signature** (text + file parts, order-sensitive) instead of blindly preserving:
- If the snapshot contains a **new** server user-message (id not already in `messageIds`) with matching content, the temp record is **dropped and cleaned up** (store keys via `produce`, plus `messageInfoCache` / prompt-display override / record-display cache) — so the `real,temp` duplicate can't form.
- If not yet echoed, it's preserved so the SSE echo still swaps it via `replaceMessageId`.
- Only **new** ids count as a confirmation, so a second in-flight duplicate-text send (`ok` twice) isn't falsely matched against the earlier already-loaded `ok`. Attachment-only sends match by file signature too.
- New tests: different-id duplicate, absent-snapshot preservation, same-id, duplicate-text guard, attachment-only.

**2. Only-active-pair refresh (blocking).** The hook now re-fetches the session list for **every instance**, force-reloads whichever session is active **after** those awaits settle (covers a switch during the fetch), and **invalidates the loaded flag** of every other loaded session so it re-fetches lazily on activation instead of returning stale content.

**3. Demoted-to-complete status.** `loadMessages` now derives status the same way as the live SSE path — user → `complete`; assistant → `error`/`complete`/`streaming` based on error and end time — so an in-flight assistant message isn't demoted.

**4. Incomplete cache preservation.** `messageInfoVersion` (the other half of message-block's cache signature alongside `revision`) is now bumped only when the info content actually changed, so an identical reconnect snapshot keeps both stable → cache hit → no re-render.

Validation: typecheck clean; 13/13 relevant store/hydration tests pass (7 new); full suite file-by-file shows the same 9 pre-existing failures as a pristine `upstream/dev` (solid-toast SSR / markdown timeout), no new failures; build passes.

pascalandr (changes requested):
Re-reviewed exact head `97e4865a6c1a80e30c510308b2e7951c323790dc`.

The original single-message duplicate is fixed, but the replacement introduces another blocking reconciliation bug:

1. **Blocking: matching is not one-to-one.** `serverUserContentSignatures` is a `Set` (`packages/ui/src/stores/message-v2/instance-store.ts:529-534`), and every pending message tests the same membership (`:537-545`). With two simultaneous optimistic `"ok"` sends and one server confirmation, both temporary records are treated as superseded and deleted. A direct reproduction ended with only `real-1`; the still-unsent optimistic message was lost. Matching server confirmations must be consumed one-to-one, as an ordered multiset rather than shared set membership.

2. **The content signature is collision-prone.** `packages/ui/src/stores/message-v2/instance-store.ts:466-489` concatenates unescaped tokens and trims the result. Two text parts `["A", "B"]` collide with one text part `"A\nT:B"`; trailing whitespace also collapses. Both cases incorrectly deleted the pending record in reproduction. Please use a collision-safe structured representation.

3. **Different-ID reconciliation drops prompt-display metadata.** The real record is normalized without the optimistic record metadata (`:554-576`) and the temporary override is then cleared (`:598-603`). The existing SSE replacement path transfers that data (`:1010-1024`). Reproduction confirmed `clientPromptDisplayMetadata` becomes `undefined`, which can change pasted-text/path presentation.

4. **REST/SSE user-message status remains inconsistent.** REST forces user messages to `complete` (`packages/ui/src/stores/session-api.ts:1137-1140`), while SSE derives a user message without `time.end` as `streaming` (`packages/ui/src/stores/session-events.ts:393-396`). Assistant status reconciliation is fixed, but status derivation should be shared for both paths.

The multi-instance/session refresh finding is resolved, and byte-identical message info now keeps its cache version stable. UI typecheck, focused tests, and UI build pass; custom reproductions confirmed the issues above.

The repeated Windows Tauri failure remains unrelated to this UI-only diff and the exact test passes locally.

**Gatekeeper verdict: REQUEST_CHANGES.**

JDis03:
Thanks for the precise reproductions — all four addressed in `6fb31148`.

**1. One-to-one matching (blocking).** The candidate Set is now **per-signature FIFO queues** of new server user-message ids, each consumed exactly once on match. Two simultaneous `ok` sends + one confirmation → first temp superseded, second stays pending until its own confirmation arrives. Regression test reproduces your exact scenario (both temps were dropped before; now `[real-1, temp-2]`, then `[real-1, real-2]` when the second confirmation lands).

**2. Collision-safe signatures.** Signatures are now `JSON.stringify` of typed token arrays (`["text", t]` / `["file", name, url, mime]`) — no delimiter joining, no `.trim()`. `["A","B"]` vs `"A\nT:B"` and trailing-whitespace cases both covered by a dedicated test; contentless messages return `null` and never match.

**3. Prompt-display metadata.** The supersede mapping (temp id → confirming server id) is now tracked; the normalize loop resolves `clientPromptDisplayMetadata` with the superseded temp record as fallback, and cleanup calls `movePromptDisplayOverride(temp → server)` — exactly mirroring `replaceMessageId`. Test asserts the real record inherits the metadata.

**4. Shared status derivation.** New `deriveMessageStatus` helper is the single source of truth for both `session-events` (SSE) and `session-api` (REST), both roles: a REST user message without `time.end` now derives `streaming` exactly as SSE did. Two consistency details handled with it: the streaming UI guards are assistant-scoped (so this is visually inert for user bubbles), and the REST bridge now keeps such user records **non-ephemeral** like the SSE path — previously a REST user-`streaming` record would have been marked ephemeral and kept that flag permanently via upsert inheritance.

Validation: typecheck clean; 15/15 message-v2 tests pass (4 new regression files' worth); full suite file-by-file identical to the upstream pre-existing baseline (8 solid-toast SSR + markdown timeout), zero new failures; build passes.

pascalandr (changes requested):
Re-reviewed exact head `6fb31148ff2236047f850386afc09b08e64ac1d3`.

The one-to-one queues, structured signatures, metadata transfer, and shared status derivation fix the prior reproductions. One blocking ambiguity remains:

1. **Blocking: content matching can delete an unrelated local send.** `packages/ui/src/stores/message-v2/instance-store.ts:531-570` treats any new server user message with the same content signature as confirmation, without identity or timestamp correlation. A local pending `"deploy"` at timestamp 5000 was deleted by an unrelated server `"deploy"` at timestamp 1000 in direct reproduction. Identical messages confirmed out of order can likewise inherit prompt metadata according to snapshot order rather than their actual source.

   The send path already creates the optimistic ID (`packages/ui/src/stores/session-actions.ts:102`), and the installed SDK accepts `promptAsync({ messageID })`, but the request currently omits it (`session-actions.ts:192-208`, `:221-226`). Please send that ID and reconcile by identity rather than content.

2. **Duplicate snapshot IDs consume multiple pending sends.** `instance-store.ts:505-506` preserves duplicate server IDs, and `:544-553` enqueues each duplicate as another confirmation candidate. With two pending `"ok"` sends and one duplicated server record, both pending sends were deleted and the session IDs became `real,real`. Deduplicate inputs by ID before candidate construction.

3. **The new regression tests are not executed by PR CI.** `.github/workflows/pr-build.yml:104-127` omits `instance-store.test.ts` and `message-status.test.ts`.

Validation:
- Prior same-content FIFO, collision, metadata-transfer, and REST/SSE status reproductions now pass.
- Focused tests: 42 passed.
- UI typecheck and build passed.
- `git diff --check` passed.

The Windows Tauri failure remains unrelated to this UI-only diff.

**Gatekeeper verdict: REQUEST_CHANGES.**

JDis03:
All three addressed in `b9e4ea2a`.

**1. Identity-based reconciliation (blocking).** `sendMessage` now passes its optimistic id as `messageID` in the `promptAsync` body (confirmed the installed SDK accepts it: `SessionPromptAsyncData.body.messageID`). The server confirms under the **same id**, so hydration reconciles in place through the existing same-id path (`serverIdSet` hit → normalize updates the temp record directly) and the SSE echo finds the record via `getMessage` — no content matching on the happy path, and identical texts can never cross-attach prompt metadata regardless of snapshot order.

Content matching survives only as a **fallback for different-id confirmations** (older servers), now with the temporal correlation you demonstrated: a candidate must satisfy `candidate.createdAt >= record.createdAt`. Your exact case — server `"deploy"`@1000 vs pending `"deploy"`@5000 — is a regression test: the pending send survives.

**2. Snapshot id dedupe.** `hydrateMessages` collapses repeated input ids (first occurrence wins) before candidate construction. One duplicated server record can no longer consume two pending sends, and `messageIds` can no longer repeat the id (`real,real`). Covered by a regression test with a literally duplicated input entry.

**3. CI coverage.** `instance-store.test.ts` and `message-status.test.ts` added to the "Test changed runnable UI behavior" step in `pr-build.yml`; both run green under `node --import tsx --test` (verified locally — 17/17 pass, including the two new regressions above).

Validation: typecheck clean; full UI suite file-by-file identical to the pre-existing upstream baseline (8 solid-toast SSR files + markdown timeout), zero new failures; build passes.

pascalandr (changes requested):
Sorry I missed some problems in my first review. Fresh full gatekeeper review of exact head `b9e4ea2a6058838644976ea95968f4239540cae9` found additional reconnect-path regressions beyond the earlier reconciliation cases:

1. **Blocking: completed messages are rehydrated as `streaming`.** `packages/ui/src/stores/message-v2/message-status.ts:12-15` checks `time.end`, but the installed SDK contract uses `time.completed` for assistant completion; persisted user messages expose only `time.created`. Applying this helper from `session-api.ts:1132-1147` changes `dev` behavior from complete to streaming after reconnect. Direct reproduction produced `assistant-completed=streaming` and `persisted-user=streaming`. `time.end` belongs to parts, not message completion.

2. **Blocking: the content fallback can consume a prompt from another client.** Identity reconciliation now works end-to-end: the optimistic ID is serialized as `messageID`, and same-ID hydration reconciles correctly. However, `instance-store.ts:564-599` still deletes any local pending send when a newer same-content server message appears. An identical remote prompt reproduced this: the local pending record was removed and its display metadata attached to the remote message. Since supported servers preserve `messageID`, please remove the ambiguous content fallback rather than guessing ownership.

3. **Blocking: failed or partial reconnect recovery is permanently forgotten.** `use-foreground-refresh.ts:43-50` clears the disconnect latch before refresh succeeds and never restores or retries it. `App.tsx:274-280` also swallows individual instance failures, so partial recovery is reported as success. A hook harness confirmed that another `connected` event performs no retry after rejection.

4. **Definitively failed sends now survive later hydration indefinitely.** The optimistic bubble remaining after `promptAsync` rejection is pre-existing, but this PR newly preserves every absent ephemeral `sending` record across authoritative non-empty snapshots (`instance-store.ts:582-599`). A rejected send remained visible as `sending` after forced hydration. Preserve only requests still considered pending, or clean up or mark failed on definitive request errors.

5. **Reconnect recovery cannot reconcile an authoritative empty snapshot.** `session-api.ts:1109-1119` marks the session loaded without clearing stale records, while `hydrateMessages` rejects empty input (`instance-store.ts:500-503`). A forced reconnect load reproduced a stale message remaining visible after the server returned no messages. This was latent on `dev`, but it prevents the new recovery feature from fulfilling its purpose.

Validation:
- UI typecheck passed.
- PR workflow UI suite: 167/167 passed.
- Additional focused suites: 30/30 passed.
- Custom harnesses reproduced all findings above.
- Installed SDK `1.17.8` confirms `messageID` serialization and the `time.completed` contract.
- `git diff --check` passed.

Non-blocking integration gap: transport status is edge-only and not replayed, so remount or manual transport restart gaps can still be missed.

**Gatekeeper verdict: REQUEST_CHANGES.**

pascalandr:
@JDis03 Thanks again for working through all these review rounds. The remaining issues are not all specific to your reconnect implementation. The reconnect refresh is exercising pre-existing inconsistencies between REST hydration, SSE events, and optimistic messages.

I think the fastest and safest path may be to split this into two steps:

1. **Message lifecycle cleanup**
   - use `messageID` as the only confirmation identity;
   - centralize status derivation using the actual SDK contract;
   - make REST snapshots authoritative, including empty snapshots;
   - explicitly handle optimistic messages as pending, confirmed, or failed;
   - run the relevant lifecycle tests in PR CI.

2. **Mobile reconnect refresh**
   - keep the reconnect hook focused on `disconnected → connected`;
   - refresh or invalidate affected sessions;
   - retain a dirty state and perform a bounded retry if recovery fails.

This should remove the need for content-signature reconciliation and make the reconnect change much smaller and easier to validate.

If you prefer, I can prepare the lifecycle cleanup patch myself, or help apply the changes directly to this branch so we can get the PR into a mergeable state faster. Let me know which approach works best for you.

JDis03:
All five addressed in `57b85adb`, restructured exactly along the two-step model you outlined (lifecycle cleanup + focused reconnect):

**1. SDK-correct status derivation.** `deriveMessageStatus` now follows the real contract (verified against the installed SDK 1.17.8 types): assistant completion is `time.completed`; `time.end` is part-level and no longer consulted; user messages are complete once persisted; only assistants carry `error`. Your two reproductions (`assistant-completed=streaming`, `persisted-user=streaming`) are covered by the rewritten `message-status.test.ts`.

**2. Content fallback removed entirely.** Reconciliation is identity-only: since the previous commit sends the optimistic id as `messageID`, a confirmed send reappears under the SAME id and is normalized in place (`previous` carries the metadata over — no transfer plumbing needed). All signature helpers, candidate queues, supersede maps, and the metadata-fallback are deleted. An identical remote prompt can no longer consume a local pending send.

**3. Explicit pending/confirmed/failed lifecycle.** New store-level `pendingSendIds`: `markSendPending` at optimistic insert, `clearSendPending` on `promptAsync` rejection, automatic retirement on same-id confirmation (hydrate and SSE upsert), carry-over in `replaceMessageId`, cleanup in `removeMessage`/`clearSession`/`clearInstance`. Hydration preserves an absent ephemeral `sending` record **only** while in flight; your rejected-send reproduction is now a passing test (`drops a definitively failed send on the next authoritative snapshot`).

**4. Authoritative empty snapshots.** New `reconcileEmptyAuthoritativeSnapshot` store method clears stale records while preserving in-flight sends; `session-api` invokes it on the forced empty branch. A forced reconnect into a zero-message server response no longer leaves stale messages visible (test: `preserves only still-pending sends across an authoritative empty snapshot`).

**5. Failed recovery retries.** The hook keeps its dirty latch until `onRefresh` fully succeeds (rejection leaves it set → next `connected` retries) plus an in-flight guard against overlapping runs. `App.tsx` uses `allSettled`, logs each failed instance fetch, catches the active-session reload error, and **throws** on any failure so partial recovery is never reported as success.

Validation: typecheck clean; 12/12 rewritten lifecycle tests + 4/4 status tests pass (both files already in PR CI since the previous round); full UI suite file-by-file identical to the pre-existing upstream baseline; build passes.

Thanks for the offer to prepare the cleanup patch — since the lifecycle work was already underway here and now matches your outline, I've kept it on this branch; happy to split the commits into two PRs if you'd still prefer that for review.

JDis03:
@pascalandr Yes — your help would be very welcome, thank you!

Everything from your two-step outline is implemented on this branch (`57b85adb`): identity-only reconciliation with the explicit pending/confirmed/failed lifecycle, SDK-correct status derivation, authoritative empty snapshots, and the reconnect hook with dirty-latch retry.

If you'd like to help directly, please feel free to:

- **Push adjustments straight to `JDis03:fix/foreground-refresh-clean`** (maintainer edits are enabled), or
- **Split it into the two PRs you proposed** — lifecycle cleanup first, reconnect refresh on top — if that's still your preferred review path, or
- **Prepare your own lifecycle patch** and I'll rebase the reconnect piece on top of it.

Whichever is fastest for you — I'm happy to adapt. And if you'd rather just re-review the current head as-is, that works too. Thanks again for the thorough reviews; the content-matching rabbit holes you caught would have been nasty in production.

pascalandr (approved):
All blocker reviews are resolved and the full CI matrix passes on commit 879b069d. The reconnect path now uses scoped status/message authority, bounded and fenced refreshes, safe pending-request resync, and same-ID optimistic send reconciliation without duplicate parts.
@JDis03, please confirm the PR works as expected on your side.

JDis03:
@pascalandr — heads-up from validating this on a real mobile device (2 workspaces open, one on a large active session of ~6,578 messages). The authoritative reconnect refresh works correctly, but I found a bandwidth issue worth a follow-up.

**What happens**

On reconnect the large active session's full payload gets downloaded **twice, concurrently**:

```
06:48:45.573  loadMessages HTTP fetch (6578 msgs) force:false   (5565ms)
06:48:45.882  loadMessages HTTP fetch (6578 msgs) force:true    (7802ms)
```

The `force:true` is your intended authoritative reload from `onRefresh`. The `force:false` comes from the active-session `createEffect` in `session-view.tsx`:

```ts
const session = () => props.activeSessions.get(props.sessionId)
```

That effect reads `session()`, so it subscribes to the **entire** reactive sessions map. During `onRefresh`, each of the ~10 concurrent `loadMessages` completions calls `setSessions`, which re-fires this effect **20-40× per reconnect** (I logged 21 firings in a single millisecond). Most no-op thanks to the `isLoading && !force` guard, but the firing that lands in the open window (`isLoading:false` + freshly-invalidated flag) issues a real `force:false` network fetch that races your `force:true` reload for the same huge session. On mobile this saturates the link — the ~9 sidebar fetches jump from 1-4s to 10-11s each, and on very large sessions the bounded retries can exhaust against `FOREGROUND_REFRESH_TIMEOUT_MS`.

**How to proceed**

I did **not** change behavior. I tried a generic in-flight dedupe in `loadMessages` first, but `session-request-authority.test.ts` correctly rejected it — a `force:true` must stay authoritative, so collapsing fetches is the wrong lever.

The clean fix is to narrow that effect's reactive dependency so it only re-runs when the active session **id** changes, not on every metadata/status mutation of the whole map — e.g. `createEffect(on(() => props.sessionId, ...))` or depending on a `session()?.id` memo. That removes both the reactive storm and the duplicate download, and leaves your `force:true` contract untouched.

Since this is a pre-existing reactivity pattern in `session-view.tsx` (not introduced by this PR), I'd suggest merging #639 as-is and handling this as a separate follow-up — happy to take it. Full live telemetry + a portable debug overlay (Ctrl+Shift+D) are on branch `debug/mobile-overlay` if you want to reproduce.

pascalandr:
Thanks @JDis03. I confirmed the issue in **session/session-view.tsx**: the message-loading effect reads the full reactive session entry, so status or metadata replacements can retrigger the effect and open the duplicate non-force fetch window you measured.

This is a pre-existing reactive-dependency issue and should **not block #639**. The approval stands, merging, thanks again...

### Follow-up scope

Please handle this in a separate PR:

- Key the effect explicitly on **isActive**, **instanceId**, and **sessionId**.
- Do not depend only on sessionId: an already-mounted inactive view must still load when activated.
- Keep the session lookup untracked inside the Solid **on(...)** callback.
- Add a regression test proving status or metadata map replacements do not call **loadMessages** again.
- Keep coverage showing inactive-to-active transitions and session ID changes still trigger the load.

The telemetry from **debug/mobile-overlay** should make it straightforward to confirm that the duplicate 6,578-message download is gone.

PR #644, opened 2026-08-11
fix(ui): stop active-session load effect from re-firing on every session mutation

JDis03:
## Summary

Follow-up to #639. On mobile, validating the reconnect refresh from #639 surfaced a bandwidth regression: the active session's initial message-load effect re-fires on **every** mutation of the reactive sessions map, and one of those re-fires issues a redundant `force:false` fetch that races the authoritative `force:true` reload for the same — often very large — session.

## What happens (validated live on a real mobile device, 2 workspaces open)

The active-session effect in `session-view.tsx` reads the whole session object:

```ts
const session = () => props.activeSessions.get(props.sessionId)
```

Reading `session()` inside the `createEffect` subscribes it to the entire sessions map. During a foreground/reconnect refresh, each of the ~10 concurrent `loadMessages()` completions calls `setSessions`, re-firing the effect **20-40× per reconnect** (21 firings logged within a single millisecond).

Most re-runs no-op via the `isLoading && !force` guard, but the firing that lands in the open window (`isLoading:false` + freshly invalidated flag) issues a real network fetch. On a ~6,578-message active session that produced a duplicate full download racing the authoritative reload:

```
loadMessages HTTP fetch (6578 msgs) force:false   (5565ms)   ← from this effect
loadMessages HTTP fetch (6578 msgs) force:true    (7802ms)   ← authoritative reload
```

both in flight ~300ms apart, saturating the mobile link (sidebar fetches degraded from 1-4s to 10-11s; bounded retries could exhaust against `FOREGROUND_REFRESH_TIMEOUT_MS`).

## Fix

Derive the active session id through a value-diffed `createMemo` and drive the effect from that memo instead of the full session object. The memo only notifies downstream when the id string actually changes, collapsing the storm to a single run per real session change while preserving existing behavior (load on activation and on session switch).

This intentionally does **not** touch the authoritative `force:true` reload contract guarded by `session-request-authority.test.ts` — a generic in-flight dedupe was tried and correctly rejected by that test (force:true must always re-fetch), so narrowing the effect's reactive dependency is the right lever.

## Validation

- `tsc --noEmit` clean
- UI tests: 34/34 (group 1), 18/18 (authoritative group 2, `--conditions=browser --test-force-exit`, incl. `session-request-authority`)
- Production UI build succeeds

Root cause was captured live via a portable debug overlay (branch `debug/mobile-overlay`). Context: NeuralNomadsAI/CodeNomad#639 (comment).

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -132,6 +132,7 @@ jobs:
       - name: Test restore ownership integration
         run: >-
           node --conditions=browser --import tsx --test --test-force-exit
+          packages/ui/src/lib/hooks/use-active-session-message-load.test.ts
           packages/ui/src/stores/instances-restore-ownership.test.ts
           packages/ui/src/stores/permission-lifecycle.test.ts
           packages/ui/src/stores/session-request-authority.test.ts
--- a/packages/ui/src/components/session/session-view.tsx
+++ b/packages/ui/src/components/session/session-view.tsx
@@ -13,6 +13,7 @@ import { clearSessionIdleFade, IDLE_STATUS_VISIBILITY_MS, getSessionStatus, isSe
 import { deleteMessage } from "../../stores/session-actions"
 import { showAlertDialog } from "../../stores/alerts"
 import { getLogger } from "../../lib/logger"
+import { useActiveSessionMessageLoad } from "../../lib/hooks/use-active-session-message-load"
 import { requestData } from "../../lib/opencode-api"
 import { useI18n } from "../../lib/i18n"
 import type { PromptInputApi, PromptInsertMode } from "../prompt-input/types"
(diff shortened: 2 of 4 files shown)


pascalandr (changes requested):
The dependency narrowing is correct and my targeted validation confirms that replacing the active session object while retaining the same ID no longer re-runs the load effect. Please add a runnable regression test for that behavior before merge: the initial active session should trigger one load, replacing/mutating the session value with the same ID should not trigger another, and a real deactivate/reactivate or session-ID change should trigger once. The existing session-request-authority tests protect the force:true contract, but they do not exercise SessionView's reactive dependency, so this exact regression could return unnoticed. Please also add the new test to the explicit PR CI test list.
The current Windows failure is unrelated: 83/84 Tauri tests passed and the concurrent cross-host election test flaked; I have re-run the failed job.

JDis03:
Thanks for the review. Addressed in `3d364e45`.

**Regression test for SessionView's reactive dependency.** I extracted the active-session load trigger (the value-diffed id memo + the hydration-gated effect) out of `SessionView` into a small dependency-injected hook, `useActiveSessionMessageLoad` (`packages/ui/src/lib/hooks/use-active-session-message-load.ts`). `SessionView` now calls that hook with accessors for `isActive`/`instanceId`/`session` and the real `loadMessages` + `waitForInstanceWorkspaceMetadataHydration`, so the component exercises the exact code the test guards — no behavior change from the earlier fix.

The test (`use-active-session-message-load.test.ts`) asserts the contract you described:
- initial active session → exactly one load;
- replacing/mutating the session value with the **same id** (repeated, to simulate the refresh storm) → **no** further load;
- a real session-id change → exactly one more load;
- a deactivate/reactivate cycle → exactly one more load.

I also added a second case for the async gate: a session switched away from while its metadata is still hydrating must not load — only the current session does.

**CI list.** Added the file to the explicit PR test list. It goes in the `--conditions=browser` group because Solid's server build makes `createEffect` a no-op (same reason `app-session-prompt-hydration` skips on server), so it needs the client build to actually run the effect.

Validation: `tsc` clean; new file 2/2; full browser-condition group **20/20** (incl. `session-request-authority`, still guarding the `force:true` contract); production UI build succeeds.

Thanks also for re-running the flaked Windows cross-host election job — agreed it's unrelated.

pascalandr (approved):
Looks good. The new test covers the regression and is included in CI. I also ran it locally: 20/20 tests, typecheck and build pass.

pascalandr:
Thanks @JDis03 !

PR #646, opened 2026-08-12
fix(ui): preserve assistant output after raw HTML

### LINKED ISSUE #645: [Bug]: Certain HTML output in agent response truncates remainder of session
### App Variant

Electron

Ubuntu 24.04

### Issue Summary

Output is cut off visually with certain HTML sequence returned by agent

### Steps to Reproduce

Get the agent to return this:

```
- status: advisory
- reason: correctness
- location: web/editor/js/preview-adapters.ts::buildSandboxSrcdoc, lines 27-29
- evidence: [source] `return \`<!DOCTYPE html><html><head><meta charset="utf-8"><style>${options.styles.join(";")}</style></head><body>${options.html}</body></html>\`;` — and the test only asserts `expect(frame?.getAttribute("srcdoc")).toContain("width: 120px;height: 80px");`
- fix_effort: trivial (1/12; cheap model; ready, ambiguity=0)
- recommendation: Emit a valid stylesheet, e.g. `<style>html,body{margin:0;padding:0;overflow:hidden}</style>` (replacing or wrapping the inert declarations — the widget markup already carries the geometry inline), and add a test that the constructed `<style>` content is parseable as CSS rules rather than raw declaration strings.
```

### Expected Behavior

The entire output should be visible in CodeNomad.

### Logs & Screenshots

Image

### Additional Context

This is a display issue only. The copy button captures the full HTML.

pascalandr:
## Summary
- sanitize raw HTML tokens in assistant Markdown with the existing allowlist
- prevent dropped tags such as `<style>` from consuming or altering the rest of a response
- add a focused regression for the malformed inline-code sequence reported in #645

## Validation
- `node --conditions=browser --import tsx --test --test-force-exit packages/ui/src/lib/markdown.test.ts`
- `git diff --check`

Closes #645

### DIFF
--- a/packages/ui/src/components/message-part.tsx
+++ b/packages/ui/src/components/message-part.tsx
@@ -238,7 +238,7 @@ export default function MessagePart(props: MessagePartProps) {
                     sessionId={props.sessionId}
                     isDark={isDark()}
                     size={isAssistantMessage() ? "tight" : "base"}
-                    escapeRawHtml={props.messageType === "user"}
+                    escapeRawHtml
                     onRendered={props.onRendered}
                   />
                 </Show>
--- a/packages/ui/src/lib/markdown.test.ts
+++ b/packages/ui/src/lib/markdown.test.ts
@@ -112,3 +112,18 @@ describe("renderMarkdown bracket math delimiters", () => {
     }
   })
 })
+
+describe("renderMarkdown raw HTML", () => {
+  it("preserves text after style tags inside malformed inline code", async () => {
+    const content = [
+      "- evidence: [source] `return ",
+      '`<!DOCTYPE html><html><head><meta charset="utf-8"><style>${options.styles.join(";")}</style></head>',
+      '<body>${options.html}</body></html>`;` - and the test remains visible',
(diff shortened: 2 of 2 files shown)


pascalandr (commented):
## Gatekeeper review - PASS

Reviewed current head `246d22606446af035f97c521f0cf81ae590d5967` against `dev`.

No actionable PR-caused findings.

### Verified
- assistant Markdown now uses the existing raw-HTML sanitizer already used for user messages and file previews
- safe allowlisted Markdown HTML remains supported while `style`, `script`, iframe, object, embed, meta, and link elements are dropped
- the issue #645 malformed inline-code sequence no longer lets a `<style>` element consume the remainder of the response
- Markdown render-cache modes already distinguish sanitized and raw output, so existing cached entries do not collide
- no platform-specific changes or new user-visible strings are introduced

### Validation
- focused Markdown tests: 12 passed
- UI typecheck: passed
- UI production build: passed
- `git diff --check`: passed

pascalandr:
@shantur Shantur LGTM

PR #647, opened 2026-08-12
feat(opencode): migrate CodeNomad to native V2

pascalandr:
# OpenCode V2 Migration and Desktop Restructuring

## Summary

This branch replaces CodeNomad's OpenCode V1 SDK, custom plugin, and per-workspace runtime architecture with the native OpenCode V2 client and one shared OpenCode service. It intentionally ships no V1 runtime fallback.

The scope includes native location ownership, Forms and Shell resources, project-wide cursor pagination, bounded resident message windows, streaming reconciliation, duplicate-folder instances, multi-window Electron/Tauri behavior, content-addressed restore, worktree mutation fencing, and reproducible desktop packaging.

The reviewed implementation head is `dea20996`. UI and server resolve `@opencode-ai/client@beta` to `beta-18999`; the runtime CLI is managed independently. The official comparison baseline remains `upstream/beta` at `eb1ac54d73`.

Full details: [`MIGRATION_V2.md`](./MIGRATION_V2.md), [`DESKTOP_V2_COMPARISON.md`](./DESKTOP_V2_COMPARISON.md), and [`dev-docs/DEV_V2_CODE_REDUCTION.md`](./dev-docs/DEV_V2_CODE_REDUCTION.md).

## Stabilization Fixes

- Preserve manual scroll escape across pagination, remote updates, capped-window shifts, tab changes, and desktop restore; only a local prompt submission forces the bottom.
- Keep the timeline resident window capped at 200 messages. Ordered window slides use Virtua shift, while reorder, pagination, and non-aligned replacement remount with a fresh measurement epoch.
- Finalize the temporary `200 -> 201 -> 200` transition in a generation-guarded microtask before paint and compensate followed content growth from a `ResizeObserver` attached to the mounted Virtua root.
- Keep all pending inbox prompts after delivered transcript messages in native inbox admission order.
- Add a persisted `steer`/`queue` follow-up preference; the alternate shortcut uses the inverse while idle sends remain `steer`.
- Edit a pending prompt only after authoritative inbox cancellation succeeds, then restore its text, structured payload, files, and agent references to the composer.
- Expose native `session.background` through the ownership-checked proxy, a prompt action, and `Ctrl/Cmd+B`.
- Remove the duplicate model-selector "show all" action and make the favorites toggle label reflect its current action.
- Restore permission/Form synchronization plus skill, reference, MCP-resource, and websearch-provider reads in the reviewed proxy allowlist.
[...]

PR #648, opened 2026-08-12
perf(V2): bound transcript memory

### LINKED ISSUE #553: [Bug]: WebKit renderer memory grows to 2.5 GB over 24h+ runtime — tool outputs fully rendered with no eviction

pascalandr:
## Summary

- bound Markdown, reasoning, tool output, diagnostics, diffs, task steps, todos, and per-message part DOM work while retaining complete authoritative content behind lazy copy actions
- coordinate inactive transcripts under a byte-aware 64 MiB LRU, protecting visible, loading, live, generating, permission-blocked, and Form-blocked sessions
- evict and reload transcripts through bounded native OpenCode V2 `latest`/`history` windows with opaque cursor authority, stale-request fencing, and preserved scroll state
- purge reducer, render-cache, pending-part, prompt-display, and per-instance/session state so hidden copies cannot bypass the memory budget

This supersedes #626 without modifying or closing that historical PR and closes #553.

## Reconnected to `dev`

After #647 merged the native V2 migration, this branch was rebuilt directly on `dev@52f0e629`; PR HEAD is `f4f02563` (`3,078+/651-`, 78 files).

- keeps only the #648 content delta above `dev`; the functional patch-id is identical to the previously validated branch
- preserves the 64 MiB transcript LRU, bounded 200-message windows, shift-in-place timeline behavior, eviction restoration, and exhaustive task-copy behavior
- retains the release and usage-provider updates included when #647 landed
- keeps persistent Developer Mode without restoring the isolated `developer-run` launcher, legacy shims, or the former Developer Automation settings surface
- contains no commits from #649

## Final OpenCode V2 audit

- remains aligned with the latest published `@opencode-ai/client` beta contract (`beta-18999`)
- routes manual `/compact` through the same per-session admission chain as prompts, commands, and shells
- preserves native `SessionInfo.metadata` across duplicate paginated inventory responses that omit the optional field
- keeps the CodeNomad memory coordinator because native `loadMore({ all: true })` and `session.evict()` do not provide a global 64 MiB LRU or replaceable bidirectional windows

## Validation

- UI and Electron root typechecks; server typecheck
- UI standard matrix: `310/310`
- UI browser matrix: `156/156`
- Electron native matrix: `189/189`
- server matrix: `376` passed, `2` platform skips
- desktop packaging invariants: `3/3`
- targeted timing-sensitive server and Electron tests repeated `10/10` in isolation after intentional local cross-suite contention
- exact old/new patch-id comparison, `git diff --check`, and conflict-marker scan
- all commits unique to #649 verified absent

GitHub CI for `f4f02563` is pending after reconnecting the PR to `dev`.


pascalandr (commented):
## Gatekeeper review, pass 1: changes required

Merge is blocked pending correction of these findings:

- Client replacement currently emits full instance-destruction semantics, which can discard active selection, drafts, expansion state, and transcript state; pending permission/question synchronization can also apply stale responses from the replaced client.
- Subagent transcript loading can start before the child session exists and does not safely share request ownership across multiple mounted consumers.
- Running native refresh/session-status continuations can survive instance teardown and recreate or mutate state after removal.
- Permission resources, task output, todo output, and apply-patch copy paths can hide or make authoritative content inaccessible while still allowing user decisions.
- Retained-size accounting misses Map/Set contents and can double-count shared ArrayBuffer storage.
- Info-view back navigation, same-diff permission identity, search partiality/revealability, descendant totals, and exact task-limit signaling need correction.

I am applying the corrections now and will publish another gatekeeper pass after validation. The PR remains Draft.

pascalandr (commented):
## Gatekeeper review, pass 2: changes still required

The first-pass findings are corrected and the combined branch passes UI typecheck, production build, and 71 focused tests. A second full review found additional blockers:

- Streaming mutations can repeatedly cancel retained-size measurement, leaving active growth unaccounted; LRU touches before accounting are also lost.
- Client/reconnect invalidation must clear old load errors, force authoritative reconciliation, and invalidate request sync even when the shared V2 proxy/client identity is unchanged.
- Shared-load lease completion can overwrite newer authority, and mounted child tasks do not react to loaded-state invalidation.
- Permission keyboard shortcuts can approve while another button is focused; event/reply continuations need current instance/client/request authority.
- Large permission review needs bounded processing and a non-clipboard access path; the modal also needs keyboard focus containment.
- Search must continue after marking a truncated part partial; aggregate-truncated diffs need fallback detection; truncated child steps and Status-tab todos need complete-content access.
- Child deletion must recompute ancestor totals, Info restore must retain its conversation target, and the new runnable tests must be added to CI.

I am applying pass-2 corrections now. The PR remains Draft and blocked until a subsequent pass reports no findings.

pascalandr (commented):
## Gatekeeper review, pass 3: changes still required

Pass-2 corrections now pass full UI/Electron typecheck, 28 standard focused tests, and 63 browser-condition tests. The third full review found remaining interactions:

- External forced transcript reloads can trigger the mounted loader and abort the caller-owned refresh; reload ownership must be single-source.
- Evicting an infinite retained-size estimate can poison LRU arithmetic with NaN.
- Long sessions must follow native V2 message pagination so eviction/reload restores complete history.
- Same-client reconnect authority is still missing from session-info and question reply continuations; permission event merging/reply completion also needs stable request semantics.
- Search indexing must match actually rendered task/apply-patch/output regions and budgets.
- Measurement and async Markdown/diff cache work must not resurrect stale data after mutation/eviction.
- The CI cleanup correctly removes Tauri, but must retain the independent Windows server spawn test.
- Permission shortcuts inside the active modal must remain usable without duplicate document handling.
- Tool-call documentation still describes the old truncation/copy behavior.

I am applying pass-3 corrections and will repeat the gatekeeper review. The PR remains Draft and blocked.

pascalandr (commented):
## Gatekeeper review, pass 4: changes still required

Pass-3 corrections typecheck successfully and fix full V2 message pagination, cache resurrection, LRU Infinity handling, reconnect authority, search alignment, and Windows Node validation. Pass 4 found remaining blockers:

- Authoritative message replacement must purge omitted message render caches; failed/empty accounting must not retain pending LRU touches.
- Tool search/render parity still has gaps for fallback output, diagnostics, collection budgets, task truncation, and input copy format.
- Shared pending-request synchronization needs consumer leases so one timeout cannot cancel reconnect recovery; several same-client reconnect paths still need authority/version checks and delta-buffer cleanup.
- The standard CI partition needs forced exit for tests that retain UI handles.
- Removing Tauri setup from the caller is insufficient while the reusable build workflow still enables Tauri jobs; the DEV-v2 PR path must disable those jobs while retaining the Windows Node spawn test.
- Migration documentation must retain unresolved security work not addressed by this UI-only PR and describe Tauri removal narrowly and accurately.

I am applying pass-4 corrections now and will run another full gatekeeper pass. The PR remains Draft and blocked.

pascalandr (commented):
## Gatekeeper review, post-rebase pass: changes still required

The branch is now rebased onto DEV-v2@c66aac79, which removed the duplicate Rust-native event transport and resolved the prior server/service findings. Redundant reconnect buffering and Tauri workflow changes were removed. Current validation passes full typecheck plus 140 focused tests.

The remaining findings are limited to this PR's renderer-memory scope:

- Per-session derived render/display caches must be included in transcript accounting now that the old five-session cap is gone.
- Structured input and pply_patch still perform a complete eager formatting/join before truncation or copy.
- Search parity has two remaining empty-diff/error-budget gaps.
- Long single task titles need a complete copy path; diagnostic dedupe, empty copy buttons, and live-region review semantics need small corrections.

I am applying this final scoped correction set and will repeat the review. The PR remains Draft.

pascalandr (commented):
## Gatekeeper review, final scoped pass: two axes clear, corrections remain

The rebased branch now has no async-authority finding and no CI/documentation finding. DEV-v2 authorization is included, upstream Tauri validation is preserved, and the current focused suites/typecheck pass.

Remaining renderer-memory findings:

- The permission modal portal must close with its owning hidden instance, and complete permission reviews must not mount multi-megabyte content while collapsed.
- Permission title/resource checks, todo normalization, apply-patch file selection, and transcript measurement still create some eager full-collection snapshots.
- Derived display caches and unknown pending parts need complete accounting/cleanup, and touching an inactive transcript must refresh accounting after render-cache growth.
- Empty child-task scan windows must still expose truncation/full-copy, and search limits must match actually renderable structured output/prompt regions.

I am applying this final renderer/memory correction set. Async and CI/docs gates remain clear; the PR stays Draft until the next pass reports no findings.

pascalandr (commented):
## Gatekeeper review: scoped implementation clear

The branch is now rebased onto DEV-v2@c66aac79 and the duplicate Rust-native Tauri event transport remains removed. The PR was re-scoped to the renderer-memory issue: bounded DOM/eager processing, one global transcript/cache budget, lossless bounded V2 reload, and complete teardown cleanup. Search and permission-policy expansions were removed.

Final validation passes full UI/Electron typecheck, production UI build, 248 standard UI tests, 71 browser-condition UI tests, 118 Electron native tests, and diff checks. The final child-session discovery race was corrected and its focused tests pass.

No scoped gatekeeper finding remains. The PR stays Draft only for the outstanding 24-hour WebKit renderer-memory soak.

pascalandr (commented):
## Revue gatekeeper

### Finding bloquant — corrigé

**[P1] Borner la signature réactive des groupes techniques à la fenêtre réellement rendue**

`technicalGroupingSignature()` parcourait deux fois `record.partIds` et concaténait une signature pour toutes les parties du message, alors que le rendu est volontairement limité à 200 parties. Une réponse contenant un très grand nombre de parties pouvait donc encore provoquer un travail synchrone et des allocations non bornés à chaque invalidation réactive, en contradiction avec l'objectif mémoire de cette PR.

Corrigé par `f9c476f6` : le rendu et la signature partagent désormais la même sélection tête/fin bornée à 200 parties via `getRecordDisplayPartIds()`. Le test de régression vérifie la borne et la conservation du début et de la réponse finale.

Validations ciblées : **44/44 tests réussis**, plus **typecheck UI réussi**.

### Audit OpenCode V2

Contrat publié vérifié jusqu'à `@opencode-ai/client@0.0.0-beta-19157` et branche upstream `beta@7a4ad68a`. Les évolutions postérieures à `beta-18999` concernent notamment l'état provider des compactions, la robustesse des refreshs Solid et le contrat worktree natif ; aucune ne révèle d'autre écart fonctionnel pertinent pour le budget transcript, les fenêtres bidirectionnelles ou l'admission par session de cette PR.

**Verdict après correction : aucun autre finding bloquant.**

pascalandr:
## Refresh against dev aa51cbb9

Published head: `00b27af63c74f8b0ddc1d82226e5e34f13bded6a`

merge(dev): preserve transcript memory bounds on the modern V2 contract

Merge dev aa51cbb9 after the official @opencode/client 2.0.4 migration and
connection-scoped compatibility work in #695. Retain the bounded transcript
windows, task-preview visibility accounting and traversal authority from #648,
while adopting the shared message-content icons and current client imports.
Keep dev's isolated automation registry fixture and both sets of session
request-authority regressions, including migrated global descendants.

Validation: server 500 passed/2 skipped; UI stores 450 passed; Chromium
session-rendering 43 passed; Electron native 194 passed/4 skipped; server,
UI and Electron typechecks passed. No interactive desktop or non-Windows
validation is claimed by this merge.


Validation above is local execution evidence. CI on this published head and the explicitly untested interactive/platform scenarios remain separate. No PR was merged.

Refactor signals: `packages/ui/src/components/tool-call/renderers/task.tsx` ~657 lines; `packages/ui/src/stores/session-api.ts` ~1711; `packages/ui/src/stores/session-request-authority.test.ts` ~1935.

pascalandr (commented):
## Review conclusion: PR #648 remains relevant against current dev

Reviewed head: `41c987ad9eb5298188a3296d3d714eee34bebbe1`, integrating `dev@16930777311eb39ffa388803555f170a490fc8ba`.

### Why this is still worth merging

Current dev already supplies native 200-message windows, direct historical-anchor navigation and pruning-backed bounded history search. Those mechanisms are retained; they are not the remaining justification for this PR.

The remaining value is memory retention and rendering cost:

- Replace the five-cached-sessions-per-instance policy with an estimated-byte-based transcript LRU coordinated across instances in the renderer, targeting 64 MiB and protecting visible, loading, generating and interaction-blocked sessions.
- Keep only the active conversation view in the active project mounted, while retaining reusable transcript data subject to the budget.
- Account for transcript-associated render/display/global caches and clear their retained references coherently on eviction.
- Bound processing/rendering of oversized individual content and pending event fragments, and defer copy projections until requested. A window of 200 messages does not bound the size of those messages.

Thus dev bounds history loading, while this PR additionally manages the retained weight and rendering work of visited conversations. These are complementary improvements.

The 64 MiB budget is not a total application RAM ceiling: accounting is estimated and protected active/live windows may exceed it. No quantitative before/after RAM benchmark was established during this review, so no percentage memory reduction is claimed.

### Integration findings addressed

- Preserve expected-window authority, committed outcomes and AbortSignal handling for bounded navigation.
- Supersede stale navigation reads without treating them as full transcript invalidations; real invalidation and eviction still require authoritative reloads.
- Align edit/read copy tests with lazy getCopyText.
- Restore system-message visibility filtering and reactive search refresh with the current server-backed history contract.
- Pin directory-browser test locale and align system-message fixtures and the search batching assertion with the current behavior.

The concurrently published dev merge was preserved; follow-up commits were pushed as ordinary fast-forwards. No V1 shim, developer-run launcher or Developer Automation panel was restored. OpenCode client/plugin dependencies remain at 2.0.15; review of 2.0.16 found no required contract change. PR #649 was inherited only through dev.

### Validation and disposition

Local validation includes UI/server/Electron typechecks, the server suite (682 passed, 2 platform skips), 107 Electron native tests, 181 Tauri Windows tests, native Tauri menu acceptance, 72 authority/cancellation tests, and the complete Chromium suite (201 passed, 1 skipped). The final batching assertion adjustment passed all 23 pagination tests.

At publication, all test jobs for 41c987ad are green, including Windows/macOS/Linux native pruning, runtime contracts, compatibility, and Tauri Windows/macOS. Desktop artifact builds are still running. The earlier Windows pruning UI timeout did not recur on this run.

Recommendation: retain and merge this PR once artifact builds pass and the required independent approval is obtained. This comment is the review report, not an independent approving review or a claim that the PR has already merged.

pascalandr:
## Measured-review correction: eviction still retains payloads

The functional CI for `41c987ad9eb5298188a3296d3d714eee34bebbe1` is green, but I am withdrawing my earlier recommendation to merge after CI/approval pending a concrete memory-retention fix.

### P1: zero accounting does not mean payload eviction

Using the actual `createInstanceMessageStore` implementation at `41c987ad`, hydrate an assistant message with a 1 MiB text part, then call `evictSessionTranscript(sessionID)`:

- Before: estimated retention approximately 2,098,300 bytes.
- After: estimated retention **0 bytes**, and `sessionOrder` is empty.
- Nevertheless, `getMessage(messageID)` still returns the complete record and its 1,048,576-character text.
- Repeating with twelve distinct sessions leaves all twelve records reachable, despite each eviction reporting zero retention.

Reproduced with both default Node export conditions and `--conditions=browser`. This is synthetic data in a local store, with no daemon/database involved.

`clearSession` in `packages/ui/src/stores/message-v2/instance-store.ts` returns copied objects with deleted keys through Solid `setState`. These updates merge objects, so omitted keys remain. Emptying the session message-ID list then hides the still-reachable records from estimation. Actual property removal/reconciliation and regression assertions on the backing records are required, not only assertions on the session's visible message IDs or reported byte count.

The implementation of this store is unchanged by the subsequently observed `5bc62d65` dev merge.

### Correction to the earlier relevance report

The reviewed base `dev@16930777` already wraps `SessionView` in `Show when={isActive()}`. Five retained cache panes therefore do **not** mean five fully mounted inactive transcripts. The PR's remaining render/content and cache changes must be assessed on their own merits; removing empty pane wrappers should not be presented as the main DOM-memory saving.

Production-UI comparisons on real large conversations are in progress. They use the same Windows Chromium engine and pinned UI commits; they do not establish a macOS WKWebView 24-hour fix. No quantitative benefit is claimed in this update.

pascalandr (commented):
## Independent Gatekeeper review — corrections required

Reviewed source: `41c987ad9eb5298188a3296d3d714eee34bebbe1` against `16930777`. The current head is now `5bc62d651ab2a4e6e0c59e4062c514878d832ccc`; its intervening dev merge leaves all five cited source files unchanged. Local uncommitted repairs on the older `d2b2aff2` worktree were excluded.

### Findings

1. **P1 — Eviction retains payloads while accounting reports zero.** `packages/ui/src/stores/message-v2/instance-store.ts:1819–1820`, underlying deletion at `1742–1746`: `clearSession` omits keys from an object returned to a merging Solid setter instead of deleting them. A 1,048,576-character part remains reachable through `state.messages` after eviction, while its retained-size estimate drops from 2,098,186 bytes to zero. This inherited deletion defect invalidates the new byte-budget mechanism that depends on it. Reproduced independently twice against the published archive.
2. **P2 — Catalogue refresh temporarily unprotects the active transcript.** `packages/ui/src/components/instance/shell/useSessionCache.ts:20–34`: a fresh array identity reruns visibility cleanup, synchronously enforcing the budget before re-pinning the same active session. An over-budget complete active transcript is evicted on a catalogue/title replacement. Reproduced.
3. **P2 — Saved snapshot plus staged undo can show empty history.** `packages/ui/src/stores/session-api.ts:1633–1668`: the refill path normalizes older cursors into `previous`, but staged-tail seeking still reads `next` as though the response were descending. Reopening when the latest page is entirely undone misses the surviving visible page. Reproduced.
4. **P2 — Large-diff fallback permission has no approval recovery.** `packages/ui/src/components/permission-approval-modal.tsx:372–375`: both approval buttons pass a permanently false full-review flag. A valid source-less request cannot acquire a tool attachment, and this fallback provides no full-diff access action to unlock approval. Direct code-path analysis; no rendered-modal reproduction claimed.
5. **P2 — Full child-step copying lacks operation lifetime.** `packages/ui/src/components/tool-call/renderers/task.tsx:452–473`: pagination and clipboard completion are fenced only by child ID, not component disposal or a newer copy action. Hidden pagination can continue and an older operation can overwrite subsequent clipboard content. Direct lifecycle analysis; no rendered click/dispose reproduction claimed.

### Validation and relevance

- UI typecheck and **139 existing focused tests passed** against the clean published-head archive; the diagnostic reproductions expose gaps in that coverage.
- Actual `@opencode/client@2.0.16` artifact comparison: 29/30 declarations identical to 2.0.15; the relevant Solid data implementation changes only an import filename.
- Native 2.0.16 CI evidence at `41c987ad`: [runtime-contracts job](https://github.com/NeuralNomadsAI/CodeNomad/actions/runs/36041231586/job/107773655642), including native cursors and 1,501-message distant-window/payload parity. This is CI evidence, not a local daemon run or renderer-memory soak.
- The PR remains relevant: native pagination/compaction cannot release CodeNomad-owned store references, DOM or derived caches. No mandatory split or closure is justified merely by the existence of native pagination.

### Correction of earlier review conclusions

The earlier `d2b2aff2` review identified integration mismatches, but overstated unused exports and the documented count-to-byte cache tradeoff as HIGH findings. The alleged segmentation regression overlooked explicit per-segment part IDs and was not demonstrated. The deletion-reconnect root cause was missing client identity, not absence of generation checks. Also, that merge's clean-typecheck claim cannot certify its committed duplicate task declaration; the local repair was not in the commit. The later published head fixes that duplicate and other hydration integrations, so those historical findings are not carried forward.

I am applying the verified corrections, then will repeat independent reviews and exact-head validation until no actionable findings remain. The user has authorized administrator merge only after that gate is clear. The seven older uncommitted files remain preserved separately.

pascalandr:
## Follow-up: production-UI measurements on real large transcripts

Compared production bundles of `dev@16930777` and PR `41c987ad` in fresh headed Chromium 153.0.8010.12 processes on Windows, same 1600×1000 viewport and preferences. Eight actual large conversations, three rounds / **24 switches per run**, expanded tool outputs/reasoning, native 200-message windows. Transcript responses were recorded and held fixed; uncaptured auxiliary reads used the same OpenCode 2.0.16 backend. The installed desktop and ongoing native sessions were not restarted.

The primary matched pair completed with **zero page/HTTP errors, zero missing captured responses, and zero pending window/outline requests at sampling**:

| Measure | dev | PR #648 |
| --- | ---: | ---: |
| Renderer + GPU private bytes after natural settling | 887.6 MiB | 859.2 MiB |
| Renderer + GPU private bytes after explicit GC | 855.8 MiB | 824.4 MiB (**−3.7%**) |
| JS heap after GC | 173.7 MiB | 159.4 MiB (**−8.2%**) |
| CDP DOM nodes after GC | 46,811 | 40,691 |
| Attached document elements | 9,397 | 9,394 |
| Highest checkpoint-sampled private bytes during visits | 2,573.0 MiB | 1,445.9 MiB |
| Long-task duration after project opening | 12.007 s | 11.890 s |
| Click → two animation-frame callbacks, median | 79.9 ms | 63.3 ms |
| Same proxy, empirical P95 / 24 clicks | 106.6 ms | **157.4 ms (worse)** |
| JS heap after project-tab close + GC | 163.5 MiB | 138.5 MiB |

The final sequence was dev → PR → PR → dev. The second PR/dev runs also completed all switches with no missing transcript responses, but encountered respectively six/one HTTP 500 responses in auxiliary live endpoints. Their post-GC private bytes were 835.8/874.9 MiB and JS heaps 156.9/178.3 MiB. They are excluded from the clean primary comparison; this small sample does not establish statistical confidence.

**Interpretation:** modest retained-memory savings are observed; sampled transient peaks are lower, but natural GC/allocator variability makes the single-pair peak ratio unsuitable as a guaranteed benefit. The median interaction proxy improves, while its tail worsens and total long-task time is essentially unchanged in the clean pair. There is no basis to claim a general responsiveness improvement.

This is not the macOS WKWebView physical-footprint measurement from #553, nor a 24-hour soak. The original many-generation parallel load ended during preparation; the remaining live workload had only two other active sessions, and the controlled replay froze transcript updates. Private bytes are not total app RAM or VRAM, and tab close is not shared-workspace shutdown.

**Disposition remains blocked by the reproduced false eviction described above:** payloads remain reachable while the byte estimate falls to zero. The measured gain does not make that defect acceptable or establish that #553 is resolved. The base already unmounts inactive transcript views; the near-identical attached element counts also correct the earlier five-mounted-views justification.

Remote head was rechecked as `5bc62d65`, whose newer dev merge does not change this eviction implementation. Measurements remain pinned to `41c987ad`. Raw private transcript fixtures and screenshots are retained locally; no conversation contents are published here, and this measurement pass changes no production source.

pascalandr (commented):
# PR648 corrected candidate — independent memory/concurrency Gatekeeper review, round 2

**Result: 2 actionable findings. The memory/concurrency gate is not clear.**

## Candidate and scope

- Reviewed the candidate in `D:/CodeNomad/.codenomad/worktrees/pr648-review-fixes`: HEAD `5bc62d651ab2a4e6e0c59e4062c514878d832ccc` plus current working changes, against base `89db55d340f942c5523de656dd92aaec6e9f3978`.
- Independently inspected the PR's store ownership, transcript eviction/accounting, global/derived caches, pending parts, page loading/cursors, undo restoration, and request/navigation lifetime changes. Read the published round-1 findings after inspecting the candidate inventory.
- Task-renderer/permission/copy changes still being finished were excluded from findings; another reviewer owns those surfaces.
- Read-only source review. No repository edits, commits, pushes, desktop actions, shared-daemon requests, browser suite, or heavy native pruning runs.

## Actionable findings

### 1. P2 — Idle cleanup cancels undelivered transcript rotation/resync work

**Path/lines:** `packages/ui/src/stores/opencode-data.ts:695–698` (new idle disposal), `704–708`; destruction reaches `invalidateTranscript` at `554–563`. The production call is `packages/ui/src/stores/instances.ts:2029`.

`finishOpenCodeDataEvent` immediately destroys the session reducer on `session.idle`, even if its bounded-window rotation or authoritative resync is still pending. `invalidateTranscript` empties the queued events, cancels the resync timer, and disposes the fresh reducer. The preceding synchronous `project(data)` has only projected the old page, so the pending last assistant message or its terminal updates never reach the visible store.

**Independent reproduction (actual module/store, mocked/no network):** emit 200 completed assistant steps, then emit the 201st step start/text/finish and `session.idle` before the pending rotation finishes. Use the same apply/project/finish ordering as `handleInstanceInvalidation`. After 100 ms the resident store has 200 messages and no `m200` final message. A second A/B probe emits 201 completed steps and idle: omitting only `finishOpenCodeDataEvent` yields `finalMessagePresent: true`; calling it yields `false`.

This is not covered by the existing bounded-burst tests, which allow rotation to settle without terminal disposal. It also affects idle arriving while the existing overflow-recovery HTTP request or retry is outstanding. `handleSessionIdle` does not guarantee recovery: `refreshSettledSessionMessages` (`session-events.ts:196–211, 545–560`) reloads only for an explicit force or a resident pending/running tool. A missing text-only final message satisfies neither condition.

**Requested correction:** retire the reducer after pending work has authoritatively drained, or preserve/fence a mandatory final native transcript reconciliation before disposal. Add a terminal-event test during both a pending rotation and a pending overflow resync; assert the final resident message and completion status survive.

### 2. P2 — Per-session disposal resets mutation authority and admits pre-cancellation history pages

**Path/lines:** `packages/ui/src/stores/opencode-data.ts:710–712`, reached on ordinary idle through `695–698`. Relevant consumer: `packages/ui/src/stores/session-api.ts:1563–1573`.

The new per-session `destroyOpenCodeData` branch deletes `mutationRevisions` and `messageRevisions` without advancing the instance generation or superseding in-flight UI message loads. A history request that captured mutation revision zero can see a cancellation/revert advance it to one, then see idle reset it back to zero. The request's equality fence now passes again and an obsolete response can republish removed content. This is a distinct authority defect from dropping reducer work; simply draining the queue before disposal does not preserve the request fence.

**Independent reproduction (actual `loadMessages` / `loadOldestMessageWindow`, stub client):**

1. Load `latest`, with older cursor `older`.
2. Start an oldest-window request and hold its response.
3. Apply `session.inbox.cancelled` for `deleted`; mutation revision becomes 1.
4. Apply and finish `session.idle`; mutation revision becomes 0.
5. Resolve the held pre-cancellation page containing `deleted`, with a newer cursor.

Observed output: `{"afterCancel":1,"afterIdle":0,"committed":true,"visible":["deleted"]}`. The existing cancellation/revert tests cover the revision increment but do not insert idle cleanup before the old response resolves. The old whole-instance destruction path advances an instance generation; the new per-session path does not.

**Requested correction:** keep per-session mutation/message authority monotonic across reducer eviction, or advance a separate durable per-session generation and incorporate it into all affected request fences. Do not release revision identity merely because its reducer payload can be freed. Add pending-history-response tests with cancellation/revert followed by idle cleanup.

## Corrected round-1 issues checked

- Eviction now actually removes Solid-store payload keys and session-owned indexes; the new regression checks `state.messages`, not just visible IDs or byte totals.
- Catalogue/loading changes preserve the active visibility pin; the new reactive test passed.
- Saved latest-window refill now selects the older cursor according to the response orientation when seeking past staged undo. Multi-page, failure, abort, and fully undone cases passed.

## Validation

Ran this lightweight command against the working candidate:

```powershell
node --conditions=browser --import tsx --test --test-force-exit packages/ui/src/lib/global-cache.test.ts packages/ui/src/lib/retained-size.test.ts packages/ui/src/lib/session-transcript-lru.test.ts packages/ui/src/stores/message-v2/transcript-eviction.test.ts packages/ui/src/components/instance/shell/useSessionCache.test.ts packages/ui/src/stores/message-v2/instance-store.test.ts packages/ui/src/stores/message-v2/message-window.test.ts packages/ui/src/stores/session-request-authority.test.ts packages/ui/src/stores/opencode-data.test.ts
```

**136 tests passed; zero existing-test failures.** The independent stdin probes above used the actual source modules and local fake SDK clients; they made no daemon calls. The final-message A/B probe reproduced the failure twice in total.

## Remaining limits / non-findings

- This is not a zero-findings report or merge certification.
- No native-event timing soak, heap snapshot, desktop rendering verification, or full-browser run was performed. The rotation repro deliberately controls a pending async boundary; the overflow-resync implication follows the same inspected disposal path, rather than a separate native reproduction.
- UI permission/copy work was in progress and intentionally outside this review's gate. Working changes may advance after this report.
- No speculative or unrelated pre-existing issue is included in the actionable count. Both findings are caused by the PR's newly added ordinary-idle/per-session disposal behavior.

pascalandr (commented):
# PR648 independent renderer review — round 2

## Result: 3 actionable findings

Reviewed the full renderer/display/cache/copy changes against `89db55d340f942c5523de656dd92aaec6e9f3978`, starting at `5bc62d65` plus the working repairs. During review, the memory/navigation and fallback-permission repairs were committed as `1d3f85c0` and `467261f2`. Final observed HEAD: **`467261f2`**, with task-copy/activity changes still uncommitted. These findings describe the actual working files read and exercised, not an assumed final implementation. The parent should recheck after the task implementer finishes.

## 1. P2 — Appended content parts do not invalidate the mounted content row

**Location:** `packages/ui/src/components/message-block.tsx:293–305, 691–706` (particularly the plain-object assignment at line 703).

`MessageContentItem.parts` now traverses `props.partIds` instead of the reactive `record.partIds`. But `flushContent` reuses the same cached content item and mutates its ordinary `partIds` property. The enclosing `<Index>` keeps the same item signal when the object identity is unchanged. Consequently, adding a second consecutive text/file part updates the cache but never invalidates the mounted row's `parts` memo. That memo only subscribed to the old part entries, not the new part or the record revision. Newly arriving parts can remain invisible until remount or some unrelated subscribed change.

**Reproduction:** hydrate an assistant message with text part `p1`, mount the same cache + `Index`/`MessageContentItem.parts` computations, then call the real store's `applyPartUpdate` with a new text part `p2`. Independently reproduced with the actual `createInstanceMessageStore`, `buildRecordDisplayData`, and Solid browser-condition `createMemo`/`indexArray` (the primitive used by `<Index>`):

```text
before [ 'p1' ]
store [ 'p1', 'p2' ]
cached [ 'p1', 'p2' ]
after [ 'p1' ]
```

This is a PR-introduced regression: the base implementation iterated reactive `current.partIds`, which notified the memo when parts were appended. It also affects the changing tail of the bounded head/tail projection.

**Fix direction:** publish a fresh content item when its selected IDs change, or make those IDs explicitly reactive. Preserve the bounded per-segment selection. Add a live-append regression that exercises a mounted computation, rather than just validating static selected IDs.

## 2. P2 — Cancelled task copies can still overwrite a newer clipboard value through fallback

**Location:** `packages/ui/src/components/tool-call/renderers/task-copy.ts:83–87`; `packages/ui/src/lib/clipboard.ts:75–85` (and the second fallback in `copyToClipboard:19–50`).

The new task-copy lifetime correctly fences pagination and the initial clipboard dispatch. However, it passes no cancellation/authority predicate into `copyTextChunksToClipboard`. Once `navigator.clipboard.write` is pending, switching sessions cancels the task operation but cannot prevent that helper from starting a *new* write after rejection: it unconditionally calls `copyToClipboard(chunks.join(""))`. `copyToClipboard` likewise has an asynchronous writeText-to-execCommand fallback without a fence. Thus an obsolete operation can overwrite a newer explicit copy after the view has been deactivated/disposed.

**Reproduction:** exercised the actual `useTaskStepCopy`, `copyTextChunksToClipboard`, and `copyToClipboard` with mocked asynchronous clipboard methods:

1. Begin child-step copy and leave `clipboard.write` unresolved.
2. Set the owner's active signal to false (the hook cancels/releases correctly).
3. Copy `NEW USER COPY` using the real text helper.
4. Reject the old `clipboard.write` promise, as can happen after a focus/clipboard-permission change.

```text
after newer copy NEW USER COPY
after stale completion [
{
  "output": "STALE CHILD COPY"
}
]
```

This is the residual clipboard-completion portion of round-1 finding 5, despite the useful new pagination/disposal fencing. It concerns a preventable **new fallback write**, not attempting to retract a write already dispatched to the OS.

**Fix direction:** carry operation authority into every asynchronous clipboard fallback and stop before starting another write after cancellation. Extend the lifecycle tests through the actual clipboard helper with a rejected delayed first strategy. **Transitional:** task-copy is still being implemented; verify its final version.

## 3. P2 — Full task-step copy ignores a page load's explicit non-commit result

**Location:** `packages/ui/src/components/tool-call/renderers/task-copy.ts:10–11, 75–76`.

The adapter discards the boolean returned by `loadOldestMessageWindow`/`loadNewerMessageWindow`. Those APIs return `false` when request authority is lost and no page was committed (see `session-api.ts:1557–1571, 1657, 1781, 1826–1828`). The copy operation's `isCurrent` only checks its own controller, child ID, activity and disposal, so a separate forced load/reconnect/mutation can invalidate the native page request without cancelling this copy operation.

If the resident child window is still `latest` when the oldest-page request returns false, `loadCompleteMessageHistory` visits that latest page and immediately declares the traversal complete. The resulting clipboard value silently contains only the resident tail, despite invoking the full-copy action specifically because earlier steps were omitted.

**Reproduction:** call the actual `useTaskStepCopy` with a still-active child, resident latest page, `loadOldest: async () => false`, and one resident tail step. These are the documented real loader outcomes under lost request authority. The hook completes successfully and writes the tail:

```text
oldest request committed=false; clipboard writes 1
[ '[\n{\n  "id": "only-latest-page-step",\n  "type": "tool"\n}\n]' ]
```

**Fix direction:** propagate the `Promise<boolean>` contract into the copy hook and abort/fail the traversal on `false` before visiting the resident page or writing the clipboard. Add oldest/newer non-commit coverage. This is a renderer/copy integration issue, independent of whether the loader correctly rejects stale commits. **Transitional:** task-copy is still being implemented; verify its final version.

## Checks and review coverage

- Read the full changed renderer paths: message-block segmentation/record projection; Markdown/global cache integration; lazy header/body input-output copy; ANSI/diff output; read/write/edit/patch/default/bash/webfetch/skill; apply_patch and diagnostic bounds; todo bounds; task summary/loading/activity/pagination/copy; inline and fallback permissions.
- The source-less fallback permission dead end from round 1 appears repaired: bounded sequential full-source pages, copy access, exact-request/diff review state, guarded approval, and stale-completion handling are present. No remaining fallback-permission finding identified in this review. The other implementer's browser fixture was left alone.
- **20 focused tests passed** with `node --conditions=browser --import tsx --test --test-force-exit` across `render-memory.test.ts`, `renderer-copy.test.ts`, `renderers/read.test.ts`, `renderers/edit.test.ts`, `permission-block.test.ts`, `renderers/task-copy.test.ts`, `permission-diff-review.test.ts`, and `global-cache.test.ts`.
- An initial run without `--test-force-exit` timed out due to retained test-process handles; the explicit-force-exit run completed with 20 pass / 0 fail.
- The three reproductions above used Node, real Solid client reactivity and the named real store/helper modules. No rendered-browser reproduction is claimed. No browser/Vite, daemon/desktop, heavy pruning, source edits, commits or pushes were performed.
- Memory/store/cursor budget review is owned by the other reviewer; this report does not duplicate that scope.

**Gate recommendation:** corrections required; do not report zero findings or certify the final head until these renderer/copy cases are addressed and independently rechecked.

pascalandr (commented):
# PR648 corrected candidate — independent memory/navigation Gatekeeper review, round 3

## Result

**ZERO FINDINGS in the reviewed memory ownership, eviction/budget, cursor/undo/navigation, and reducer-retirement scope.** Both round-2 P2 findings are corrected. No new actionable introduced/exposed fault was identified in this review.

This is a scope-specific review result; renderer/permission/copy work belongs to the separate reviewer.

## Exact candidate

- Worktree: `D:/CodeNomad/.codenomad/worktrees/pr648-review-fixes`.
- PR comparison base: `89db55d340f942c5523de656dd92aaec6e9f3978`.
- Review began at HEAD `467261f2` plus working retirement fixes. During review the parent committed those fixes as **`748c531998d0c3d089acd5100508ce68ebb4a51e`** (`fix(ui): drain terminal transcript work before reducer retirement`).
- The reviewed memory/navigation source and regression-test contents match that commit. Concurrent UI/copy/browser-fixture working changes were not included in this gate.
- Content fingerprints captured at completion (`git hash-object`):

| File | Git blob hash |
| --- | --- |
| `packages/ui/src/stores/opencode-data.ts` | `95d4aa627eb8ee0f9427b6e52e12a5a5a0bafc29` |
| `packages/ui/src/stores/opencode-data-idle.test.ts` | `5d916d3c63e6e534a5e99474c0663ec97671ef33` |
| `packages/ui/src/stores/session-request-authority.test.ts` | `56d9608a64ec726d26ce4b53f77e3992f51f74bc` |
| `packages/ui/src/stores/message-v2/instance-store.ts` | `17f0d4f4b6f21c111fa8815de4bc14219e0d61cf` |
| `packages/ui/src/stores/session-api.ts` | `61b290af9e86a33c6e76c0da7410a1df2bcf0851` |
| `packages/ui/src/stores/session-transcript-memory.ts` | `4876ce40b096dfd5546f73ea034aa39dd0e9a6d1` |
| `packages/ui/src/components/instance/shell/useSessionCache.ts` | `f68740d0eb15c0b764cc583531e3f936fef3ece9` |

## Independent inspection

Reassessed the complete memory/navigation candidate against the PR base, building on the previous independent source review rather than limiting this pass to the new tests:

- Solid-store payload deletion and sibling isolation, retained-size accounting, active/live/loading protection, visibility ownership, measurement cancellation, global/derived-cache invalidation, and pending-part bounds.
- Message-load epoch/client/instance/mutation/window fences; shared page-request consumer cancellation; retry lifetimes; empty boundary pages; saved ascending/descending cursors; staged-undo seeking and restoration.
- Every reducer retirement path: normal idle, queued rotation, scheduled/in-flight/retrying authoritative resync, content invalidation, new execution, session deletion, instance reconnect/destruction, and late async completion.

### Round-2 finding 1 — corrected

`retireWhenDrained` now records idle cleanup intent without cancelling outstanding work. `retireDrainedTranscript` requires the current transcript identity and no rotating/resyncing/queued/retrying work. Rotation finalization and successful authoritative resync reconsider retirement only after their projections. Recovery failures retain their retry path. A newer execution or appending event cancels the preceding retirement intent.

### Round-2 finding 2 — corrected

Ordinary per-session reducer disposal no longer deletes the message/mutation scalar fences. Their payload lifetime is now independent of request authority. Explicit deletion still clears them while session removal invalidates pending UI loads; whole-instance reset still advances instance generation. The original pre-cancellation/revert response cannot regain authority just because idle releases its reducer.

## Validation performed

**155 focused tests passed, zero failures**, using actual source modules with local test fixtures:

```powershell
node --conditions=browser --import tsx --test --test-force-exit packages/ui/src/lib/global-cache.test.ts packages/ui/src/lib/retained-size.test.ts packages/ui/src/lib/session-transcript-lru.test.ts packages/ui/src/stores/message-v2/transcript-eviction.test.ts packages/ui/src/components/instance/shell/useSessionCache.test.ts packages/ui/src/stores/message-v2/instance-store.test.ts packages/ui/src/stores/message-v2/message-window.test.ts packages/ui/src/stores/session-request-authority.test.ts packages/ui/src/stores/opencode-data.test.ts packages/ui/src/stores/opencode-data-idle.test.ts packages/ui/src/stores/opencode-data-settlement.test.ts packages/ui/src/stores/opencode-projection-isolation.test.ts packages/ui/src/stores/session-pruning-projection.test.ts packages/ui/src/stores/session-pruning-pagination.test.ts
```

The pruning-named tests in this command are lightweight UI fixtures, not native database/pruning runs.

### Additional independent adversarial probes

Executed three stdin probes through the real reducer/store modules, using fake SDK clients rather than the daemon. Each filled the bounded reducer, forced overflow recovery, held its first native-page response, delivered idle, and then varied the lifetime transition before releasing the obsolete response:

1. **Reconnect:** a replacement execution retained its reducer and text; the old deferred resync never published or retired the replacement. One mocked page read, zero stale resync callbacks.
2. **Deletion:** the old generation's held response could neither restore its obsolete message nor retire a subsequent replacement reducer. One mocked page read, zero stale resync callbacks.
3. **Resume without deletion/reconnect:** the new execution cancelled retirement while the old read was outstanding. The revision-conflicted first response was discarded; one trailing authoritative read published the resumed execution, which remained alive for subsequent text events. Two mocked reads, one current resync callback.

All three asserted that the obsolete message was absent and the replacement/resumed assistant accepted `new generation survives` as its text. All assertions passed.

## Limits

- No desktop/browser run, native-event timing soak, heap snapshot, shared-daemon request, or heavy native pruning test was performed.
- UI typecheck was reported by the implementer but was not independently rerun in this pass; it is not counted as reviewer validation here.
- Rendering, permission approval, clipboard behavior, and their in-progress plumbing were outside this review's ownership.
- No source files were edited and no commit/push/merge was performed by this reviewer. Only this temporary report was created.
- ZERO FINDINGS applies to the source fingerprints and scope above, not to subsequent changes or the separate UI review gate.

pascalandr:
## Corrections published — candidate `7a147c7c`

Four correction commits now follow the reviewed `5bc62d65` head:

- `1d3f85c0`: actual Solid-store payload/index deletion, stable active-session visibility ownership, correct saved-window undo-tail cursor orientation.
- `467261f2`: source-less permission diffs have bounded complete paging/copy access, request-scoped acknowledgement and approval recovery; all ten locales updated.
- `748c5319`: ordinary-idle retirement drains pending rotation/resync (including retries), preserving scalar request revisions across payload disposal.
- `7a147c7c`: mounted rows receive changed part-ID selections; task copies inherit parent activity, cancel pending reads, require committed pages, and fence every clipboard fallback. Task-section copy controls now receive real pointer clicks. Correction unit tests are registered in CI.

### Review loop

- Round 1: **1 P1 + 4 P2**, published before corrections.
- Round 2 memory/navigation: **2 P2**, both reproduced and fixed.
- Round 2 renderers: **3 P2**, all reproduced and fixed in the candidate; final independent renderer recheck is running.
- Round 3 memory/navigation: **ZERO FINDINGS** on `748c5319`, with 155 focused tests and additional independent reconnect/deletion/resumed-execution probes. This scoped result is already published; it is not a claim that the renderer gate has passed.

### Executed validation

- Earlier integrated correction candidate: 428 pure UI tests and 341 UI integration tests passed, plus production UI build.
- Idle correction: 111 focused tests passed; independent reviewer subsequently passed 155 in its broader scope.
- Final task-copy implementation: 14 lifecycle/non-commit/actual-clipboard-fallback tests, four standalone clipboard tests, and six real-component browser cases passed.
- Fallback permissions: three real-modal browser cases passed, including clipboard failure, full sequential paging, request changes and dismissal.
- Session rendering: 43 existing browser cases passed before the final live-part fix; its added mounted native-event append regression and three targeted tool-scroll/reprojection neighbors then passed.
- All three tool-image browser cases passed across two targeted invocations.
- UI typecheck passed on committed `7a147c7c`; `git diff --check` passed before commit.

Early failed browser attempts are not counted as success: permission fixture locale/serialized-helper issues were corrected, and task pointer clicks exposed a real styling defect which was fixed and retested with normal clicks. An accidentally overbroad local Node discovery timed out and is not a passing full-suite result; explicit UI lists replaced it. No renderer heap/24-hour soak claim is made.

The old detached worktree's seven local changes remain preserved separately. GitHub checks for the final head and the final independent renderer review are pending. Administrator merge is authorized by the user only after these gates are clear.

### Existing file-size signals

Touched oversized files: `message-block.tsx` ~1,965 lines; `message-section.tsx` ~1,566; `tool-call.tsx` ~1,090; `tool-call/renderers/task.tsx` ~636; `message-v2/instance-store.ts` ~1,877; `opencode-data.ts` ~757; `session-api.ts` ~1,962; `session-request-authority.test.ts` ~2,247. These are follow-up refactoring signals, not additional unverified review findings.

pascalandr (commented):
# PR648 independent renderer review — round 3

## Result: 1 actionable P2 finding

Review began at `748c5319` plus the final working UI changes. Those exact renderer/copy repairs were committed during review as **`7a147c7c465c705083aef13325df2116f587be0e`** (`fix(ui): preserve live content and fence complete task copies`). Final working tree was clean. I rechecked the remaining finding against that committed head. Full-PR comparison base remains `89db55d340f942c5523de656dd92aaec6e9f3978`.

## P2 — Several new full-source copy controls remain permanently hidden and unclickable on desktop

**Primary location:** `packages/ui/src/components/tool-call/diagnostics-section.tsx:7–16`, specifically the class at line 13.

**Same defect at:**

- `packages/ui/src/components/message-block.tsx:1077–1092` (class at **1082**): full message-parts copy after the 200-part display cap.
- `packages/ui/src/components/tool-call.tsx:403–411` (class at **408**): full tool error copy after the 10,000-character cap.
- `packages/ui/src/components/tool-call/renderers/todo.tsx:94–100` (class at **97**): full todo overflow copy.

These newly introduced body/footer actions still carry `tool-call-header-copy`. On a normal mouse desktop, the existing stylesheet explicitly sets every such element to `opacity: 0; pointer-events: none` (`packages/ui/src/styles/messaging/tool-call.css:223–228`). It restores visibility and pointer events only for a **direct child** of a hovered/focused `.tool-call-header` or `.message-technical-group-header` (lines **230–233**). None of these four buttons has such a parent: they live in a diagnostic row, truncation notice, or error body. Hovering the surrounding card/header therefore never restores them.

The task renderer's final correction correctly removed this class in favor of `tool-call-io-copy`, but the equivalent newly added access controls elsewhere were missed. There is no overriding rule for these body actions in the stylesheet; `.tool-call-io-copy` itself only sets a margin, so it must replace, rather than merely accompany, the header-only class.

**User-visible reproduction:**

1. On a fine-pointer/hover desktop, render an edit/apply_patch result with more than 100 diagnostics, or one diagnostic message longer than 2,000 characters.
2. Expand diagnostics. The bounded preview and truncation notice are visible, but the new full-diagnostics copy button is invisible and cannot receive a mouse click. The tool-header output copy contains the patch/output, not this diagnostic payload, so it does not recover omitted diagnostic content.
3. Equivalent loss of the intended full-source access occurs for a message with more than 200 parts and a tool error longer than 10,000 characters. Todo's header copy is an alternate route, but its overflow copy button is still broken by the same class.

**Evidence level:** direct JSX ancestry and complete matching stylesheet analysis, not a claimed browser reproduction. No browser was launched, as requested. These particular body/footer controls were added by this PR; this finding does not treat the existing generic IO-header styling as a new blocker.

**Correction:** apply the same non-header copy class used by the corrected task buttons to these four new body/footer actions, or otherwise use an appropriate shared body-action primitive. Validate at least the full-diagnostics and full-message/error controls with the real stylesheet and pointer clicks, checking untruncated clipboard payloads. Grouped as one finding because all four occurrences have the same root cause and fix.

## Round-2 findings: independently rechecked and resolved

1. **Appended content parts:** `flushContent` now publishes a fresh content item when selected part IDs change. I repeated the independent real-store/Solid `indexArray` probe: an appended second text part now reaches the mounted memo. I also grew the message to 203 parts and verified the mounted selection contains exactly 200 parts, preserves the first 100, and updates its tail through `p202`. This probe does not substitute for the parent's real rendered session test.
2. **Stale clipboard fallbacks:** task-copy now passes both `signal` and `isCurrent` to the actual clipboard helpers; checks cover chunk-write rejection, text-write rejection, and synchronous focus/selection reentrancy before `execCommand`. Actual-helper tests pass, including preservation of a newer copy after cancellation.
3. **Uncommitted copy pages:** the hook now requires `Promise<boolean>` and rejects `false` before visiting/copying the old resident data. Both oldest and newer non-commit cases pass, with no partial clipboard write.

The source-less fallback permission recovery remains present and coherent: fixed-size sequential source pages, copy access, exact instance/session/request/diff review identity, guarded approval, and late-completion fencing. No residual permission finding identified.

## Review scope and validation

- Rechecked the final changes and their integration through `MessageSection` → `MessageBlock`/exploration groups → `ToolCall` → nested task renderers; activity propagation, visible-child registration, disposal and copy lifetime.
- Revisited full-PR renderer/display behavior: message segmentation and bounded head/tail projection; Markdown/cache integration; lazy complete input/output copy; ANSI and diff bounds; read/write/edit/patch/default/bash/webfetch/skill output; apply_patch/diagnostic access; todo and task-summary bounds; inline and fallback permissions. Inspected actual stylesheet consequences for newly introduced copy actions, which exposed the finding above.
- **52 focused tests independently passed**, 0 failed, via `node --conditions=browser --import tsx --test --test-force-exit`, across:
  - `components/tool-call/render-memory.test.ts`
  - `components/tool-call/renderer-copy.test.ts`
  - `components/tool-call/renderers/read.test.ts`
  - `components/tool-call/renderers/edit.test.ts`
  - `components/tool-call/permission-block.test.ts`
  - `components/tool-call/renderers/task-copy.test.ts`
  - `components/permission-diff-review.test.ts`
  - `components/message-history-pagination.test.ts`
  - `lib/clipboard.test.ts`
  - `lib/global-cache.test.ts`
- Independent append/tail probe passed using real `createInstanceMessageStore`, `buildRecordDisplayData`, and Solid client-reactivity primitives.
- Parent/implementer-reported browser and typecheck results were not rerun or represented as independently obtained. No browser/Vite, daemon/desktop, source edits, commits, or pushes were performed by this reviewer.
- Memory/store/navigation budget review remains the separate reviewer's scope; its reported zero-findings result is not duplicated or independently certified here.

**Gate recommendation:** correct the remaining non-header full-copy actions, then recheck their rendered accessibility and payloads. This report is **not ZERO FINDINGS**.

pascalandr (commented):
# PR648 renderer source gate — round 4

## Result: four reported controls fixed; one related P2 full-input access gap remains

Reviewed `7a147c7c465c705083aef13325df2116f587be0e` plus the four working class replacements in `message-block.tsx`, `tool-call.tsx`, `tool-call/diagnostics-section.tsx`, and `tool-call/renderers/todo.tsx`. This is a source/CSS gate, with the parent's separate real-stylesheet pointer regressions still pending. Full-PR baseline: `89db55d340f942c5523de656dd92aaec6e9f3978`.

## Verification of round-3 correction

All four flagged body/footer controls now use `tool-call-header-icon-button tool-call-io-copy`, with `tool-call-header-copy` removed:

- `message-block.tsx:1082` — complete message parts.
- `tool-call/diagnostics-section.tsx:13` — complete diagnostic payload.
- `tool-call.tsx:408` — complete long error.
- `tool-call/renderers/todo.tsx:97` — complete todo list.

They no longer match the desktop hiding/pointer-disabling selector in `styles/messaging/tool-call.css:223–228`. Their complete-source callbacks are unchanged. The matching class in the task renderer's three full-access buttons remains corrected too.

## P2 — Oversized generic tool input now depends on a still-hidden IO copy button

**Location:** `packages/ui/src/components/tool-call.tsx:476–480` (class at **479**), in conjunction with the PR's bounded input path at **284–292** and `packages/ui/src/components/tool-call/utils.ts:182–187`.

The generic `renderIoHeader` copy button still has:

```tsx
class="tool-call-header-icon-button tool-call-header-copy tool-call-io-copy"
```

Its parent is `.tool-call-io-header`, not `.tool-call-header`. The exact same desktop rule therefore sets `opacity: 0; pointer-events: none`, and the direct-header-child hover/focus recovery does not match. Merely including `tool-call-io-copy` does not override that rule: the class only sets a margin.

**Why this is actionable in this PR, despite the inherited class:** previously the input body rendered the complete YAML, so the user could inspect/select all input even with the existing copy-button styling defect. This PR changes the body to `formatToolInputForRender`, which replaces large structured input with a short omission message. It moves complete input access into `copyToolInput` (lines 443–445), reachable through this hidden button. The outer tool-header copy and overflow menu copy copy **output**, not input. For a custom tool with large input and short output, the new rendering cap consequently removes the prior full-input access while the intended recovery action cannot be clicked. This is the new dependency on the inherited defect, not a request to block on unrelated legacy styling.

**Reproduction scenario:** on a fine-pointer/hover desktop, render a completed custom/default tool with input `{ payload: "x".repeat(20000) + "INPUT_TAIL_SENTINEL" }` and output `"ok"`; expand Input. The body shows only `"Structured output omitted from rendering because it is too large."`. The input-header copy button is invisible/non-clickable. The visible outer copy returns `ok`, not the input. The extra Markdown full-source button cannot recover it because the Markdown component receives only the short placeholder.

**Independent evidence:** a small Node probe using the real `formatToolInputForRender`, `formatToolInputForCopy`, and baseline YAML formatter verified:

```json
{
  "rendered": "\"Structured output omitted from rendering because it is too large.\"",
  "fullCopyLength": 20038,
  "baselineRenderedFullInput": true
}
```

Assertions confirmed the sentinel is absent from the rendered value and present in both full-copy text and the baseline full YAML. Button ancestry and CSS selector behavior were checked directly in source; no browser reproduction is claimed.

**Correction:** remove `tool-call-header-copy` from the shared IO-header copy button as well, retaining `tool-call-header-icon-button tool-call-io-copy`. Add an oversized-input pointer-copy case to the in-progress stylesheet regressions and check the sentinel in the copied JSON.

My round-3 report noted this inherited IO-header class but did not identify the newly introduced loss of the full rendered input. This round's complete-access sweep establishes that direct PR dependency, so it is now reported precisely as the remaining input-access regression.

## Sweep and limits

- Searched all renderer component uses of the header-only class, IO-copy class, Markdown source-copy control, and full-diff access callbacks.
- Checked Markdown source-copy against `.message-action-button` styling; it does not use the header-only hiding class.
- Checked shared diff/apply_patch full-copy and fallback-permission copy/page controls: they use `.file-viewer-toolbar-icon-button`, with appropriate toolbar styling and no matching header-only hiding rule.
- Checked the actual tool-header controls and responsive overflow access; their remaining header-only classes are in the intended header ancestry.
- Checked unchanged full-source callbacks for the four fixes and task summary/title/full-history controls. No additional actionable source finding from this focused sweep.
- Round-2 fixes and previously passing 52-test/append-probe evidence were not rerun or re-certified. No broad tests were run in this round.
- The new narrow input probe produced all asserted results; its first invocation retained imported process handles and timed out after output. Repeating with explicit successful process exit completed normally. No browser/Vite, daemon/desktop, source edits, commits, or pushes were performed.
- Rendered pointer verification for the four corrected controls belongs to the parent's concurrent agent and remains outside this report's independently observed evidence.

**Gate:** not ZERO FINDINGS yet. The four requested fixes are correct at source level; correct and render-test the generic input access control before closing this renderer gate.

pascalandr (commented):
# PR648 independent renderer gate — round 5

## ZERO FINDINGS — source review complete for the reviewed renderer changes

**Reviewed snapshot:** HEAD `7a147c7c465c705083aef13325df2116f587be0e` plus the five working copy-control class corrections across four source files. Full-PR comparison base remains `89db55d340f942c5523de656dd92aaec6e9f3978`. At inspection, the new `tests/browser/full-source-copy.test.ts` and `fixtures/full-source-copy.tsx` were untracked work belonging to the browser agent; this report does not certify their completion or execution.

## Final residual fix verified

`packages/ui/src/components/tool-call.tsx:479` now uses exactly:

```tsx
class="tool-call-header-icon-button tool-call-io-copy"
```

The generic IO-header button no longer matches `.tool-call-header-copy`'s desktop `opacity: 0; pointer-events: none` rule. Its parent `.tool-call-io-header` is therefore no longer dependent on an inapplicable direct-tool-header hover/focus recovery rule.

The input callback remains correctly wired through `onCopy: copyToolInput` (line 572), which obtains `formatToolInputForCopy(props.toolInput())` at click time (lines 443–445). It copies the original input JSON rather than the bounded display placeholder. The shared output IO action also retains its lazy full-output callback. The correction changes only the button classes.

## Control sweep and prior findings

- Rechecked all remaining `tool-call-header-copy` matches in component sources. They are the actual direct tool-header copy, speech and action buttons, plus the header overflow trigger. Checked the enclosing JSX and speech/overflow helper use; no remaining non-header IO/body action carries the hiding class.
- Confirmed the previous four corrections remain present: complete message-parts access, complete diagnostics access, long-error access, and todo overflow access all use `tool-call-header-icon-button tool-call-io-copy`.
- Task full-history, legacy-summary and long-title copy actions retain the corrected non-header class.
- The Markdown source-copy and diff/apply_patch/fallback-permission controls retain their appropriate separate button primitives. No newly identified access gap from this final source sweep.
- All round-2 findings remain resolved in the reviewed code: reactive publication of changed content selections, clipboard fallback operation authority, and rejection of uncommitted full-copy pages.
- The fallback-permission recovery and full-diff review identity corrections remain covered by the preceding independent source review. No residual actionable renderer finding remains from rounds 2–4.

## Scope and evidence limits

This final gate closes the residual source findings from the cumulative renderer/display review: bounded message/Markdown/tool views, lazy lossless copy paths, diagnostics/todo/apply_patch access, permissions, and task activity/copy lifecycle. The independent memory/store/navigation review remains separate.

**New work in this round:** source diff inspection, complete remaining class-use search, callback/DOM-ancestry verification, and matching CSS-rule inspection. No new concern justified broad test repetition, so no tests or browser were run in this round.

**Previously obtained independent evidence, not rerun here:** 52 focused tests passed in round 3; real-store/Solid probes confirmed live content append and the 203-part bounded head/tail projection; the round-4 narrow input probe confirmed full JSON retention independently of the render placeholder. These prior results are evidence for their reviewed snapshots, not new rendered validation of the current class changes.

The parent must independently await successful **real-stylesheet pointer/clipboard regressions for all five affected access paths, including the 20,000-character exact-JSON input case**, and required CI on the final committed head before merge. This source-only ZERO FINDINGS result does not assert that those pending checks have passed.

No browser/Vite, daemon/desktop operations, source edits, commits, or pushes were performed by this reviewer. Only this temporary report was written.

pascalandr:
## Independent review loop complete — ZERO FINDINGS; final-head CI pending

Final candidate: **`d086181fe8b36767ddd1a881b0bfdb3b2f40b517`**, pushed to this PR. Correction worktree is clean.

- Memory/navigation: round 3 **ZERO FINDINGS** at `748c5319`, with 155 focused tests plus independent reconnect/deletion/resumed-execution probes. Later commits affect renderers, clipboard and their tests, not that reviewed store/navigation implementation.
- Renderers/display/copy: round 5 **ZERO FINDINGS** after resolving the round-2 content-reactivity/clipboard/page-commit findings and the round-3/4 desktop pointer-access findings. Round 5 reviewed `7a147c7c` plus the exact five class replacements now committed in `d086181f`; its report explicitly does not claim browser execution.
- The separate browser verification has now passed **6/6** real-stylesheet desktop cases with normal pointer clicks and actual Chromium clipboard readback: 137 diagnostics, a long diagnostic, 237 message parts, a 16k+ error, 237 todos, and a 25k+ tool input. Complete original content is copied while the DOM stays bounded; input and output copy routes are independently verified.
- Earlier correction validations are recorded in [the published ledger](https://github.com/NeuralNomadsAI/CodeNomad/pull/648#issuecomment-5821619457), including task lifetime/fallback tests, real task/permission browser fixtures, live native-part append, image output, UI typecheck and build.

All identified findings are closed in the final candidate. The source-review gate is clear. Final-head GitHub CI is still pending and is being watched; the user-authorized administrator merge will occur only after those checks pass and the exact head is revalidated.

No application reinstall/replacement, shared-daemon stop or local heavy native pruning UI fixture was used for the corrections. The old worktree's uncommitted edits remain preserved separately.

Issue #319, opened 2026-04-12
Feature Request: Conversation Categorization by Working Tree & Open File Explorer Support


youfch:
OpenCode Desktop has a working tree with dedicated conversations for each tree. Can conversations be categorized by working tree?
Add the ability to open the system file explorer.

shantur:
Thanks for using CodeNomad @youfch 

In CodeNomad you can assign a worktree to any session, and change worktree to another. We see worktree in the session list.
Are you looking for a session by worktree grouping?

pascalandr:
These features are in 0.20.0 dev builds

PR #649, opened 2026-08-12
feat(v2): manage session families across worktrees

pascalandr:
## Summary

Manage complete session families across the opened local repository's native OpenCode worktrees. This supersedes #642 with the V2 implementation and addresses #319.

- OpenCode owns `worktree.refresh/list/create/remove`; CodeNomad supplies the default directory, named-branch policy, local repository scope and verified family transactions.
- A project is the opened local checkout and its registered worktrees, identified through the physical Git common directory. Independent clones sharing an OpenCode project ID stay separate.
- Session rows, composer selection, search and the Git panel follow native `SessionInfo.location.directory`. Durable metadata is preserved, but old `worktreeSlug` metadata does not override native placement.
- Creation uses the selected checkout's revision and the main local checkout's `.codenomad/worktrees` parent. Uncommitted edits stay in the source checkout. External registered worktrees remain discoverable.
- Keep copy/open/delete actions in the accessible selector, remove the redundant action bar, hide root-workspace badges, and explain “Create and use” in all ten locales.

## Current base and native contract

This branch includes merged `dev@e47e01c6`, including #697's separately reviewed discovery/status compatibility fix. UI/server pin `@opencode/client@2.0.4`; the independently managed daemon has no exact-version startup gate.

- Complete bounded inventories use opaque native cursors; continuations send only the cursor.
- Moves are serialized across complete families, block active sessions, verify authoritative native locations and roll back partial failures.
- Native move events refresh placement and the catalogue without speculative cross-project transfers.
- Opaque worktree identifiers are separate from mutable branch labels. The main checkout cannot be removed.
- Native detached creation is followed by ordinary Git branch selection, preserving Git's refusal to reuse a checked-out branch; branch-attachment failure cleans up through the native API.
- Agent placement instructions distinguish creating a checkout, one-off commands and moving the conversation's execution context. Explicit attachment choices are preserved.

## Deletion safety

- Physical checkout reservations cover published and in-flight logical workspaces.
- Project serialization and deletion fences cover all family source/destination checkouts.
- Complete inventories and cross-project checks run inside the deletion fence, including nested workspaces and owned Shell/PTY resources.
- Canonical path, filesystem identity, registration, branch, HEAD and active sessions are rechecked before removal. Replacement checkouts fail closed; rollback requires the original identity.

## Recovery and Git-panel corrections

Validation on the real repository exposed serial enumeration of 162 worktrees and concurrent ownership misses launching overlapping inventories. Git annotations now run with bounded concurrency, pending catalogue reads are shared, and negative ownership results are snapshot-cached. A regression exercises 40 concurrent misses.

The Git panel now treats its complete local staged/unstaged inventory as authoritative instead of resurrecting native-only stale rows. It also rechecks the selected worktree/request after waiting for native VCS status.

## Validation

Latest integration validation:

- Isolated official **OpenCode 2.0.0 and 2.0.7**: production discovery/status, native worktree discovery/create/remove, selected HEAD/default parent/named branches, independent-clone scope, stable IDs, nested paths, checked-out/dirty guards, complete family moves and rollback.
- Real Solid selector browser tests: mouse, keyboard, touch, actions without session movement, creation source and returned stable ID.
- Targeted server ownership/cache/manager/proxy/route/location tests; UI session projection/request-authority/lifecycle tests; family evacuation/rollback and Git model regressions.
- UI/server typechecks and production server build including the UI.
- Installed Windows Tauri application on the actual repository: dedicated session creation/use, native location and row/composer badges, Git marker visibility, move back, and agent `session_move` reflected in both badges.
- Rebuilt UI/server resources installed; full Developer Mode restart restored the original conversation. Disposable sessions, marker, worktrees and branches cleaned up. The original conversation remained attached to `pr649-final`.
- `git diff --check`.

An immediate post-restart desktop run exceeded a 30-second refresh-button wait; the subsequent completed run passed. These checks validate the workflows, not a startup latency bound or exhaustive compatibility of every OpenCode payload.

Earlier branch validation also covered Electron/Tauri native actions, packaging and desktop parity; those results remain in the PR's historical validation comments. Detailed contract and latest evidence: `dev-docs/WORKTREE_SESSION_PLACEMENT.md`.

## Residual limit

OpenCode and Git expose no shared snapshot/CAS transaction. CodeNomad narrows races with reservations, locks, fences and repeated authoritative checks; unrelated processes can still mutate state outside those controls.


pascalandr:
## Refresh against dev aa51cbb9

Published head: `bedc8bae063547c7b8ef4b4224a861aa7d29def7`

merge(dev): adapt worktree families to shared V2 location compatibility

Integrate dev aa51cbb9 and preserve #695's guarded proxy, legacy location
identity and exact destination validation. Keep project-wide inventory and
family transactions, worktree mutation/deletion reservations and #687's union
of migrated global descendants. Do not restore permissive opaque cursor
forwarding or obsolete workspace query parameters at the proxy boundary.

Route forward moves and rollbacks through moveSessionToLocation, carry full
legacy identities through the internal location context, and use the modern
project-scoped worktree refresh contract with narrow legacy translation.
Reject same-project destination redirection before mutations. Adapt shell/PTY
blocker checks, retain both workspace-manager suites and fix the RPC event
fixture's native owner identity and timeout cleanup.

Add a synthetic native family fixture to the existing private-daemon location
test: verify complete-family moves, refresh, evacuation failure and exact
rollback against official 2.0.3 and 2.0.5, without touching user storage or the
shared daemon. Server suite: 523 passed/2 skipped, followed by the final 20
family regressions including the added destination refusal. UI stores: 449
passed. Electron native: 194 passed/4 skipped on rerun (initial Windows
process-identity timing assertion failed). Server/UI/Electron typechecks pass.


Validation above is local execution evidence. CI on this published head and the explicitly untested interactive/platform scenarios remain separate. No PR was merged.

Refactor signals: `packages/server/src/workspaces/manager.ts` ~1079 lines; `packages/ui/src/stores/session-request-authority.test.ts` ~1922. The inherited large modules were not split during conflict resolution.

pascalandr:
### User-chosen conversation placement — cad0ae33

Included in this PR because native worktree movement and CodeNomad conversation organization need one explicit user-intent contract.

- Add native `codenomad.session-placement` context before prompt, slash-command and session-shell admission. Creating or working in a checkout is not permission to move the conversation; explicit user reattachment remains supported.
- Reapply for existing sessions on their next submission and new sessions on their first submission, alongside voice-mode context. Preserve other named entries, propagate synchronization failures, and avoid pinning an initial directory or caching success only in the browser.
- Keep native location/event authority and the existing transactional UI family moves.

The upstream move recommendation is injected by the **core runtime** (`opencode.tools`), not only the TUI. The models can coexist: a tool working directory is distinct from session placement, while location-scoped services continue to follow the native session. This instruction is model guidance, not an agent-vs-user runtime permission or a replacement for family transactions.

Analysis and source references: [`dev-docs/WORKTREE_SESSION_PLACEMENT.md`](https://github.com/pascalandr/CodeNomad/blob/cad0ae33/dev-docs/WORKTREE_SESSION_PLACEMENT.md). It also records a separately observed native Windows directory-filter discrepancy (`D:\codenomad` vs `D:\CodeNomad`); this increment does not claim to fix runtime path canonicalization.

**Validation:** 139 focused UI tests passed, including delayed setup failures, voice-mode coexistence, existing-session repair, request authority, native move reconciliation, restoration and worktree families. UI typecheck, production UI build and `git diff --check` passed. Desktop visual verification was unavailable after the live session moved because Developer Mode no longer resolved the visible session.

pascalandr:
Follow-up in 3921ba08: the worktree filter now matches the displayed conversation's own native directory, consistently with its badge. Descendant locations/activity do not override the parent's placement; text-search ancestry and complete families under matching roots are preserved.

Added mixed-location/activity regressions (14 session-tree tests pass), with UI typecheck/build passing. The reported installed-app case was checked: hidden under the root-folder filter, visible under its own worktree and All worktrees, without closing the conversation.

pascalandr:
Follow-up in 56a1c4c1: search/filter mode now uses flat, independent session results, superseding the root-only worktree filter in 3921ba08. A localized **Show subsessions** switch (off by default) lets matching children appear without their parents. Text/worktree filters, activity/name/worktree sorting, badges and bulk selection use each result itself. Closing search restores normal hierarchy. Native search results no longer require ancestor hydration to render.

Validation: UI typecheck/build; 14 session-tree and 65 request-authority tests; real-component browser coverage for the toggle, sorts, combined filters, independent badges/selection, orphan results and hierarchy restoration. Rebuilt UI loaded in the installed Windows Tauri renderer: an existing child in D:\CodeNomad appeared alone under Workspace while its parent remained in pr649-final; toggling it off or filtering the parent checkout hid it, and closing search restored the hierarchy without changing session locations. A first 30-second search wait expired; the repeat completed within the extended 90-second bound. Search still traverses the local worktree catalogue, so this does not establish a latency guarantee.

No merge performed.

pascalandr:
This follow-up landed on the source branch after #649 had already been squash-merged. It is now proposed separately in #698, based directly on current dev, with the exact tested final tree.

PR #654, opened 2026-08-17
docs: fix broken star history chart

OctoBored:
The Star History chart was broken because it depended on the GitHub stargazer API. Switch it to an alternative provider using the same data source so the community chart renders correctly. No API token required.


pascalandr (approved):
This new endpoint returns the actual chart for this repository. Thanks @OctoBored !

PR #664, opened 2026-08-27
fix(usage): parse credit-based quota limits for z.ai coding plans

Kondziosz:
## User-visible behavior

The provider usage panel for z.ai coding plans (`zai-coding-plan`) currently shows **"Usage is temporarily unavailable."** Since around the GLM-5.3 release, the panel stays broken permanently — not intermittently.

This patch makes credit-style quota entries render again as normal usage windows (percentage bar + reset time), so users see e.g. a `5h` window at ~9% and the plan-level window at ~1% instead of an error placeholder.

## Root cause

The quota endpoint (`https://api.z.ai/api/monitor/usage/quota/limit`) still answers `200` quickly, but z.ai migrated coding-plan quotas from token/time windows to credit-based limits with the GLM-5.3 rollout. `data.limits[]` entries are now typed `CREDIT_LIMIT`:

```json
{"type":"CREDIT_LIMIT","unit":3,"number":5,"usage":2000,"currentValue":157,"remaining":1842,"percentage":7,"nextResetTime":1787849891189}
```

`createTokenLimitProvider` only accepted `TOKENS_LIMIT` / `TIME_LIMIT`, dropped every entry via its type filter, and returned an empty windows map — which the UI surfaces as "temporarily unavailable".

## Implementation

Extend the shared factory to parse both generations instead of assuming one shape:

- `TOKENS_LIMIT` / `TIME_LIMIT` handling is byte-for-byte unchanged.
- Any other limit type with a usable `percentage` or `remaining`/`usage` pair produces a window: `usedPercent` prefers the server-computed percentage and falls back to `(usage - remaining) / usage`; `valueLabel` is intentionally not set so the UI renders its native percent display.
- Only hour-based windows map to explicit `windowSeconds` (unit 3, matching the legacy HOUR handling). Unknown unit codes fall back to the generic "usage" label rather than guessing window lengths, and entries without any usable numbers are skipped so unrelated limit kinds can't produce garbage rows.
- `zhipuai-coding-plan` (bigmodel.cn) shares this factory and inherits dual-format support for free if/when Zhipu migrates that endpoint too.

## Validation

- `tsc --noEmit` clean for the server package.
- Compiled provider exercised end-to-end against the live endpoint returning `CREDIT_LIMIT` entries; parser output:
  ```json
  {"windows":{"5h":{"usedPercent":10,"windowSeconds":18000},"usage":{"usedPercent":2}}}
  ```
- Verified in CodeNomad against a real lite-tier coding plan: panel renders both windows with percentages and reset timestamps where the previous code showed the unavailable fallback.

Happy to adjust labeling/UX (e.g., dedicated label for unknown-unit windows) if maintainers prefer a different trade-off.

### DIFF
--- a/packages/server/src/usage/providers/api-key.test.ts
+++ b/packages/server/src/usage/providers/api-key.test.ts
@@ -77,8 +77,9 @@ test("parses credit-based limits into usage windows", async () => {
   assert.equal(windows["5h"]?.usedPercent, 7)
   assert.equal(windows["5h"]?.windowSeconds, 18_000)
   assert.equal(windows["5h"]?.resetAt, CREDIT_RESET_AT)
-  assert.equal(windows["usage"]?.usedPercent, 1)
-  assert.equal("valueLabel" in (windows["usage"] ?? {}), false)
+  assert.equal(windows["weekly"]?.usedPercent, 1)
+  assert.equal(windows["weekly"]?.windowSeconds, 604_800)
+  assert.equal("valueLabel" in (windows["weekly"] ?? {}), false)
 })
 
 test("derives usedPercent when the credit payload omits percentage", async () => {
@@ -89,18 +90,20 @@ test("derives usedPercent when the credit payload omits percentage", async () =>
     () => zai.fetchQuota(),
   )
 
-  const window = result.usage!.windows["usage"]
+  const window = result.usage!.windows["weekly"]
   // (2000 - 1842) / 2000
   assert.ok(Math.abs((window?.usedPercent ?? 0) - 7.9) < 0.001)
+  assert.equal(window?.windowSeconds, 604_800)
 })
 
-test("coerces string quota fields and skips unusable entries without throwing", async () => {
+test("coerces string quota fields and skips unusable or unknown entries without throwing", async () => {
   const { result } = await withFetchStub(
     {
       data: {
         limits: [
           null,
           { type: "UNKNOWN_LIMIT" },
+          { type: "PROMO_LIMIT", percentage: 55, remaining: 10, usage: 20 },
           { type: "CREDIT_LIMIT", unit: 6, number: 1, usage: "10000", remaining: "9800", percentage: "2", nextResetTime: TOKENS_RESET_AT },
         ],
       },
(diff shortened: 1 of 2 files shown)


Kondziosz:
CI status note after investigating the two failing checks:

**`tests-tauri-windows`**: the single failure is `client_state::cross_host::tests::electron_and_tauri_simultaneous_startup_with_legacy_detection_elects_one` (`87 passed; 1 failed`), reproduced identically on both runs of this PR. This branch contains **zero** changes outside `packages/server/src/usage/providers/api-key.ts` (TypeScript quota parser), so no Rust/Tauri surface is touched. For reference, #659 (`add-turkish-localization`, which landed as dev tip `14586cb3`) also recorded a failed validation run before merge.

**`tests`**: failed in the second run only (`instance-client.test.ts` — "applies the loopback timeout and aborts a stuck instance", `failureType: cancelledByParent`, dangling-promise cancellation cascade) while passing the first run on byte-identical parser code — a runner-timing flake unrelated to the change.

Given both failures sit in subsystems unreachable from this diff, they look pre-existing/upstream-infra rather than regressions from this PR. Re-running the failed jobs in case the races resolve. Happy to help dig into the Tauri cross-host startup test separately if it's newly persistent on dev.

(Also pushed `31d45316`: null-guard `limit?.number` before field access, from review feedback.)

pascalandr (changes requested):
@Kondziosz  this fix is needed (and should port cleanly to `DEV-v2`), but two parser details need correction before merge.

- `packages/server/src/usage/providers/api-key.ts:135-139` treats `unit: 6, number: 1` as unknown and emits the second z.ai bucket as generic `usage`. This is the known weekly credit window (the fixture's Lite allowance is 10,000); OpenChamber maps unit 6 to 7 days, and z.ai documents 5-hour + weekly limits. Please map it to `weekly`/`604800` seconds and update the test to assert that window.
- `packages/server/src/usage/providers/api-key.ts:130-140` accepts every non-legacy limit type with numeric-looking fields. An unrelated future type can therefore be rendered as credit usage or overwrite another derived window. Please restrict this branch to `CREDIT_LIMIT` and add an unknown type with a percentage to the skip test.

The focused tests (11/11), server typecheck, legacy Zhipu path, and UI output otherwise check out. The Windows failure is the existing unrelated cross-host concurrency test.

Kondziosz:
Both review points addressed in 1f52e0c5: unit 6 now maps to the documented weekly window (604800s, corroborated via OpenChamber's mapping), and the credit branch is restricted to `CREDIT_LIMIT` entries — the skip fixture now includes a percentage-bearing unknown type to lock the allowlist in. Fixtures updated to assert the weekly window; full suite green.

pascalandr (dismissed):
Both requested changes are correctly addressed: `CREDIT_LIMIT` is explicitly allowlisted, and unit 6 now maps to the tested weekly window. Rechecked the final head: 11 focused tests pass, server typecheck passes, and the diff remains limited to the provider and its tests.

pascalandr (approved):
thanks @Kondziosz !

PR #671, opened 2026-09-04
test(server): isolate automation registry on Linux

pascalandr:
## Summary

- isolate the Developer Mode automation registry test on Linux as well as Windows
- derive the seeded registry path through the production resolver
- restore `LOCALAPPDATA`, `XDG_RUNTIME_DIR`, and `WSL_DISTRO_NAME` after the test

## Root cause

Both final CI runs for #647 failed in `prunes stale registry pressure before limiting discovery`. The test redirected only `LOCALAPPDATA`, which Linux ignores, then wrote fixtures into a hard-coded Windows registry path that did not exist.

## Validation

- `node --import tsx --test packages/server/src/opencode/automation-plugin.test.ts`
- 20 repeated targeted runs
- `git diff --check`

This fixes the deterministic Linux failure from PR Build Validation runs 33862538051 and 33863984160.

### DIFF
--- a/packages/server/src/opencode/automation-plugin.test.ts
+++ b/packages/server/src/opencode/automation-plugin.test.ts
@@ -229,7 +229,11 @@ test("pins parallel sessions to their independently inspected bridges", async ()
 test("prunes stale registry pressure before limiting discovery", async () => {
   const root = await mkdtemp(path.join(os.tmpdir(), "codenomad-automation-stale-"))
   const previousLocalAppData = process.env.LOCALAPPDATA
+  const previousXdgRuntimeDir = process.env.XDG_RUNTIME_DIR
+  const previousWslDistroName = process.env.WSL_DISTRO_NAME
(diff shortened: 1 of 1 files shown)


pascalandr (commented):
## Gatekeeper review of `fc206729`

No blocking code findings remain.

The change isolates the failing stale-registry case across Windows, native Linux, and WSL by pinning both `LOCALAPPDATA` and `XDG_RUNTIME_DIR`, clearing `WSL_DISTRO_NAME`, deriving the production-selected registry directory, and restoring every environment variable in `finally`.

### Validation

- PR Build Validation is green on Linux and Windows, including the complete server test suite and the Tauri crate.
- PR artifact builds are green on every required desktop target.
- The branch contains one focused test-only change and `git diff --check` is clean.

### Verdict

**Gatekeeper-approved.** Repository policy still requires an approving review from a reviewer with write access; that is the only remaining merge blocker.

PR #674, opened 2026-09-04
fix(desktop): avoid incompatible startup shells

pascalandr:
## Summary

- restrict the POSIX desktop startup script to compatible Bash and Zsh executables
- fall back from Nushell, Fish, empty, or arbitrary `$SHELL` values to `/bin/zsh` on macOS and `/bin/bash` on Linux
- preserve configured Bash/Zsh paths so existing login-environment loading remains unchanged
- keep Tauri and Electron startup behavior aligned and cover both implementations with regression tests

## Validation

- `cargo fmt --check`
- `cargo test --locked -- --test-threads=1` — 141 passed
- `npm run test:native --workspace @neuralnomads/codenomad-electron-app` — all passed
- `npm run typecheck`
- `npm run build --workspace @neuralnomads/codenomad-electron-app`
- deterministic regression coverage verifies that `/opt/homebrew/bin/nu` selects `/bin/zsh` for a macOS launch

A packaged macOS runtime smoke test is not available from the Windows development host; the platform selection is isolated and tested without relying on the host OS.

Closes #670

PR #672, opened 2026-09-04
fix(opencode): reject incompatible V1 binaries

pascalandr:
## Summary

Closes #669.

CodeNomad V2 requires the official OpenCode V2 shared-service lifecycle. A legacy binary must not be accepted merely because `--version` succeeds, nor should opening a folder display a raw legacy CLI help dump. This is an actionable compatibility fix, **not V1 runtime support or automatic installation**: install/select `opencode2` to open the workspace.

- Validate the selected executable using only `--version` and `service --help`; neither probe starts or changes the daemon.
- Require positive service-help evidence for `start`, `status`, and `get`, without pinning a release number or rejecting custom version labels.
- Recognize legacy root help, including ANSI/CRLF, stdout/stderr, and exit-zero wrappers. Preserve the localized V2 diagnosis for previously saved incompatible paths during the actual host/WSL lifecycle.
- Run both validation probes asynchronously with a 5-second subprocess timeout and 64 KiB output limit, leaving the shared backend responsive while a CLI is slow.
- Treat saved versions as display metadata only: explicit add/browse always validates the path again.
- Match the compatibility error code precisely, including its JSON HTTP envelope, so unrelated missing-file/configuration errors are not hidden by a matching filename.
- Reuse the existing translated message in all ten locales. No daemon ownership, V1 fallback, package update, or native-shell-selection change.

## Follow-up verification of the original PR

The original head `226e2d7b` already fixed the exact reported failure with a real V1 binary. Independent rechecking found and corrected adjacent gaps in this same PR:

| Check | Original PR | Completed PR |
| --- | --- | --- |
| Real OpenCode 1.18.25 rejected; real V2 accepted | PASS | PASS |
| Empty/unrelated successful help rejected | FAIL | PASS |
| Legacy help with exit zero localized at launch | FAIL | PASS |
| Other backend work proceeds during slow probes | FAIL | PASS |
| Explicit add rechecks a persisted V1 version | FAIL | PASS |
| Unrelated diagnostic mentioning `opencode_v2_required` retained | FAIL | PASS |

The new regression tests were run against the old production code first: three assertions failed (false capability acceptance, exit-zero help, unrelated-error masking), then passed after the fix. A separate source-extraction check executes the unchanged published probe and component validation body: two 250 ms subprocess delays block the old probe for about 638 ms, versus a Promise returned in about 5 ms after the fix; a cached V1 version previously produced zero validation calls and now produces one and rejects the path.

## Validation of the compatibility fix (`903d9159`)

[...]

PR #666, opened 2026-08-29
feat(browser): add autonomous native previews

pascalandr:
## Summary

- Add autonomous native browser previews for Electron and Windows Tauri, using the authenticated backend/native bridge already shared with desktop inspection.
- Ship one bundled **`codenomad.automation`** V2 plugin for browser, inspect, act and screenshot. All tools are available by backend presence; ordinary desktop startup prepares the instrumentation. Remove the Developer Mode toggle, marker-based activation and activation restart requirement.
- Provision through native discovery using the **connected daemon's authenticated `config.get` roots**, including WSL and reconnects. Share installation/lease mechanics with pruning, preserve user-authored entries and older backends' leases, and unwind stale or interrupted preparation.
- Preserve session/location ownership, inspected-host pinning, browser registration/navigation generations, local-window targeting and authenticated transport. The OpenCode daemon remains externally owned.

## Gatekeeper review — September 18, 2026

Merged current `dev` at `96ef38c6` and repeated independent UI, native and server reviews after each finding until all three scopes reported **zero remaining actionable findings**.

Corrections include ordinary-project discovery without Developer Mode; connected-daemon namespace selection and stale lease cleanup; Tauri child-webview capability isolation, native-window counting, primary-renderer reload cleanup, fragment navigation, snapshot lock ordering and deadline-fenced protocol dispatch; and Electron guest readiness, redirects and explicit navigation intent. Shared-frame regressions also cover insecure HTTP and late/rejected native IPC.

### Architecture decision

Current OpenCode V2 documents MCP `_meta.sessionID`; the historical missing-session-metadata argument no longer applies. A single native V2 plugin remains appropriate for the requested unified automation surface: it reuses the existing authenticated transport and native controllers, supports backend presence, and needs no second protocol server or mode-based permission framework. Plugin loading, tool availability and execution targeting are separate concerns.

## Validation

- Full server suite: **570 passed, 2 skipped**.
- Electron native suite: **214 passed, 4 skipped**.
- Tauri native suite: **180 passed**.
- Real Solid/browser preview suite: **14 passed**, including real isolated Electron guests and insecure HTTP.
- Actual isolated OpenCode automation fixtures: **beta-19271, 2.0.3, 2.0.5 and 2.0.7**. Covers mismatched backend/daemon environments, late discovery, all four tools without activation, execution fences, independent leases, shutdown and reopening.
[...]

PR #675, opened 2026-09-04
test(opencode): isolate automation registry cross-platform

pascalandr:
## Summary

- isolate automation bridge registry tests through both `LOCALAPPDATA` and `XDG_RUNTIME_DIR`
- suppress inherited WSL registry discovery inside these unit tests
- derive the asserted registry path with the production resolver instead of hard-coding the Windows layout

## Why

The V2 migration added a registry-pressure test that always constructed `<temp>/CodeNomad/automation-bridges`. Linux ignores `LOCALAPPDATA` and resolves the registry through `XDG_RUNTIME_DIR` or the home directory, so the test failed with `ENOENT` in every current PR while also allowing neighboring tests to touch the runner's real registry location.

## Validation

- `node --import tsx --test "packages/server/src/**/*.test.ts"` — 376 passed, 2 skipped
- reproduced in Linux CI on #672 and #674 before this fix

### DIFF
--- a/packages/server/src/opencode/automation-plugin.test.ts
+++ b/packages/server/src/opencode/automation-plugin.test.ts
@@ -6,6 +6,7 @@ import path from "node:path"
 import test from "node:test"
 import {
   AUTOMATION_BRIDGE_PATH,
+  automationBridgeDirectory,
   automationBridgeDirectories,
   createAutomationBridgeRegistration,
   parseDeveloperAction,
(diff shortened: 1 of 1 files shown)


pascalandr:
@shantur This is the isolated CI fix to prioritize. It contains only the automation registry test isolation, and the complete Linux/Windows native test and cross-platform packaging matrix is green. Please review for squash merge when available.

pascalandr:
## Gatekeeper overlap verdict for `c0106bd7`

The implementation is sound, but this PR must not be merged alongside #671.

#675 contains the same stale-registry correction as #671 and additionally refactors the other Developer Mode tests through a shared isolation helper. The two branches modify the same test block from the same base and are not patch-equivalent; merging both as independent PRs would create avoidable overlap/conflict.

To preserve the required #671 → #666 merge order, this PR is superseded by #671. The broader all-test helper can be proposed later as a clean follow-up rebased on the post-#671 `dev` head if still desirable.

**Verdict: close as superseded; do not merge this head.**

PR #677, opened 2026-09-05
fix(tauri): prevent window-state capture deadlocks

### LINKED ISSUE #676: Tauri app hangs on launch — main thread deadlocked behind hundreds of schedule_flush threads contending on window-state lock
On macOS, the app hangs indefinitely on the loading screen after launch. `codenomad-tauri` pins at 100% CPU (confirmed via `sample`).

**Root cause (from a `sample` trace):**
- The main thread is blocked inside `client_state::window::capture_window_in_memory`, waiting on a mutex, every time it handles a window move/resize event.
- At the same time, 190+ background threads are alive, all running `client_state::window::schedule_flush` — each one spawned to persist window state to disk (with an fsync + atomic rename) while holding that same mutex.
- It looks like a new `schedule_flush` thread is spawned per window move/resize event instead of debouncing or reusing one thread. With that many threads queued on the same lock and hitting disk, the main thread — which needs the same lock just to process window events and render — never gets it back.

**Environment:** macOS 26.6.2, CodeNomad 0.20.0-dev-v2-20260904-44a1816b (arm64)

**Repro:** Launch the app, move/resize the window a few times. It appears related to window flush thread spawning.

**Suggested fix:** Debounce state-flush writes to a single background worker instead of spawning a new thread per window event.

pascalandr:
## Summary

Closes #676.

This PR now addresses both mechanisms behind the window-state hang, rather than only reducing the thread count:

- one lazy named worker, a capacity-one wakeup queue, and a 250 ms debounce;
- native getters run before **any** client-state lock, including explicit close/shutdown captures;
- move/resize/scale-factor/zoom handlers enqueue only the latest geometry per known window and never acquire the disk writer lock;
- serialized publication merges queued captures, preserves normal bounds across maximize/fullscreen, and seeds clamped restore bounds before maximizing;
- rollback preserves events arriving during snapshot/clear/disable/removal publication, including the window whose policy is tentatively changing; a successful destructive commit discards those speculative events;
- shutdown closes capture admission, joins the worker without retaining the scheduler/writer mutex, drains an omitted wakeup, and retries captures merged by a failed flush before ownership release;
- no application handle retained by the idle worker;
- ownership-file validation in `load_window` moved outside the shared in-memory state lock.

Electron already debounces native geometry; its production code is unchanged, with a new parity regression covering bursts, maximization, explicit flush, and closed-window cleanup.

## Why the initial version was insufficient

The original head `bb38b1a2` fixed thread proliferation, but `capture_window_in_memory` still held `write_lock` across native getters. A slow fsync could block the UI through that mutex; a secondary-thread capture could also hold it while waiting for the same UI thread that was handling a move/resize event.

This follow-up retains the useful scheduler behavior and completes the capture/persistence separation in the same PR. No competing #676 PR is being introduced.

## Before/after evidence

An external Rust harness compiles the exact production capture, flush, release and scheduler function bodies, replacing only the native/disk adapters. It counts real Windows OS threads and uses controlled channel gates rather than relying on a randomly timed freeze.

| Assertion | Base `52f0e629` | Initial PR `bb38b1a2` | Completed PR |
| --- | --- | --- | --- |
| 128 flush requests | +128 threads | +1 thread | +1 thread |
| Capture completes before blocked disk publication is released | FAIL | FAIL | PASS |
| UI can complete move/resize while a secondary capture waits for a native getter | FAIL | FAIL | PASS |
| Explicit capture/flush retains geometry | PASS | PASS | PASS |
| Future envelopes and ephemeral windows remain protected | PASS | PASS | PASS |

Ten repetitions per compared version produced the same outcomes. This is a controlled reproduction of the concurrency mechanisms, not a claim of having run the reporter's macOS GUI session.

## Committed regression coverage

The crate now includes **19 focused scheduler/capture tests** (the original three plus sixteen additional regressions), using the actual `ClientState`, native-read seam, scheduler and atomic-write adapter. Coverage includes:

- 1,000 captures during a stalled disk write; bounded memory and latest geometry;
- no client-state mutex held by native getters; simulated secondary getter/UI event cycle;
- one worker/one trailing wakeup, callback lifetime, concurrent stop, cheap late submission and no restart after release;
- independent windows, maximization/fullscreen, zoom, unknown-window rejection;
- successful and failed destructive publication, simultaneous events on the mutating and another window, retry after failed flush;
- final persistence and cross-host ownership handoff;
- secondary ownership and future-envelope fencing.

## Validation

- Full Windows Tauri crate: **158 tests passed**, including default test parallelism.
- Electron native suite: **190 tests passed**.
- UI/Electron typechecks passed.
- Official Tauri resource preparation and server/UI builds passed during verification.
- Windows release executable builds successfully.
- `cargo fmt --check`, `git diff --check`, and workflow YAML parsing passed.
- Added an independent **macOS ARM64 (`macos-26`) Tauri test job** to exercise the crate on the reported platform; packaging now depends on it as well. Its remote result must be checked before merging.

## CI dependency and limits

The inherited Linux automation-registry test failure remains isolated in **#671**. #675 was closed as its duplicate and is not part of this fix. No duplicate registry patch or unrelated PR was merged into this branch. After #671 is integrated, update this branch and require the complete matrix before merge.

Windows controlled reproductions and tests are not a substitute for a macOS ARM64 interactive launch/move/resize smoke test. Persistent disk failures are still logged and cannot guarantee saving to an unavailable disk. No claim of absolute absence of side effects is made.

Maintenance note: `packages/tauri-app/src-tauri/src/client_state.rs` remains oversized (about 1,160 lines); focused capture and scheduler code/tests are kept in sibling modules instead of undertaking an unrelated refactor.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -50,6 +50,7 @@ jobs:
       - authorize
       - tests
       - tests-tauri-windows
+      - tests-tauri-macos
     if: ${{ needs.authorize.outputs.allowed == 'true' && !github.event.pull_request.draft }}
     uses: ./.github/workflows/build-and-upload.yml
     with:
@@ -217,3 +218,36 @@ jobs:
       - name: Test Tauri crate on Windows
         working-directory: packages/tauri-app/src-tauri
         run: cargo test --locked -- --test-threads=1
+
+  # Exercise the window persistence regressions on the architecture reported in
+  # #676, independently of the Linux server gate. Packaging alone cannot test them.
+  tests-tauri-macos:
+    needs: authorize
+    if: ${{ needs.authorize.outputs.allowed == 'true' && !github.event.pull_request.draft }}
+    runs-on: macos-26
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Setup Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Prepare Tauri test resources
+        run: >-
+          npm run dev:prep --workspace @codenomad/tauri-app &&
+          node -e "require('fs').mkdirSync('packages/tauri-app/src-tauri/resources/server',{recursive:true})"
+
+      - name: Test Tauri crate on macOS ARM64
+        working-directory: packages/tauri-app/src-tauri
+        run: cargo test --locked -- --test-threads=4
--- a/packages/electron-app/electron/main/window-state.test.ts
+++ b/packages/electron-app/electron/main/window-state.test.ts
@@ -1,10 +1,57 @@
 import assert from "node:assert/strict"
+import { EventEmitter } from "node:events"
 import test from "node:test"
 import { clampWindowBounds, installWindowZoomInput, normalizeNativeWindowState, normalizeZoomFactor, restoreWindowState, WindowStateTracker } from "./window-state"
 import type { BrowserWindow } from "electron"
 import type { ClientStateManager } from "./client-state"
 
 const primaryDisplay = { x: 0, y: 0, width: 1920, height: 1080 }
+
+test("move/resize bursts debounce and final flush preserves pre-maximize bounds", async (t) => {
+  t.mock.timers.enable({ apis: ["setTimeout"] })
+  const events = new EventEmitter()
+  let x = 0
+  let maximized = false
+  const window = Object.assign(events, {
+    isDestroyed: () => false,
+    getPosition: () => [x, 20],
+    getContentSize: () => [1200, 800],
+    isMaximized: () => maximized,
+    isFullScreen: () => false,
+    webContents: Object.assign(new EventEmitter(), {
+      isDestroyed: () => false,
+      getZoomFactor: () => 1.25,
+    }),
+  }) as unknown as BrowserWindow
+  const saved: unknown[] = []
+  const manager = {
+    activeWindowId: "window-a",
+    saveWindowState: async (state: unknown, id: string) => { saved.push({ state, id }); return true },
+    flush: async () => undefined,
+  } as unknown as ClientStateManager
+  const tracker = new WindowStateTracker(window, manager)
+  for (x = 1; x <= 1000; x++) {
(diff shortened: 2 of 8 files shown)


pascalandr (commented):
## Gatekeeper review of `bb38b1a2`

No blocking code findings remain.

The scheduler keeps one worker and a one-item wakeup queue, resets the debounce while events arrive, preserves one trailing flush when an event arrives during persistence, drains queued state before releasing cross-process ownership, and permanently rejects post-release scheduling. The shutdown lock order avoids joining a worker while holding the write lock it needs.

### Validation

- Complete Tauri suite: 142/142 passed.
- Focused scheduler regressions cover burst coalescing, trailing flushes, and disconnect draining.
- Rust formatting and `git diff --check` passed.

### Gate status

**Code-approved, CI-blocked.** The first Linux job failed on the inherited automation-registry test fixed by #671; the later rerun was externally cancelled during checkout. Rerun the complete PR matrix after #671 lands before merging.

pascalandr:
Follow-up published as 727197d55a00ce335ecb3a059eafe593d744a46f (fast-forward, retaining the original PR). The earlier worker-only review applies to bb38b1a2, not to the complete capture fix. The same five-assertion source-extraction harness now passes 10/10 repetitions; the old commit still reproduces both lock failures. All 158 Windows Rust tests pass at default parallelism, all 190 Electron tests pass, and the 19 committed scheduler/capture regressions additionally pass 20/20 repeated runs. Release build, typechecks and formatting checks pass. The PR description documents the implementation and limits; the new independent macOS ARM64 test job must be checked before merge. The existing Linux gate remains dependent on #671, and #675 stays closed as its duplicate. No competing #676 PR or unrelated registry patch was created.

pascalandr:
CI results are now confirmed for 727197d55a00ce335ecb3a059eafe593d744a46f, run https://github.com/NeuralNomadsAI/CodeNomad/actions/runs/34035208752 : macOS 26 ARM64 passed all 149 platform-applicable Rust tests, including all 19 scheduler/capture regressions; Windows passed all 158 Rust tests. Linux reached the server suite and failed only the inherited 'prunes stale registry pressure before limiting discovery' test (375 passed, 1 failed, 2 skipped), with the same ENOENT on /tmp/codenomad-automation-stale-*/CodeNomad/automation-bridges fixed by #671. Packaging was consequently skipped. The separate failed artifact-comment job inspected a superseded cancelled run, not a native test failure. No additional production change is indicated by these logs. The remaining merge gate is #671 integration followed by the full PR matrix; an interactive macOS GUI smoke test is still recommended.

PR #708, opened 2026-09-18
fix(worktrees): bound ownership refreshes and lazily cache inventories

pascalandr:
## Summary
Bound miss-triggered ownership refreshes and retain completed worktree inventories with demand-driven stale-while-revalidate reads for display consumers.

The shared native event bridge awaits directory ownership sequentially. Previously, each new foreign directory discarded the inventory and its negative cache, triggering another complete native/Git scan. Twenty foreign directories caused 21 inventory loads; the ownership regression now requires two and covers explicit invalidation and expiry.

The manager also used to retain only an in-flight inventory promise. Sequential UI reads therefore repeated the complete scan. `WorktreeInventory` now retains the last successful snapshot:
- Display reads reuse it immediately. Expiry (10 seconds) or invalidation only schedules one background refresh when a consumer next reads it; there is no polling timer.
- Concurrent callers share the scan, failures retain display data with demand-driven retry backoff, and invalidation fences obsolete results with serialized trailing validation.
- Directory authorization uses validated reads, which wait if the snapshot is stale. Family create/move/delete checks force a fresh scan. Workspace disposal fences pending results.
- A changed completed snapshot publishes `workspace.worktreesChanged`; existing UI consumers queue a reload behind older HTTP responses. Unchanged scans do not create notification feedback loops.
- The selector opens before the reload completes. Duplicate selection notifications during inventory reconciliation no longer dismiss an open menu or produce a family move, and a delayed response cannot reopen a dismissed menu.

## Measurements and scope
- Before the initial miss-refresh correction, the installed renderer lagged native events by 331-336 seconds, versus 42-84 ms on a direct read-only native subscription.
- The first post-restart sample caught up to 340-699 ms, but a subsequent 45-second sample still showed session median 851 ms, p95 12.2 seconds, and overall spikes of 15.5 seconds. The corresponding direct native session median was 52 ms and p95 339 ms. This is **not a complete startup/event-latency fix**.
- A real 165-worktree inventory measured location.get 22 ms, native refresh 282 ms, and native list 5 ms, versus 24,067 ms for the whole CodeNomad inventory. The local validation/annotation pass launches three Git commands per inspected checkout.
- This memory cache removes repeated display scans and lets warm display reads continue while stale data refreshes. It does not persist across backend restarts or reduce the cost of a cold or required authoritative scan. The serial event bridge can still await such validation.

## Validation
- 110 targeted server tests passed: inventory cache (6), ownership directory cache, event bridge, workspace lifecycle/routes, worktree routes, family transactions and evacuation.
- Five existing UI worktree store tests passed with browser conditions.
- Four Chromium tests passed: pointer/keyboard/touch actions, background completion racing an older HTTP response, immediate cached opening and Escape dismissal during a blocked refresh, and creation/selection by stable ID.
- The isolated native location/worktree fixture passed against OpenCode 2.0.7, including real branch rename, cached reuse, lazy invalidation and forced validation with native events connected.
- Server/UI typechecks, emitted server TypeScript, UI production build and diff check passed.

## Installed state
Only the earlier compiled ownership-cache module (`70af9c84`) has been installed locally. The new inventory cache and selector changes are built and tested in the existing worktree but have **not** been deployed to the running installation. No shared daemon or user profile was restarted or modified for these new tests.


pascalandr:
Follow-up live measurement (45 seconds, after restart): the multi-minute backlog is reduced, but this is not a complete startup/latency fix. Session event lag in the renderer had a median of 851 ms and p95 of 12.2 s, with overall spikes up to 15.5 s. Direct native events had session median 52 ms and p95 339 ms.

Startup resource timing also shows a 9.6 s worktree request and an 18.5 s project metadata request. Investigating those remaining inventory/hydration costs separately from this bounded miss-refresh correction.

pascalandr:
The remaining inventory cost is now isolated on the real repository: native location.get 22 ms, worktree.refresh 282 ms, worktree.list 5 ms for 165 entries; complete listNativeWorktrees took 24,067 ms and returned 165 owned worktrees. The implementation launches three Git commands per inspected checkout (495 for those entries, plus root checks), at up to eight concurrent checkouts. The dominant remaining cost is the local validation/annotation pass rather than the native list API. This PR bounds repetition of that pass; it does not eliminate its cold-start cost.

PR #717, opened 2026-09-19
fix(ui): keep command and search windows non-modal and persistent

pascalandr:
## Summary
- Share a non-modal utility window for the command palette and transcript search, without a scrim or outside-click dismissal.
- Mark open header buttons and toggle them on repeated clicks; preserve Escape, close buttons and command execution dismissal.
- Scope search visibility to its instance/session and close inactive views.
- Preserve input focus when launching from compact overflow menus and correct search positioning in RTL.

## Validation
- Five Chromium browser tests using the real InstanceShell, palette and transcript search.
- UI TypeScript check passed; git diff --check passed.
- Rendered RTL captures inspected.

## Review gate
An independent autonomous gatekeeper review will be published before merging. Administrative squash merge is authorized only if that review has no outstanding findings.

## Maintenance notes
Existing oversized files touched: instance-shell2.tsx (~1,296 lines), message-section.tsx (~1,440 lines).


pascalandr (commented):
## Autonomous gatekeeper review — changes required

Reviewed full HEAD: **6abe3e170c6229f4ec63aa4b8fb5b706e0ffdaf9** against `origin/dev`.

**Verdict: BLOCKED — two P2 findings. Administrative merge is not authorized by this review.**

### 1. [P2] Consume utility-window Escape before the global session shortcut sees it

Location: [`packages/ui/src/components/dismissible-window.tsx:19-23`](https://github.com/NeuralNomadsAI/CodeNomad/blob/6abe3e170c6229f4ec63aa4b8fb5b706e0ffdaf9/packages/ui/src/components/dismissible-window.tsx#L19-L23), together with the removed document Escape handler in `message-section.tsx`.

Reproduction: while a session is working, open transcript search, click outside it, then press Escape. Search closes, but the same key event reaches `use-app-lifecycle.ts`'s window listener and arms the double-Escape abort state. A subsequent Escape within the debounce interval therefore counts as the second abort press, although the first was used to dismiss search. Previously the search document handler called `stopPropagation()` when dismissing. Kobalte's default dismissal only calls `preventDefault()`, and the application listener does not check `defaultPrevented`.

Confirmed in isolated Chromium with the real `useAppLifecycle` mounted: one outside-focus Escape closed search and emitted `setEscapeInDebounce(true)`. With palette and search both open, one outside-focus Escape also closed both: Kobalte dismissed search, then the global handler saw the remaining dialog and called `hideCommandPalette()`.

Consume Escape when a utility window actually handles it, or make the application shortcut dispatcher respect handled events; preserve topmost-window dismissal. Add a regression with the real global keyboard lifecycle and a working session, rather than only InstanceShell.

### 2. [P2] Refocus the persistent palette when its keyboard shortcut is invoked again

Location: [`packages/ui/src/components/command-palette.tsx:220-227`](https://github.com/NeuralNomadsAI/CodeNomad/blob/6abe3e170c6229f4ec63aa4b8fb5b706e0ffdaf9/packages/ui/src/components/command-palette.tsx#L220-L227).

Reproduction: open the palette with Ctrl+Shift+P, click the composer (the palette correctly stays open), then press Ctrl+Shift+P again. Focus remains in the composer, so subsequent typing changes the prompt instead of filtering commands. The same problem applies to the Cmd variant.

The existing shortcut calls `showCommandPalette(instance.id)`, which only sets an already-true open state. The new non-modal window autofocus runs only when mounted, and there is no focus request for an already-open palette. Confirmed in isolated Chromium with the real keyboard lifecycle: after the second shortcut, `document.activeElement` was still the outside target and the palette input was not focused. Give repeated show requests a focus path while retaining header-button toggle behavior, and cover open → outside focus → shortcut → palette input focus.

### Checks and scope

- Read applicable `AGENTS.md`, the actual diff, and surrounding search/session authority, keyboard dispatch, Kobalte dismissal/focus internals, and all shared ActionOverflowMenu consumer action definitions.
- `node --import tsx --test tests/browser/header-windows.test.ts`: **5/5 passed** on the reviewed rebased HEAD.
- `npm run typecheck --workspace @codenomad/ui`: **passed**.
- `git diff --check origin/dev...HEAD`: **passed**.
- Additional read-only Chromium probes using the existing fixture plus the real `useAppLifecycle` reproduced both findings above.
- The supplied fixture covers persistent outside interaction, explicit toggles, search/session transitions, execution, compact touch controls and RTL, but does not mount the global keyboard lifecycle; that is why it misses these failures.
- No additional concrete defect found in the delayed overflow actions or inspected inactive-session handling. Native Electron/Tauri, macOS keyboard events, and exhaustive runtime coverage of every overflow consumer were not exercised.

No source changes were made during this review. Please fix and request re-review of the new full HEAD before merging.

pascalandr (commented):
## Autonomous gatekeeper re-review — no outstanding findings

Reviewed full HEAD: **eb84bc8d4256bab89349fd875c01de7e79f0c700** against `origin/dev`, including the corrective diff from previously reviewed `6abe3e170c6229f4ec63aa4b8fb5b706e0ffdaf9`. Local HEAD and the published PR HEAD match.

**Verdict: PASS — no actionable findings. Both previously reported P2 findings are resolved.** This supersedes the blocked verdict in [review 5257155900](https://github.com/NeuralNomadsAI/CodeNomad/pull/717#pullrequestreview-5257155900). The zero-outstanding-findings review gate is satisfied for this exact HEAD. No merge was performed.

### Resolution verification

1. **Escape is consumed by the topmost utility window.** The shared Kobalte `onEscapeKeyDown` prevents default and stops immediate propagation before dismissal; removing the local input/palette Escape handlers avoids bypassing the layer guard. The real application lifecycle regression confirms that stacked-window dismissal neither closes the lower window nor arms Stop, and that double-Escape still legitimately interrupts once both windows are closed.
2. **Repeated palette shortcuts restore input focus without clearing query or composer draft.** Per-instance focus requests advance independently of the open boolean. The palette's focus effect responds without retriggering the open-state reset. The real keyboard regression confirms that subsequent typing updates the palette query while preserving the composer draft.

### Independently executed validation

- `node --import tsx --test tests/browser/header-windows.test.ts`: **7/7 passed** on the reviewed HEAD.
- `npm run typecheck --workspace @codenomad/ui`: **passed**.
- `git diff --check origin/dev...HEAD`: **passed**.
- Additional isolated Chromium probes using the real fixture/lifecycle: **passed** for reversed stacking order (search before palette), Escape from a focused palette with zero results, Escape from focused search, Escape while the lower window owns focus, explicit close/reopen query reset, absence of Stop leakage, and absence of page errors.
- Inspected the follow-up's store batching, per-instance focus-request wiring, shared dismissal behavior and revised regression fixture. No additional concrete regression found. The earlier review's shared overflow-consumer and inactive-session inspection remains applicable; those implementations were not changed by the correction.

Validation limitations: native Electron/Tauri hosts, macOS keyboard events and exhaustive runtime exercise of every overflow-menu consumer were not performed. This is a comment review rather than a GitHub approval, as permitted for the autonomous gatekeeper workflow.

Read-only review completed; no source edits or commits.

PR #718, opened 2026-09-19
feat(sessions): fork after user and assistant messages

pascalandr:
## Summary

I added **Fork after this message** to completed agent responses and user prompts. Both actions keep the selected message and open the new conversation with an empty composer. The source conversation and its draft are preserved.

I resolve the cut against native message history, including hidden records and cursor pages. A tail fork checks the native boundary; if another client appended in the meantime, I discard that fork rather than open the wrong history. Pending prompts and unfinished responses do not offer the action.

Closes #693.

## Validation

- UI typecheck passed.
- 7 fork regressions and 65 existing request-authority tests passed.
- 5 Chromium scenarios passed: prompt, historical response, latest response, unfinished response, and narrow overflow menu. These exercise the real SessionView, empty composer, preserved source draft, and absence of an extra prompt.
- The isolated native fixture passed on OpenCode 2.0.10 and beta-19271, including page crossings, a hidden model-change message, both tail roles, and preserved source history. I used private homes/databases and left the shared daemon untouched.
- I added the unit regression to CI; the browser regression is picked up by the existing browser suite.

The shared UI implementation applies to Tauri and Electron. I have not rebuilt or installed a desktop binary.

## Maintenance note

The touched existing files are already large: `session-api.ts` (~1,826 lines), `message-item.tsx` (~817), and `session/session-view.tsx` (~693). I kept the new boundary logic in a separate module.

PR #719, opened 2026-09-19
fix(events): isolate slow routing from the shared native stream

pascalandr:
## Summary
Live transcript events were measured arriving 139–160 seconds late through CodeNomad while the direct native stream arrived in 22–55 ms. Serial event routing awaited ownership checks across every logical workspace before accepting the next SDK event.

- Buffer before slow routing I/O, preserving per-session/PTY/Shell ordering and independent per-recipient FIFO delivery.
- Share validated ownership checks by recipient and complete native location; fence inventory invalidation, disconnected generations and stopped/reopened workspace incarnations.
- Bound retained jobs, payload size and routing age; reconnect through normal authoritative recovery on overflow instead of silently dropping deltas.
- Cancel pending session lookups and add rate-limited stage-specific lag diagnostics without event contents.

## Base
Rebased onto upstream/dev at 7312ac22, including #717 and #718.

## Validation
- 43 targeted server routing/worktree regressions passed.
- Server TypeScript compilation and diff whitespace check passed.
- Isolated native OpenCode 2.0.9 fixture passed, including a deliberately blocked recipient and another subscriber on the same SDK client.

## Desktop testing
A Windows Tauri build of this PR is being prepared for manual testing with the existing profile. Live before/after measurements will be added after installation.


pascalandr:
Windows Tauri build 0.20.0-dev-20260919-70647193 compiled and installed successfully. Executable SHA-256 matches the build; packaged-resource smoke checks passed. The 43 targeted tests passed again after rebasing onto upstream/dev. Existing dev profile configuration hash was preserved and the shared OpenCode daemon stayed alive. Native automation confirmed the same window/session was restored. Initial 12-second observation: two session events reached CodeNomad in 140–315 ms, versus the previous 139–160 seconds. One event matched across both streams: 57 ms direct native, 140 ms through CodeNomad. This small sample is preliminary, not a full latency benchmark; other event types included older events up to 11.8 seconds. An initial history-loading error cleared without manual retry; 29 message rows were rendered afterward. Manual testing is now available.

PR #724, opened 2026-09-20
feat(ui): add workspace-root navigation to directory browser

bluelovers:
## Summary

Adds a **"Workspace Root"** button to the directory browser dialog that navigates the view back to the configured workspace root (`FileSystemListingMetadata.rootPath`).

## What changed

- **New button**: a Home-icon button that calls `navigateTo(rootPath())` to jump back to the workspace root. It is disabled when there is no root path, when the user is already at the root, or while a folder is being created (driven by a new `canGoToWorkspaceRoot` memo).
- **Placement / separation**: the button lives in its own grid cell to the right of the "Select folder or enter path" label, fully separated from the `New Folder` and `Open` actions. It uses a new transparent `selector-button-ghost` variant so it reads as a navigation tool rather than a filled action button.
- **i18n**: added `directoryBrowser.goToWorkspaceRoot` for `en` plus `de`, `es`, `fr`, `he`, `ja`, `ne`, `ru`, `tr`, and `zh-Hans` (used for tooltip/`aria-label`).
- **Layout**: `.directory-browser-current` now uses a four-column grid; the flexible middle column pushes `New Folder` to the far right and keeps the root button next to the label. Responsive rules stack the controls at `<=640px` and `<=380px`.

## Edge cases

- In files mode the `New Folder` action is hidden, but the Workspace Root button stays available so users can still jump back to the root.
- For the Windows drives pseudo-view `rootPath` is empty, so the button is disabled.

## Validation

- No typecheck was run in this environment (`npx` is disallowed here); manual browser verification of spacing and the localized label is recommended.

🤖 Generated with [CodeNomad](https://github.com/NeuralNomadsAI/CodeNomad)


packages/ui/src/components/directory-browser-dialog.tsx
```diff
@@ -423,6 +426,16 @@ const DirectoryBrowserDialog: Component<DirectoryBrowserDialogProps> = (props) =
                     aria-label={t("directoryBrowser.currentFolder.inputAriaLabel")}
                     class="selector-input directory-browser-current-path"
                   />
+                  <button
+                    type="button"
+                    class="selector-button selector-button-ghost directory-browser-go-root"
+                    disabled={!canGoToWorkspaceRoot()}
+                    onClick={() => void navigateTo(rootPath())}
```
pascalandr: This bypasses `handleNavigateTo()`, which deliberately clears `pathInputDirty` before changing directories. Reproduction: navigate below the root, edit the path field without submitting it, then click this button. The listing moves to the root but the field keeps the stale edited path; the next **Open** or **New Folder** action operates on that stale target. Route this action through the shared navigation handler (or otherwise clear/synchronize the edit state) and cover the sequence with a regression test.

packages/ui/src/styles/components/directory-browser.css
```diff
@@ -82,8 +82,13 @@
   width: auto;
 }
 
+.directory-browser-go-root {
+  grid-area: go-root;
+  width: auto;
+  flex-shrink: 0;
+}
+
 .directory-browser-open-path {
```
pascalandr, line 99: The grid templates still reserve an `open` area, but this rule no longer assigns `grid-area: open`. Chromium therefore auto-places **Open** into the desktop spacer on the label row; below 640px it occupies only half the intended row, and in files mode it lands beside the root button. Restore the explicit area assignment. Please cover both dialog modes and responsive layouts so this cannot regress silently.

pascalandr (changes requested):
The convenience is understandable, but the destination needs an explicit product contract before this shared control is added.

`FileSystemListingMetadata.rootPath` is the server browsing/authorization origin, not necessarily the dialog's starting folder or the active workspace. Electron and Tauri launch the server with `--unrestricted-root` and no explicit `--workspace-root`, so this value defaults to the server process working directory. The shared dialog is used for workspace selection, clone destinations, OpenCode binaries, and file attachments, each with different `initialPath` semantics. A Home icon labelled “Workspace Root” therefore has no stable meaning across callers.

Please first choose the intended destination and model it explicitly: capture the canonical initial location for “back to start”, pass an active workspace target from callers, use `homePath` for filesystem home, or clearly expose the configured browsing root. The editable path field does not provide history. CodeNomad currently persists recent opened project folders, but has no generic directory history or favourites; those should not be conflated. A per-dialog start shortcut or back/forward history can stay local, while persistent favourites deserve a separate scoped design.

There are also two confirmed implementation regressions in the inline comments. Please add a real browser regression covering directory and file modes at desktop and narrow widths, plus the edited-path navigation case.

packages/ui/tests/browser/fixtures/directory-browser.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const params = new URLSearchParams(location.search)
+const initialPath = params.get("initialPath") ?? ""
+const mode = (params.get("mode") as "directories" | "files") ?? "directories"
+const title = params.get("title") ?? "Test"
+
+let navigations: string[] = []
+const [open, setOpen] = createSignal(true)
+
+render(
```
pascalandr, line 17: `DirectoryBrowserDialog` calls `useI18n()` immediately, so rendering it without `I18nProvider` throws before the dialog mounts; `I18nProvider` in turn requires `ConfigProvider`. This is why all five new tests time out waiting for `.directory-browser-current-path` in required CI instead of exercising any assertions. Wrap this fixture in the same `ConfigProvider` + `I18nProvider` stack used by the other real-component fixtures (and preferably surface `pageerror` so fixture startup failures fail fast).

packages/ui/src/components/directory-browser-dialog.tsx
```diff
@@ -289,6 +289,55 @@ const DirectoryBrowserDialog: Component<DirectoryBrowserDialogProps> = (props) =
...
+    if (
+      meta.scope === "unrestricted" &&
+      meta.homePath &&
+      normalizePathKey(meta.homePath) !== normalizePathKey(meta.rootPath)
+    ) {
+      shortcuts.push({ id: "home", target: meta.homePath, labelKey: "directoryBrowser.goToHome" })
+    }
+    const initial = props.initialPath?.trim()
+    if (initial) {
+      const target = isAbsolutePathLike(initial) ? initial : resolveAbsolutePath(meta.rootPath, initial)
```
pascalandr: This does not preserve the server's `initialPath` semantics. `initialize()` sends a relative initial path unchanged, and unrestricted `FileSystemBrowser.resolveUnrestrictedPath()` resolves it against `homePath`; this line instead rewrites it under `rootPath`. With root `/srv/start`, home `/home/user`, and initial `projects`, the dialog opens `/home/user/projects` but **Initial Path** jumps to `/srv/start/projects`. The button also remains after an invalid/missing initial path made `initialize()` fall back to the default root, leaving a permanently failing shortcut. Capture the canonical absolute path only from a successful initial navigation and use that for the target/deduplication; omit the shortcut after fallback.

packages/ui/tests/browser/directory-browser.test.ts
```diff
@@ -0,0 +1,174 @@
...
+})
+
+test("restricted scope shows only the workspace-root shortcut and dedupes the initial path", async () => {
+  const page = await browser.newPage()
+  try {
+    const scenario: Scenario = { scope: "restricted", rootPath: "/ws", homePath: "/home" }
+    // initialPath equals rootPath, so the workspace + initial shortcuts collapse into one.
+    await openFixture(page, scenario, "initialPath=/ws&mode=directories")
+    assert.equal(await shortcutCount(page), 1)
+    await page.locator(".directory-browser-shortcut").first().click()
```
pascalandr: Once the fixture mounts, this click still cannot succeed: the initial path is `/ws`, the sole deduplicated shortcut also targets `/ws`, and `atTarget()` therefore disables the button. Playwright will wait for it to become enabled and time out. Assert that the deduplicated button is disabled here, or initialize below the root before testing navigation.

packages/ui/tests/browser/directory-browser.test.ts
```diff
@@ -0,0 +1,174 @@
...
+  const page = await browser.newPage()
+  try {
+    await page.setViewportSize({ width: 360, height: 800 })
+    const scenario: Scenario = { scope: "unrestricted", rootPath: "/cwd", homePath: "/home" }
+    await openFixture(page, scenario, "initialPath=/projects/start&mode=directories")
+    assert.equal(await shortcutCount(page), 3)
+    assert.equal(await page.locator(".directory-browser-open-path").count(), 1)
+    // Open button should span the full row at this width, not half.
+    const box = await page.locator(".directory-browser-open-path").boundingBox()
+    assert.ok(box !== null && box.width >= 300, `open button width ${box?.width} should fill the row`)
```
pascalandr: At a 360 px viewport the modal is at most `90vw` and the body has 48 px horizontal padding, so the Open button cannot be 300 px wide; this assertion will fail after the fixture is repaired. Compare its width/edges with the grid content box instead of a hard-coded viewport threshold. Also, 360 px exercises the separate `<=380px` template, not the `381–640px` two-column breakpoint where the original half-row/file-mode regression occurred; please cover that breakpoint in both directory and file modes as requested.

pascalandr (changes requested):
The destination contract is now much clearer, and the two production regressions from the previous review are corrected: shortcut navigation clears the edited-path state and `grid-area: open` is restored.

This HEAD is still not merge-ready. The new browser suite never mounts the dialog and all five tests fail in required CI; after fixing the fixture, two assertions are themselves guaranteed to fail. There is also a user-visible mismatch in how a relative initial path is resolved in unrestricted mode, plus a broken shortcut after initial-path fallback. Please address the inline findings and rerun the browser suite. No merge performed.

packages/ui/src/components/directory-browser-dialog.tsx
```diff
@@ -86,6 +86,7 @@ const DirectoryBrowserDialog: Component<DirectoryBrowserDialogProps> = (props) =
   const [loadingPaths, setLoadingPaths] = createSignal<Set<string>>(new Set())
   const [currentPathKey, setCurrentPathKey] = createSignal<string | null>(null)
   const [currentMetadata, setCurrentMetadata] = createSignal<FileSystemListingMetadata | null>(null)
+  const [initialAbsolutePath, setInitialAbsolutePath] = createSignal("")
```
pascalandr, line 89: This new per-opening value is not cleared by `resetState()`. The dialog component remains mounted while `open` is false, so a successful first opening can leak its captured target into a later opening whose `initialPath` is absent or rejected. After fallback to the default root, the old Initial Path shortcut is shown and points back to the stale/failing location, contrary to the comment in `initialize()`. Clear the signal before each initialization and cover successful open → close → fallback/reopen.

packages/ui/tests/browser/directory-browser.test.ts
```diff
@@ -0,0 +1,208 @@
...
+  )
+  try {
+    await page.setViewportSize({ width: 600, height: 900 })
+    assert.equal(await shortcutCount(page), 3)
+    assert.equal(await page.locator(".directory-browser-open-path").count(), 1)
+    const bodyBox = await page.locator(".directory-browser-body").boundingBox()
+    const openBox = await page.locator(".directory-browser-open-path").boundingBox()
+    assert.ok(bodyBox && openBox, "layout boxes must be measurable")
+    // Open spans the full content row at the 381-640px two-column breakpoint.
+    assert.ok(openBox.width >= bodyBox.width - 16, `open width ${openBox.width} should fill content width ${bodyBox.width}`)
```
pascalandr, line 182: This still compares against the body border box rather than the grid content box. `.directory-browser-body` includes horizontal padding, so required CI on this exact HEAD measures the button at 512px and the body at 536px; this assertion and its file-mode copy at line 203 both fail. Compare the button with `.directory-browser-current` (preferably matching left/right edges with a small subpixel tolerance) instead.

packages/ui/tests/browser/directory-browser.test.ts
```diff
@@ -0,0 +1,208 @@
...
+  const currentPath = requestedPath && requestedPath.length > 0 ? requestedPath : scenario.rootPath
+  return {
+    entries: [],
+    metadata: {
+      scope: scenario.scope,
+      currentPath,
+      parentPath: scenario.parentPath,
+      rootPath: scenario.rootPath,
+      homePath: scenario.homePath,
+      displayPath: currentPath,
```
pascalandr, line 27: The fixture forces every response to `pathKind: "absolute"` and echoes the requested path, while every test supplies an absolute `initialPath`. Consequently the regression motivating the production change—an unrestricted relative initial path being canonicalized under `homePath` rather than `rootPath`—would not be caught here. Add a scenario such as `initialPath=projects` whose fake server returns `/home/projects`, navigate away, then assert that the Initial Path shortcut returns to that canonical target.

pascalandr (changes requested):
The fixture now mounts and capturing the canonical result of the initial navigation is the right direction. This HEAD is still not merge-ready: the required browser job fails both new responsive assertions, the captured initial shortcut can survive a later fallback when the mounted dialog is reopened, and the relative-path contract being fixed is not covered by the fixture. Please address the inline findings and rerun the focused browser suite.

pascalandr (dismissed):
Re-reviewed the updated HEAD `77c3d2a6296c7ad66448ed8916b9c99ff82ab8ea`. The outstanding findings from my previous review are addressed; no remaining blocking finding in the directory-browser changes.

Confirmed:
- `resetState()` clears the captured canonical initial target before each opening, preventing the old shortcut from surviving a subsequent opening without an initial path.
- Both responsive tests now compare the Open button's left/right edges against `.directory-browser-current`, at the actual two-column breakpoint and in both dialog modes.
- The relative-path fixture now models unrestricted resolution under `homePath`, and the regression navigates away and back to the canonical initial location.
- The prior fixes remain in place: provider-backed fixture, edited-path synchronization, explicit Open grid area, and distinct/deduplicated shortcut destinations.

Validation on this exact HEAD in an isolated worktree:
- UI typecheck: passed.
- `node --import tsx --test tests/browser/directory-browser.test.ts`: 8 passed, 0 failed.

CI/integration follow-up before merge: run 35870907908 is still red and did not reach the browser suite. Its current workflow invokes files absent from this older branch (`view-menu.test.ts`, `runtime-support.ts`, `test-session-aside-native.mjs`, and `test-opencode-install-native.mjs`). Tauri Windows additionally reports two cross-host client-state election failures (178 passed, 2 failed), outside this PR's diff. Please synchronize with current `dev` and rerun required CI; this code approval is not a claim that the global checks pass. No merge performed.

pascalandr (commented):
--

pascalandr:
@bluelovers I pushed a follow-up to this PR (`55e25b8a`) after revisiting the UX with the maintainer. Thank you for working through the earlier review rounds. The implementation was technically sound, but we weren't convinced that three permanent icons for the server root, Home, and the initial path were easy to understand across this dialog's different uses.

I kept your canonical initial-path handling and changed the navigation surface instead:

- The existing path field now reveals contextual destinations when focused. After leaving an explicitly supplied starting folder, it offers **Back to [folder name]**; it also offers Home in unrestricted browsing and the parent folder when applicable. There is no shortcut to the server's default working directory.
- Typing a folder-name prefix suggests matching directories from the listing already loaded. Arrow keys + Enter choose a suggestion; Enter without a selection still opens the typed path; Escape discards an unsubmitted edit. The suggestions expand below the field instead of covering Open, New Folder, or the directory list.
- The address component is used by the shared dialog, so project selection, clone destinations, attachments, and executable selection get the same behavior. Labels are localized in every existing locale. I also fenced late responses from an earlier opening so they cannot overwrite a reopened dialog.

I reviewed the final local diff for regressions, inspected the actual 600px and 360px browser renders, and passed the 11 focused browser scenarios, UI typecheck, UI build, and `git diff --check`. The CI run for this new HEAD has just started.

**Does this change of UX direction work for you?** In particular, does the contextual return destination still address the navigation problem that motivated your original Workspace Root button? Please say if you'd prefer a different interaction or label. We'll wait for your feedback before deciding on the merge.

pascalandr:
@bluelovers Here are actual browser captures of the updated shared directory browser (English locale, deterministic example paths). These show the interaction I described [above](https://github.com/NeuralNomadsAI/CodeNomad/pull/724#issuecomment-5797613858):

**Focus after moving away from the starting folder** — “Back to start” and Home appear under the path field; the normal folder list is below it.

Directory browser with contextual Back to start and Home destinations

**Typing `/srv/start/pic`** — matching `pictures` is offered for keyboard or mouse selection, alongside the available location shortcuts.

Path field suggesting the matching pictures folder

I noticed the old “Up one level” list row repeated the Parent folder destination in the address. That duplication is fixed in `cfd7a05a`: parent navigation stays in the address, and the redundant row is gone in both folder and file modes (13 focused browser scenarios pass). Does the resulting UI and interaction work for you?

PR #727, opened 2026-09-20
fix(chat): hide system context updates by default

pascalandr:
## Summary

I added **System messages** to Message content and Chat settings, hidden by default. OpenCode context updates such as "Today's date is now..." now have a dedicated System disclosure instead of appearing as agent replies.

I preserve native types and content, persist hidden/collapsed/expanded modes, and keep hidden system text out of search. Visible system cards do not inherit assistant actions or timeline markers. Labels are translated in all ten locales.

## Validation

- UI and server typechecks passed.
- 115 unit/integration regressions passed.
- Three real SessionView browser scenarios passed, covering initial hydration, native live instructions events, popup/Chat settings synchronization, disclosure controls, and persisted settings after reopening.

## Maintenance

Existing large files touched: `packages/ui/src/components/message-block.tsx` (~1,920 lines), `message-section.tsx` (~1,443), `message-timeline.tsx` (~652), and `packages/ui/src/stores/preferences.tsx` (~1,422). I kept the new System renderer in a separate file.


pascalandr (commented):
Gatekeeper pass 1 — self-review of edd6dc90.

[P2] Restore search highlights when reopening a System card (`packages/ui/src/components/system-message.tsx:26-27`). I reproduced this with the real SessionView: show System messages, search for system-reminder, then collapse and reopen the card. The match count remains 2 / 2, but the recreated body has no search marks. The parent highlighting effect does not observe disclosure changes. I added a failing browser regression.

I also found a pre-existing integration blocker in `message-section.tsx`: the search effect tracks synchronous paging reads, then paging writes restart that effect until it throws Maximum call stack size exceeded. I am isolating those imperative calls from dependency tracking so the new System search behavior can be exercised end to end.

I will push the fixes and repeat this review.

pascalandr (commented):
Gatekeeper pass 2 — review of d645ce256b84aacca788c6afe399b08870c3934d.

I rechecked the full PR and the corrective commit. I found no remaining confirmed findings.

I fixed the pass-1 disclosure finding: reopening a System card now restores search marks. I also fixed the pre-existing reactive paging loop by keeping imperative traversal and result location outside Solid dependency tracking. Query, session, connection and mutation authority checks remain in place.

I verified hidden/collapsed/expanded preferences, native history and live instructions, persistence, active-search visibility changes, occurrence navigation, disclosure reopening and ordinary assistant search. UI typecheck, 136 focused unit/integration tests and all five real SessionView browser scenarios passed; diff check is clean.

PR #729, opened 2026-09-20
feat(chat): upload files from device alongside server attachments

### LINKED ISSUE #725: feat: let attachFiles use the native system file picker for uploads
## Feature request (additive — does NOT replace the existing selector)

Allow `attachFiles` to **additionally** support uploading files through the operating system's **native file browser / file picker** dialog. This is meant to sit **alongside** the current in-UI path selector, not replace it. Both options should remain available.

## Motivation / benefits

1. **Faster and more convenient file discovery.** Native OS file pickers usually provide thumbnails/previews, search, and recent-files, making it much easier to locate the file you want to attach compared to typing or navigating a bare path list.
2. **Cross-environment uploads.** When the CodeNomad server runs in a different environment than the user's local machine (e.g. remote host, WSL, or a container), the in-UI selector appears to be restricted to paths on the **server's** filesystem. A native picker on the user's own device would let them upload files from their local environment, which the current server-side selector cannot reach.

## Proposal

- Keep the existing server-side path selector exactly as it is.
- **Add** an "Upload from device" / "Browse…" entry next to it in the attachment UI.
- The chosen local files would be streamed/uploaded to the server (or otherwise made available to the session) so they can be attached just like files selected via the path picker.
- The two flows are complementary: use the native picker for local/convenience uploads, and keep the path selector for server-side paths when both environments are the same.

## Additional context

This request is based on the assumption that the current UI file selector resolves paths on the host where the server process runs. If that assumption is wrong, the cross-environment benefit may require a different mechanism, but the native-picker convenience benefit still stands on its own. The key point is that this is an **extra capability**, not a replacement of what exists today.

🤖 Generated with [CodeNomad](https://github.com/NeuralNomadsAI/CodeNomad)

pascalandr:
## Summary

Closes #725.

I added **Upload from device** alongside the existing server-side **Attach files** browser. Both choices are available in web, local desktop and remote desktop windows.

Device selection uses the native file picker and sends file bytes through the existing prompt attachment flow. It never falls back to a device-local path that a remote server or WSL environment cannot read. I kept the existing 5 MB per-file limit, report unreadable or oversized files, and block sending while files are being read. Late reads cannot attach files to a different session or an unmounted composer.

I translated the new labels in all ten locales.

## Validation

- Nine Chromium scenarios passed with the real SessionView and native client prompt serialization, covering web/Tauri/Electron runtime contexts, multiple files, cancellation, reselection, server browsing, size limits, failed reads and session changes.
- 62 attachment, persistence and prompt-submission tests passed.
- UI typecheck and `git diff --check` passed.
- I inspected a rendered capture of the attachment menu.

## Maintenance

Existing large files touched: `packages/ui/src/components/prompt-input.tsx` (~1,144 lines) and `packages/ui/src/components/prompt-input/usePromptAttachments.ts` (~529 lines).


pascalandr (commented):
## Gatekeeper verdict: BLOCK — not merge-ready

### Issue relevance

#725 is valid and worth solving. The existing filesystem browser resolves on the CodeNomad server, so it cannot reach device-local files when the renderer and server are separated by a remote host, WSL, or a container. Keeping the server browser and adding a device-byte flow is the correct product model. A standard `<input type="file">` is also preferable to separate Electron/Tauri adapters, and reusing byte-backed OpenCode prompt files avoids temporary-file lifecycle and path-translation problems.

### Blocking regressions

1. **The real transport rejects files far below the advertised 5 MiB limit.** Production clients send prompts through `/workspaces/:id/instance/...`, whose Fastify JSON parser retains the default ~1 MiB body limit. Base64 adds roughly 33% plus JSON overhead. A minimal reproduction using the exact parser configuration accepts a 700 KiB raw file but rejects an 800 KiB raw file with `413 FST_ERR_CTP_BODY_TOO_LARGE`. The desktop remote-proxy server independently uses the same default limit with its raw-buffer parser, so remote desktop requests can fail at two hops. Multiple smaller files fail on their combined encoded size as well.

2. **The selection has no aggregate count or byte budget.** `multiple` permits an arbitrary number of individually valid 5 MiB files. Every file is read concurrently and retained as both `Uint8Array` and a base64 data URI before submission. A large selection can freeze or exhaust the renderer long before any server-side limit responds. Raising the proxy limit without first defining and enforcing a total budget would turn the accidental 1 MiB guard into a larger denial-of-service surface.

### Required direction

- Define one explicit maximum file count and aggregate raw-byte budget, in addition to the per-file limit.
- Reject the excess before starting `FileReader`s; use bounded/sequential reads, preserve picker order, and aggregate failures into one result.
- Mirror the accepted request budget at the server seam. Raise body limits only on the necessary authenticated proxy paths, including the desktop remote proxy, with a finite value that accounts for base64/JSON overhead.
- Add a server integration regression that sends a near-limit prompt through the real CodeNomad instance proxy, plus remote-proxy coverage. Verify the supported payload against the real OpenCode endpoint too.
- Keep one explicit device-file interface rather than leaking a `requireData` boolean through the generic selection interface. That module should own limits, byte-only semantics, session/disposal fencing, ordering, and failure aggregation.

### Product/UI follow-up

The rendered menu is mechanically sound, but **Attach files** versus **Upload from device** does not explain the essential origin distinction. In remote contexts use unambiguous labels such as **Browse workspace/server files** and **Upload from this device**.

The five “web/Tauri/Electron” browser cases all execute in Chromium and only change runtime globals; they do not exercise Electron, WebView2, WKWebView, or WebKitGTK file pickers. Since the implementation is intentionally host-agnostic, one shared browser behavior test plus an honest native smoke is more valuable than five duplicate scenarios.

### Validation performed

- UI and Electron typechecks: pass.
- `git diff --check`: pass.
- New Playwright suite: 9/9 pass and rendered menu inspected.
- Instance proxy suite: 38/38 pass, but contains no large prompt-body case.
- Broader targeted attachment/session/client-state tests: 118 pass, 1 environment-specific skip.
- GitHub test/runtime jobs are green; package builds were still running at review time.

The green CI does not invalidate the blockers because the new browser fixture intercepts prompt traffic before either production proxy parser sees the payload.

packages/ui/tests/browser/fixtures/device-upload.tsx
```diff
@@ -0,0 +1,49 @@
...
+const native = OpenCode.make({ baseUrl: location.origin })
+const info = (id: string) => ({ id, title: id, agent: "build", model, projectID: "fixture", location: { directory },
+  cost: 0, tokens: { input: 0, output: 0, reasoning: 0, cache: { read: 0, write: 0 } }, time: { created: 1, updated: 1 } })
+const client: any = {
+  session: {
+    active: async () => ({}), inbox: { list: async () => ({ data: [] }) },
+    list: async () => ({ data: [info("source"), info("other")], cursor: {} }), get: async ({ sessionID }: any) => info(sessionID),
+    update: async ({ sessionID }: any) => info(sessionID),
+    instructions: { entry: { remove: async () => {}, put: async () => {} } },
+    switchAgent: async () => {}, switchModel: async () => {}, prompt: native.session.prompt,
```
pascalandr, line 28: **Blocker:** this borrowed `prompt` closure is created with `baseUrl: location.origin`, whereas production clients use `/workspaces/:id/instance/`. The Playwright route interception then fulfills the request before either CodeNomad Fastify parser sees it. Consequently the suite passes with payloads that the application rejects: using the same parser configuration, 700 KiB raw succeeds but 800 KiB raw becomes a >1 MiB base64 JSON request and returns 413. Add a near-limit integration test through `registerInstanceProxyRoutes`; remote desktop also needs coverage through `RemoteProxySessionManager` because it has an independent default body limit.

packages/ui/src/components/prompt-input/usePromptAttachments.ts
```diff
@@ -416,53 +420,82 @@ export function usePromptAttachments(options: PromptAttachmentsOptions): PromptA
...
+    const isCurrent = () => !disposed && options.instanceId() === instanceId && options.sessionId() === sessionId
     let skippedCount = 0
+    let tooLargeCount = 0
 
     for (let i = 0; i < files.length; i++) {
       const file = files[i]
-      const nativePath = getFilePath(file)
+      // Device uploads must carry bytes even when the desktop exposes a local path:
+      // that path may not exist on the selected server or in its WSL environment.
+      const nativePath = selectionOptions?.requireData ? null : getFilePath(file)
```
pascalandr: **Blocker:** byte-only selection still needs a batch budget. The input is `multiple`, this loop accepts an unbounded number of files up to 5 MiB each, starts every `FileReader` concurrently, and keeps both raw bytes and a base64 URL. That can consume hundreds of MiB or more before submission. Validate a maximum count and aggregate raw size before starting reads, use bounded/sequential processing, and preserve `FileList` order (the current test sorts the submitted files and therefore hides completion-order reordering). The server must enforce the corresponding encoded request ceiling as well.

pascalandr (commented):
## Gatekeeper verdict: PASS — merge-ready once required CI is green

Reviewed HEAD `611cf1f9ac88dc8c2f1090f8f833637f24ee2679` against `dev` at `ca94224b74f39abab8a65ee9ceb7ec0ba9334642`.

### Product and UX

#725 remains relevant: the workspace browser resolves on the CodeNomad server and cannot represent a file that exists only on the user's device. The implementation keeps the correct host-agnostic design (`<input type="file">` plus byte-backed OpenCode prompt files), without temporary files or Electron/Tauri-specific upload adapters.

The source choice is now understandable in the rendered menu: **Browse workspace files** and **Upload files from this device**. The workspace dialog title follows the same vocabulary, all supported locales are updated, and an inspected real-component capture shows both actions clearly without layout regressions.

### Previous blockers

1. **Transport limit: resolved.** The contract is now 5 MiB per inline file, 10 files, and 20 MiB decoded bytes total. A 32 MiB parser allowance accounts for base64 and JSON overhead. It is scoped to the generated prompt route in the instance proxy and the activated remote-window proxy; unrelated routes retain Fastify's default cap. The server independently validates decoded data-URI size/count/aggregate and rejects malformed data.

2. **Unbounded/concurrent renderer reads: resolved.** Device selection is a dedicated module. It rejects per-file/count/aggregate excess before reads, reads sequentially in picker order, aggregates failures, rechecks actual byte sizes at commit, blocks submission while pending, and discards the complete batch if its session or component becomes stale. Renderer-local paths are never used as a remote fallback.

3. **Coverage seam: resolved.** Browser coverage now tests one honest host-agnostic behavior instead of five runtime-global aliases. Separate integrations exercise an exact-limit body through the real CodeNomad instance proxy and HTTPS remote proxy. The isolated native fixture also sent exactly 5,242,880 bytes through the generated client, CodeNomad proxy, and real OpenCode 2.0.11 prompt endpoint.

### Regression and architecture review

- Workspace/server browsing remains available and sends the expected bytes.
- Empty selections, repeated selection, unreadable files, oversized files, aggregate count, picker ordering, send fencing, and cross-session completion are covered.
- The limits are shared by UI and server; enforcement remains server-authoritative.
- Remote-proxy activation is checked in `onRequest`, before the widened body parser reads the payload.
- The focused selection and budget modules keep the new policy out of the already-large prompt and HTTP modules.
- No blocking correctness, security, architecture, or product findings remain.

### Validation performed

- UI, Electron, and server typechecks: pass.
- UI production build: pass.
- Server suite: 607 pass, 2 expected platform skips.
- Device-upload browser suite: 6/6 pass.
- Remaining browser suite: 93 pass, 1 expected skip; the native browser-frame file then passed 14/14 after restoring the missing local Electron binary.
- Isolated OpenCode 2.0.11 native proxy/file contract: pass at the exact 5 MiB limit.
- `git diff --check`: pass.

GitHub's new workflow run is still queued/in progress at review time. This PASS does not waive required CI; merge only after those checks complete successfully. No merge was performed.

pascalandr:
On review, I think this adds too many overlapping attachment flows. The technical difference between server paths and device uploads does not justify separate neighboring actions in the composer.

I would simplify this to:
- Use the existing @ picker for project files on the server.
- Keep one Attach file action that opens the device's native picker, in both local and remote contexts.
- Remove the separate server file browser from the composer, since it overlaps with @.

I suggest revisiting the UX before merging this PR. I'll leave the implementation as it is for now and defer the next steps.

pascalandr:
Merge conflicts with `dev` have been resolved and pushed in `4391c155` (integrated `dev@d71a5155`). The composer keeps both the device-upload wiring and the current session-aside imports.

I also added a concrete UX harmonization plan in `3c83333c`: **[Composer attachment UX plan](https://github.com/NeuralNomadsAI/CodeNomad/blob/3c83333c/dev-docs/ATTACHMENT_UX_PLAN.md)**, following the direction in https://github.com/NeuralNomadsAI/CodeNomad/pull/729#issuecomment-5751384618.

Proposed UX:
- Existing **`@` picker** for project files/folders on the execution host.
- One **Attach file…** paperclip action opening the device's native picker, consistently across web, Electron, Tauri, local/remote and WSL execution.
- Remove the second Upload from device entry and the composer-specific server browser.
- Share ingestion limits, pending-state feedback and draft/lifecycle fencing across native selection, drop and clipboard files.

The plan explicitly calls out a scope change: `@` is project-scoped, so it does not replace arbitrary outside-project remote filesystem browsing. It also details keyboard/focus behavior, all-locale labels, partial-file failures, session/worktree transitions, persistence, and implementation/test locations. **This push contains the merge resolution and the plan; the proposed UX is not yet implemented.**

Validation after resolving the conflict:
- UI typecheck passed.
- 6 device-upload browser tests passed (including the runtime-context matrix).
- 10 session-aside browser tests passed.
- 10 attachment/submission unit tests passed.
- 58 server attachment-budget and local/remote proxy tests passed.
- `git diff --check` passed.

Size note: the resolved `packages/ui/src/components/prompt-input.tsx` is already 1,182 lines. The plan keeps new ingestion/lifecycle logic out of that component and removes obsolete browser wiring during implementation.

pascalandr:
I completed the first independent gatekeeper pass on the attachment UX implementation. It found four P2 issues:

- A delayed clipboard image could replace a selection made after the paste gesture.
- A failed custom command containing retained image tokens could lose its original draft and consumed pasted-text attachment.
- Finishing a file read could steal focus from a newer target in the same session.
- Failed file reads could keep planned budget reserved and reject a later readable file that actually fits.

I am correcting these cases and adding regressions before another independent review. The PR remains open.

pascalandr:
I fixed the four findings from the first gatekeeper pass and added regressions for each. The second independent pass found one remaining P2: updating the textarea after a delayed image paste could move its caret to the end or collapse a newer selection. I now capture the live selection before the reactive update and map it through the inserted token. I am verifying both middle-of-text typing and a later backward selection before the next review.

pascalandr (commented):
## Independent gatekeeper — final pass

I completed the independent review/correction loop and confirmed **zero actionable findings** on `d69d8e6f`.

The first pass found four P2 issues: delayed paste replacing a newer selection, failed-command draft restoration, read completion stealing focus, and failed reads retaining budget reservations. The second pass found one remaining caret/selection issue after the reactive textarea update. I fixed all five and added regressions. The final reviewer also checked the latest `dev` merge and native-fixture changes.

Validation: 20 attachment browser scenarios, 11 session-aside/custom-command browser tests, 11 attachment/submission/persistence unit tests, server budget/proxy/ownership regressions, UI typecheck/build, Tauri release build, and actual Windows picker selection/cancel/reselection in isolated desktop fixtures passed. Local/remote/WSL runtime coverage is fixture-based, not an end-to-end remote or WSL execution test.

I inspected the narrow French and wide Hebrew captures linked in the PR. The requested single-action UX and outside-project scope change are documented in the updated description.

pascalandr:
I am merging with the requested admin override after the independent gatekeeper reached zero findings and the local checks passed. The final Tauri release build and native selection/cancellation/reselection smoke also passed.

CI run 36059123255 is still in progress. Runtime contracts (minimum/latest), OpenCode compatibility on all three OSes, pruning on Ubuntu/macOS, and Tauri macOS have passed. The main tests, Tauri Windows and pruning Windows are still running; I am not claiming a fully green CI run. The previous run was superseded and cancelled when the PR description was updated.

PR #723, opened 2026-09-20
feat(history): add full-history search, cleanup and global timeline navigation

pascalandr:
## Summary

Combine full-history search/counts and technical cleanup with a global timeline and direct message-window navigation through the existing bundled native pruning plugin.

- **Cleanup:** plan from stored history, including pre-compaction messages, using IDs, content revisions and counts. Execute bounded batches with progress/cancellation and explicit skipped/failed-message reporting.
- **Counts and search:** refreshable history snapshots, entire-session/workspace scope, optional technical content and bounded excerpts. Current-session results jump directly to the selected passage; other-session results open a single-message preview.
- **Global timeline:** lightweight metadata covers the whole session independently of the resident transcript. Distant clicks load at most 200 destination messages without reading intermediate pages. Search clicks, keyboard navigation and missing restoration anchors share the direct loader.
- **Reading position:** retain the visible transcript while loading, let the latest click win, cancel obsolete work on navigation/view changes, overlap adjacent windows and preserve historical reading during live generation.
- **Standard native scrollbars:** timeline and transcript share standard geometry. Only marker rectangles narrow; icon dimensions and vertical spacing stay fixed. Timeline extent is calculated from measured marker/gap primitives rather than estimated hidden rows.
- **Session return:** retain four recent metadata-only outlines across view unmounts; interrupted scans resume their accepted cursor. Reuse the completed rail immediately while refreshing, and expose progress during the initial scan. Search excludes tools/reasoning by default.
- Keep search non-modal, preserve keyboard focus and translate new visible strings in every supported locale.
- Keep unchanged timeline marker identities stable during streaming; manual rail browsing cancels a pending automatic reveal, so a second destination remains clickable during a slow jump.

## Implementation and boundaries

- Dedicated CodeNomad routes expose fixed `history`, `pruneBatch`, `outline` and `window` operations. Generic RPC proxy access remains closed.
- A shared helper binds read-only SQLite access to the authenticated daemon through a fresh plugin-storage challenge. Ownership includes native directory/project/legacy workspace identity; navigation rechecks session location, project, revert and workspace authorization before publication.
- History queries scan at most 32 rows per page; outlines return up to 256 compact entries with a sequence horizon. Navigation reads yield every 16 rows and bound parsed page payloads to 24 MiB / 16 MiB per message. Scheduler waits no longer fragment an outline into tiny pages.
[...]

PR #730, opened 2026-09-20
fix(ui): read edit tool diff from OpenCode 2.x metadata.files

markerikson:
Since OpenCode 2.x, `edit` tool calls render as just "Edited path (1 replacement)" with no diff, and the call title is a bare "Edit".

OpenCode 2.x's edit tool changed its shape: the input is `{ path, oldString, newString }` (not `filePath`), and the diff lives at `metadata.files[]` as `FileDiff` entries (`{ file, patch, additions, deletions, status }`) instead of a `metadata.diff` string.

- `extractDiffPayload` now also accepts `metadata.files[0].patch` as the diff and `metadata.files[0].file` as the path. The V1 `metadata.diff` key still takes priority so older servers keep working.
- `editRenderer.getTitle` falls back to `input.path`.
- Adds `diff-payload.test.ts` (extraction for both shapes) and `renderers/edit.test.ts` (title), both wired into the force-exit CI test list.

This PR is rendering only. Making V2 edit diffs searchable in the server-side history search is a separate change: #735.


packages/ui/src/components/tool-call/search-text.ts
```diff
@@ -99,6 +100,7 @@ export function getDiffToolSearchText(context: ToolSearchTextContext): string[]
   appendString(values, input.filePath)
   appendString(values, input.path)
   appendString(values, metadata.diff)
+  appendString(values, extractDiffPayload(context.toolName, context.toolState)?.diffText)
```
pascalandr: **Blocking — avoid indexing the same legacy diff twice.** `extractDiffPayload()` gives `metadata.diff` first priority, so for a V1 edit this line appends the exact value already appended on line 102. `extractToolText()` joins both values, and `buildSessionSearchMatches()` counts both copies; a query that occurs once in the patch therefore produces two search results (I reproduced `needleCopies=2`). The same duplication can occur when the helper falls back to `output`, which is appended again below. Please select one canonical diff value or deduplicate the collected values, and add V1/V2 search assertions so both formats yield one hit per real occurrence.

packages/ui/src/components/tool-call/renderers/edit.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    status: "completed",
+    input: { filePath: "src/example.ts" },
+    metadata: { diff: patch },
+    output: "",
+  } as never)
+  assert.equal(payload?.diffText, patch)
+  assert.equal(payload?.filePath, "src/example.ts")
+})
+
+it("renders the V2 edit diff and titles by input.path", () => {
```
pascalandr: **Blocking — make this regression test runnable in CI.** The `Test changed runnable UI behavior` step enumerates test files explicitly, and `edit.test.ts` is not in that list, so the green `tests` job never executed these assertions. Running this file with the same base command (`node --import tsx --test ...`) prints all three passes but keeps the process alive indefinitely because this renderer import leaves an open handle; it only completed locally with `--test-force-exit`. Please wire the coverage into the CI list and structure it so the existing runner exits normally (for example, keep pure extraction/search coverage outside the renderer import and use an appropriate terminating renderer fixture).

pascalandr (changes requested):
CHANGES_REQUESTED

The V2 `metadata.files[]` extraction and `input.path` title fallback are directionally correct, and the UI typecheck/build pass. Two regressions still block this HEAD:

1. V1 diff content is indexed twice, producing duplicate session-search hits.
2. The new regression test is not part of the explicit CI test list and does not terminate under the repository's normal `node --import tsx --test` invocation.

The Windows Tauri failure appears unrelated to this change (two cross-host timing/concurrency tests failed while the other 178 crate tests passed), but the required workflow is still red and the build was skipped. No merge performed.

packages/ui/src/components/tool-call/search-text.ts
```diff
@@ -99,6 +100,7 @@ export function getDiffToolSearchText(context: ToolSearchTextContext): string[]
   appendString(values, input.filePath)
   appendString(values, input.path)
   appendString(values, metadata.diff)
+  appendString(values, readFirstFileDiff(metadata)?.patch)
```
pascalandr: **[P1] Exercise the production history-search seam**

After #723, `MessageSection` calls `findHistoryMatches()` → `/session-history/query`; `buildSessionSearchMatches()` has no production callers, so this renderer search hook cannot make V2 edit diffs searchable for users. I inserted one native V2 edit whose unique token existed only in `state.metadata.files[0].patch`: `queryHistoryPage()` scanned the message and returned `hits: []`, while this new unit test remained green. Please either teach the bounded native `history-store.ts` search to read the safe V2 metadata shape and cover that path in `history.test.ts`, or drop this now-ineffective search change/claim. If this helper remains, also deduplicate it: `getApplyPatchToolSearchText()` calls this function and then appends every `files[].patch` again, so its first patch currently produces two occurrences (locally reproduced with `count: 2`).

pascalandr (changes requested):
Second gatekeeper pass on `063f027d`. The two original findings are fixed: V1 edit diffs are no longer duplicated, both new suites run and terminate in the explicit CI step, and their 5 focused assertions pass locally. I resolved those two old threads.

One integration issue still blocks this head. Since #723, production session search no longer calls `buildSessionSearchMatches()` or renderer `getSearchText()` hooks; it uses the bounded native history query. The new search assertion therefore exercises an orphaned client projection rather than the user path, while the native query still ignores `state.metadata.files`. It also makes the retained `apply_patch` search helper append its first patch twice. Please either move the V2 metadata search behavior and coverage to the bounded history implementation, or remove this ineffective search change/claim and keep this PR scoped to rendering.

The branch is also 20 `dev` commits behind. Its current red jobs are caused by the head-only CI checkout missing files required by the newer base workflow, so refresh against current `dev` after the code correction and rerun the full matrix. No merge performed.

packages/ui/src/components/tool-call/search-text.ts
pascalandr: Current HEAD `f4eddee3` correctly removes the orphaned client-search change (and therefore the remaining `apply_patch` duplication), so the code side of this finding is addressed. The PR description still says “Search text indexes the V2 diff so message search still matches edited content,” which remains false for the production bounded history search. Please remove or correct that claim; then I can resolve this thread.

packages/ui/src/components/tool-call/utils.ts
```diff
@@ -124,7 +124,8 @@ export function extractDiffPayload(toolName: string, state?: ToolState): DiffPay
   if (!diffCapableTools.has(toolName)) return null
 
   const { metadata, input, output } = readToolStatePayload(state)
-  const candidates = [metadata.diff, output, metadata.output]
+  const fileDiff = readFirstFileDiff(metadata)
+  const candidates = [metadata.diff, fileDiff?.patch, output, metadata.output]
```
pascalandr, line 128: **Blocking — do not truncate multi-file V2 `patch` calls.** `extractDiffPayload()` is shared by both `edit` and `patch`, so adding `metadata.files[0].patch` as a generic candidate changes valid V2 patch output: `patchRenderer` now renders and copies only the first `FileDiff`, silently hiding every later file. OpenCode V2 `patch` emits the complete `metadata.files[]` array. Please either restrict this fallback to `toolName === "edit"` or teach the patch renderer to present every file, and add a two-file V2 patch regression test.

packages/ui/src/components/tool-call/renderers/edit.tsx
```diff
@@ -11,7 +11,8 @@ export const editRenderer: ToolRenderer = {
     const state = toolState()
     if (!state) return undefined
     const { input } = readToolStatePayload(state)
-    const filePath = typeof input.filePath === "string" ? input.filePath : ""
+    const filePath =
+      typeof input.filePath === "string" ? input.filePath : typeof input.path === "string" ? input.path : ""
```
pascalandr, line 15: **Test gap — verify Windows paths by exact title.** V2 `input.path` may be an absolute Windows path, but `getRelativePath()` currently splits only on `/`. The new `/example\.ts$/` assertion still passes when the UI leaks the full `C:\\repo\\src\\example.ts`. Please handle both separators and assert the exact rendered title.

pascalandr (changes requested):
Fresh gatekeeper pass over all three commits. The V1/V2 edit rendering change and its focused tests pass, and the obsolete client-search change is gone. One product regression still blocks this HEAD: the shared extraction helper now truncates valid multi-file V2 patch calls to their first file. I also left a path-handling test note below. No merge performed.

markerikson:
Addressed in `d876b417`:

- **Multi-file patch:** `readFirstFileDiff` is now `readSingleFileDiff` and only returns a diff when `metadata.files` has exactly one entry. A multi-file V2 `patch` therefore gets `null` from `extractDiffPayload` and the patch renderer falls through to its existing `metadata.diff` / output fallback, same as before this PR, instead of showing only the first file. I went with this rather than teaching the patch renderer to show every file because `parsePatchToBeforeAfter` in the diff viewer is single-file; a proper multi-file view is a bigger change than this fix. Added a two-file patch regression test plus a single-file patch test.
- **Windows paths:** `getRelativePath` splits on both `/` and `\\`. The edit renderer test now asserts the exact title `Edit example.ts` for both `src/example.ts` and `C:\\repo\\src\\example.ts`.

Both new tests fail on the previous head and pass now; UI typecheck clean.

pascalandr (approved):
Thanks @markerikson !

PR #731, opened 2026-09-20
fix(ui): refresh session usage info on step and usage events

markerikson:
The Status tab's INPUT / OUTPUT / COST chips show 0 / 0 / $0.00 for sessions that were empty when opened, and never update while a session runs.

The Status tab reads `sessionInfoByInstance`, which is only written by `updateSessionInfo()`. The V2 migration (#647) removed the calls to it from the live SSE handlers, leaving only the message-load and agent/model-switch paths. `updateSessionInfo()` also summed usage from resident messages only, which in V2 is a bounded window rather than the whole session.

The displayed totals are now the session's **lifetime usage**, taken from the server's cumulative `Session.cost` / `Session.tokens` counters:

- `updateSessionInfo()` uses `session.cost` / `session.tokens` whenever the server reports them, independent of which messages are loaded. Message sums only stand in when the server sends no session usage.
- It is called from the `session.usage.updated` handler (outside the active-transcript gate, so inactive sessions update too), after `session.step.ended` / `session.step.failed` projection, and on load for an already-hydrated session that has no info yet.
- Reverts do not change the displayed totals. OpenCode's projector does not decrement session usage on `session.revert.committed`, so subtracting client-side was only correct for messages that happened to be resident and was lost on reload. Lifetime totals are the same number in every window and after every reload.

`stores/session-usage-contract.test.ts` (added to the force-exit CI list) covers: active step completion, usage-only event with an empty transcript, inactive session, initial hydration from the session record, authoritative totals over a partial window, revert with resident messages, revert from an anchored historical window with the newer message not resident, and post-reload hydration after a revert.


packages/ui/src/stores/instances.ts
```diff
@@ -2021,13 +2029,16 @@ function handleInstanceInvalidation(instanceId: string, event: Parameters<NonNul
...
-    )
+    )) {
+      projectOpenCodeMessages(
+        instanceId,
+        sessionId,
+        data,
+        preserveOmitted,
+        force || event.type !== "session.inbox.enqueued",
+      )
+      if (force || USAGE_EVENT_TYPES.has(event.type)) updateSessionInfo(instanceId, sessionId)
```
pascalandr, line 2039: **Blocking — `session.usage.updated` cannot update the displayed totals from here.** The installed OpenCode reducer handles this event by updating `data.session.info[sessionID].cost/tokens`; it does not write usage onto an assistant message. `projectOpenCodeMessages()` only projects messages, while `updateSessionInfo()` reads only `messageStoreBus` usage and ignores both `event.data` and the native `Session.cost/tokens`. In addition, `SSEManager` invokes `onInvalidation` before `handleNativeSessionEvent` updates the CodeNomad session, and this entire call is gated to the active session with a latest window. Consequently this branch is a no-op for a usage-only event and never repairs initial or inactive-session totals. Please apply the authoritative usage payload at a session-usage seam outside the active transcript gate (with identity/generation fencing), and initialize the same projection from hydrated `SessionInfo.cost/tokens`.

pascalandr (changes requested):
CHANGES_REQUESTED

The active `session.step.ended` / `session.step.failed` projection can refresh the chips, but this does not yet restore the complete V2 usage contract:

1. `session.usage.updated` carries the authoritative cumulative `cost`/`tokens`, yet the new path does not consume that payload and cannot update inactive sessions or initial hydration. This is the same first-load gap explicitly noted in the PR description.
2. `session.revert.committed` removes normalized messages after the new refresh point, so `sessionInfoByInstance` and thread totals retain the removed usage.
3. No regression test was added for this central event-ordering path. Please cover active step completion, usage-only/inactive or initial hydration, and a usage-decreasing revert in one of the runnable store suites (for example `runtime-contract.test.ts`).

I reproduced both stale-state cases locally: an authoritative session with cost `2.5` and tokens `1200/300` still published `0/0/$0`, and removing a `10/5/$1` assistant message cleared the message-store usage while the displayed session info remained `10/5/$1`.

The test/typecheck jobs and completed builds are green; the Tauri Windows and macOS x64 builds were still running when this review was submitted. No merge performed.

packages/ui/src/stores/message-v2/session-info.ts
```diff
@@ -40,7 +40,16 @@ export function updateSessionInfo(instanceId: string, sessionId: string): void {
   let contextAvailableFromPrevious = false
   let isSubscriptionModel = false
 
-  if (!hasUsageEntries && previousInfo) {
+  // Message-derived totals win when the transcript is loaded, since they drop
+  // after a revert while the server's session counters only ever grow. The
+  // session record fills in for usage-only events and unloaded transcripts.
+  if (!hasUsageEntries && session.tokens) {
```
pascalandr: **[P1] Keep complete session totals independent of the resident message window.** `hasUsageEntries` only proves that at least one usage-bearing assistant message is currently normalized; the V2 transcript is bounded, so it does not prove that every message contributing to `Session.cost/tokens` is resident. For a long session whose latest window contains (say) `$1` while `session.usage.updated` reports `$20`, this condition keeps publishing `$1` and discards the authoritative event payload. The inactive test has an empty store, so it misses this case.

This fallback also breaks a revert of the first/only message: once removal empties `usage.entries`, the branch restores `session.tokens`/`session.cost`, but OpenCode `v2.0.11`'s `RevertEvent.Committed` projector deletes messages without decrementing `SessionTable` usage. The new revert test leaves `m1` resident and therefore does not exercise that boundary.

Please preserve authoritative session totals independently of transcript residency and represent the revert adjustment separately (rather than making a partial window authoritative). Add regressions where native totals exceed the one resident message and where the committed revert removes the only/first message.

pascalandr (changes requested):
CHANGES_REQUESTED on `10858125`.

The new event ordering, inactive-session path, hydration fallback, and runnable regression file address much of the earlier review. One correctness gap still leaves the Status totals dependent on the bounded resident transcript: any single loaded usage entry wins over the authoritative session totals, and an all-message revert falls back to the same non-decrementing native counters. Please address the inline P1 and cover both boundaries. No merge performed.

markerikson:
Addressed in `489e8a9c`:

- `session-info.ts`: when the session record carries `tokens`, the server's cumulative `cost`/`tokens` are the base totals regardless of which messages are resident. Message sums only stand in when the server reports no session usage.
- Reverts are tracked as a separate adjustment (`recordRevertedUsage`): the usage entries of the removed messages are recorded before removal and subtracted from the authoritative totals, clamped at zero. A later `session.usage.updated` with the same non-decremented counters keeps the adjustment. Cleared on instance removal.
- New tests in `session-usage-contract.test.ts`: native totals $20 with one resident $1 message shows $20; revert of the only message shows 0; revert followed by an unchanged `usage.updated` stays lowered; message-sum fallback when no session usage exists. Both new boundary cases fail without the change; 13/13 pass with it.

Known limit: `session.revert.committed` carries only `{ sessionID, to }`, so the adjustment can count only reverted messages that were resident in the loaded window, and it is in-memory (a reload shows the server's lifetime counters until the next revert). Both would need the server to either decrement its counters or include the removed usage in the event.

packages/ui/src/stores/instances.ts
```diff
@@ -2021,23 +2029,29 @@ function handleInstanceInvalidation(instanceId: string, event: Parameters<NonNul
...
     projectMessages(data, preserveOmitted)
     if (sessionId && event.type === "session.inbox.cancelled") {
       removeMessageV2(instanceId, event.data.inboxID, sessionId)
     }
     if (sessionId && event.type === "session.revert.committed") {
-      for (const messageId of messageStoreBus.getOrCreate(instanceId).getSessionMessageIds(sessionId)) {
-        if (messageId >= event.data.to) removeMessageV2(instanceId, messageId, sessionId)
-      }
+      const store = messageStoreBus.getOrCreate(instanceId)
+      const removed = store.getSessionMessageIds(sessionId).filter((messageId) => messageId >= event.data.to)
```
pascalandr: **Blocking — this only records resident usage.** `getSessionMessageIds()` is the bounded transcript window, not the full session. When the user navigates to an older anchored window and reverts from there, newer assistant messages deleted by the native commit can be outside this list, so their usage is never added to the adjustment. I reproduced `$5` native total with only the `$1` boundary resident: the empty post-revert transcript still displays `$4`. The adjustment source must cover the complete reverted range, independently of the mounted window.

packages/ui/src/stores/message-v2/session-info.ts
```diff
@@ -2,7 +2,46 @@ import type { Provider } from "../../types/session"
...
+  cost: number
+  inputTokens: number
+  outputTokens: number
+  reasoningTokens: number
+}
+
+// OpenCode's session usage counters only ever grow; a committed revert deletes
+// messages without decrementing them. Track the reverted usage separately so
+// the displayed totals can subtract it from the authoritative session totals.
+const revertedUsage = new Map<string, UsageTotals>()
```
pascalandr: **Blocking — this authority disappears on renderer reload.** OpenCode persists cumulative lifetime counters and the committed messages are already deleted, so this module-local map cannot be reconstructed during hydration. Reverting the only `10 / 5 / $1` message displays zero, then the same session displays `10 / 5 / $1` after reload. Either keep lifetime totals consistently (including immediately after revert) or make the current-transcript adjustment complete and durable/reconstructible.

pascalandr (changes requested):
CHANGES_REQUESTED on `489e8a9c`.

The two boundaries from the previous review are fixed, and the submitted usage tests pass. However, the acknowledged limitations are normal CodeNomad flows and leave the same displayed totals with different values depending on transcript residency and renderer lifetime:

1. **Bounded historical window:** with native totals `50 / 25 / $5`, resident boundary message `m1 = 10 / 5 / $1`, and newer `m2 = 40 / 20 / $4` outside the resident window, committing the revert to `m1` leaves `40 / 20 / $4` instead of zero. `removed` can only enumerate resident records.
2. **Reload after revert:** reverting the only `10 / 5 / $1` message shows zero until the renderer state is rebuilt; hydration then restores `10 / 5 / $1` from OpenCode's persisted cumulative counters because the module-local adjustment is gone.

OpenCode 2.0.11 deleting messages without decrementing `SessionInfo.cost/tokens` explains the constraint, but it does not make the inconsistent result safe to ship. Please choose one stable contract: either lifetime usage (never subtract on revert) or current-transcript usage backed by complete, reload-reconstructible authority. Add regressions for a revert from an anchored historical window and post-reload hydration.

Local validation: UI typecheck passed; 81/81 relevant native-event, request-authority, and usage tests passed; both additional boundary reproductions failed as described. No merge performed.

markerikson:
Switched to the lifetime-usage contract in `761968a2`.

- `updateSessionInfo()` now uses `session.cost` / `session.tokens` as-is whenever the server reports them. The client-side revert adjustment (`revertedUsage` map and the revert-handler bookkeeping) is removed, so the displayed number no longer depends on which messages are resident or on renderer lifetime.
- Both reproductions from the review are now tests in `session-usage-contract.test.ts`: a revert from an anchored window with only the $1 boundary message resident keeps showing the $5 native total, and reverting the only message then re-hydrating in a fresh setup shows the same totals before and after.
- Message sums still stand in only when the server sends no session usage at all.

PR description updated to match. UI typecheck clean; the usage, runtime-contract, and native-events suites pass (22/22) under the force-exit command.

packages/ui/src/stores/session-usage-contract.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    const base = { sessionID: sessionId, assistantMessageID: "m1" }
+    first.emit("session.step.started", { ...base, agent: "build", model, started: 1 }, 1)
+    first.emit("session.step.ended", { ...base, finish: "stop", cost: 1, tokens: tokens(10, 5) }, 2)
+    first.emit("session.usage.updated", { sessionID: sessionId, cost: 1, tokens: tokens(10, 5) }, 3)
+    first.emit("session.revert.committed", { sessionID: sessionId, to: "m1" }, 4)
+    assertUsage(instanceId, sessionId, { cost: 1, input: 10, output: 5 })
+  } finally { first.cleanup() }
+
+  // A fresh renderer only has the persisted session record and an empty transcript.
+  const reloaded = setup(instanceId, sessionId, { session: { cost: 1, tokens: tokens(10, 5) } })
```
pascalandr: **Blocking test gap — this is not a fresh hydration.** `first.cleanup()` calls `removeInstance()`, but that path does not remove this session from `sessionInfoByInstance`. Reusing the same instance/session IDs means `loadMessages()` sees the existing info and skips `updateSessionInfo()`, so the assertion reads the pre-reload value rather than `options.session`. I verified this by changing the second record to `90 / 45 / $9`: the test still observed the old `10 / 5 / $1`. Please clear the session-info authority or use a genuinely fresh identity/reset seam, and make the fixture distinguish the persisted value so the test proves hydration actually ran.

pascalandr (changes requested):
The new lifetime-usage contract resolves the two production inconsistencies from the prior review: anchored-window reverts and renderer reloads now intentionally retain the same native cumulative totals. One focused regression-test flaw remains before signoff; the current reload case never discards the first in-memory session-info value. The known system-message timeouts are ignored per project-owner direction.

markerikson:
Fixed the reload test in `16d70b9f`: the re-hydration now runs under a fresh instance id (`removeInstance()` leaves `sessionInfoByInstance` entries in place), asserts the info is absent and the transcript empty before `loadMessages()`, then checks the hydrated record. I confirmed it now tracks the persisted record by temporarily changing the second record to 90 / 45 / $9: the test fails with those values and passes with the matching ones.

pascalandr (approved):
Thanks @markerikson !

PR #733, opened 2026-09-20
fix(ui): update tool visibility settings for OpenCode 2.x tool names

markerikson:
The per-tool "Tool call expansion" settings still list OpenCode V1 tool names. OpenCode 2.x renamed `bash` → `shell` and `task` → `subagent`, retired `apply_patch` and `todowrite`, and added `websearch` and `execute`. Under 2.x, shell and subagent calls fell into the "Other tools" bucket and used the default renderer, so their expansion setting could not be controlled separately.

- Alias `shell` and `subagent` to the existing `bash` and `task` registry entries. Stored preference keys are unchanged, so existing per-tool settings keep applying, and the bash/task renderers are used for these calls again.
- Add `websearch` and `execute` as configurable tools.
- Hide `apply_patch` and `todowrite` from the settings list. Their registry entries remain so older sessions render as before. `patch` stays configurable: OpenCode 2.x still exposes it for supported GPT models.
- Task renderer reads the V2 `agent` input alongside V1 `subagent_type`, and the task search text includes it.
- Icons for `execute`, `websearch`, and `subagent`.
- Browser fixtures that rely on the shell tool call starting collapsed now set that expansion preference explicitly (`tests/browser/fixtures/ui-settings.ts`), since aliasing `shell` to `bash` changes its balanced-preset default to expanded.

Adds `tool-presentation.test.ts` covering the alias routing and the configurable list, wired into the CI test list.

(side note: I'd love to have a config option to control expanding _input_ vs _output_ separately on a per-tool basis - I often want to see the inputs for some tools like `execute`, but not for others.)


packages/ui/src/components/tool-call/tool-presentation.ts
```diff
@@ -58,13 +59,13 @@ export const TOOL_REGISTRY: ToolRegistryEntry[] = [
   {
     tool: "patch",
     label: "patch",
-    configurable: true,
+    configurable: false,
```
pascalandr: **[P1] Keep the active V2 `patch` tool configurable**

OpenCode V2 has not removed `patch`: the official V2 tools reference explicitly says it is exposed for supported GPT models, and the pinned `v2.0.11` runtime registers `opencode.tool.patch`. Setting this to `false` removes `patch` from both Chat settings and the transcript visibility popup, so current GPT sessions can no longer choose a per-tool visibility policy and instead inherit fallback/custom behavior. Please keep `patch` configurable; only the retired `apply_patch` / `todowrite` entries should be hidden, and the new assertion should reflect that.

packages/ui/src/components/tool-call/tool-presentation.ts
```diff
@@ -88,12 +95,19 @@ export const TOOL_REGISTRY: ToolRegistryEntry[] = [
...
     label: "todowrite",
-    configurable: true,
+    configurable: false,
     expansionPresets: presets({ minimal: expanded, balanced: expanded, detailed: expanded, everything: expanded }),
   },
   {
     tool: "task",
-    label: "task",
+    label: "subagent",
+    aliases: ["subagent"],
```
pascalandr, line 104: **[P2] Include the V2 agent name in transcript search**

This alias also selects `taskRenderer.getSearchText`, but `getTaskToolSearchText()` still appends only the V1 `input.subagent_type`. For a V2 call such as `{ agent: "explore", prompt: "Inspect it" }`, the search projection contains `subagent` and the prompt but not `explore`, so searching by the chosen agent no longer finds the call. Please append `input.agent` (while retaining the V1 field for old transcripts) and cover that path.

packages/ui/src/components/tool-call/tool-presentation.ts
```diff
@@ -33,7 +33,8 @@ function presets(values: Record<ToolCallExpansionPreset, VisibilityPreference>)
 export const TOOL_REGISTRY: ToolRegistryEntry[] = [
   {
     tool: "bash",
-    label: "bash",
+    label: "shell",
+    aliases: ["shell"],
```
pascalandr, line 37: **[P1] Restore the shell browser regression suite**

This alias intentionally moves `shell` from the `Other tools` balanced default (collapsed) to the `bash` default (expanded). Five browser regressions still click the header assuming they are opening it; they now close it and time out waiting for `.tool-call-markdown` (`session-rendering.test.ts:229`, `:596`, `:615` twice, and `:636`). I reproduced the first timeout locally. Please make those fixtures/tests establish the required state explicitly rather than assuming the initial expansion, then rerun the full browser suite.

packages/ui/src/components/tool-call/tool-presentation.test.ts
```diff
@@ -0,0 +1,25 @@
+import assert from "node:assert/strict"
+import { describe, it } from "node:test"
+
+import { getConfigurableToolEntries, getToolRegistryEntry, OTHER_TOOL_NAME } from "./tool-presentation"
+
+describe("tool presentation registry", () => {
```
pascalandr, line 6: **[P2] Wire this regression test into CI**

`.github/workflows/pr-build.yml` runs source-level UI tests through an explicit `node --test` path list, and this new file is absent from it. The green `Test changed runnable UI behavior` step therefore did not execute these assertions. Please add this path to that list (or another CI-owned test entrypoint) so the registry contract is actually enforced.

pascalandr (changes requested):
Gatekeeper verdict: changes requested on `624d8a81`.

The alias direction and preservation of the existing `bash` / `task` preference keys are sensible, but this head is not merge-ready:

- OpenCode V2 still exposes `patch`, so the PR removes a setting for an active V2 tool.
- V2 `subagent` calls lose `input.agent` from transcript search.
- The browser suite is red: 5 shell-output regressions time out after the alias changes the default expansion state.
- The new registry test is not wired into the workflow's explicit Node test list.

Please address the inline findings, refresh against current `dev` (`92426ebe` at review time; the merge tree is currently clean), and rerun the complete browser suite. Validation performed: official V2 tools reference plus pinned `v2.0.11` source inspection; focused registry test (3/3 pass); focused browser reproduction (same timeout at `session-rendering.test.ts:229`); CI log review (96 pass, 5 fail, 1 skipped in the browser step).

packages/ui/src/components/tool-call/tool-presentation.ts
pascalandr: The code and registry assertion on current HEAD `7e10a259` now correctly keep `patch` configurable. The PR description still says OpenCode 2.x removed `patch` and that this PR hides it; please update those two statements to match the corrected behavior before this thread is resolved.

packages/ui/src/components/tool-call/tool-presentation.ts
```diff
@@ -64,7 +65,7 @@ export const TOOL_REGISTRY: ToolRegistryEntry[] = [
   {
     tool: "apply_patch",
     label: "apply_patch",
-    configurable: true,
+    configurable: false,
```
pascalandr, line 68: **[P1] Preserve hidden legacy-tool overrides.** Changing these entries to non-configurable removes them from `getConfigurableToolEntries()`, but `transcriptVisibilityPatch()` rebuilds the complete custom `tools` map from that filtered list. Editing any visible row therefore deletes persisted `apply_patch` / `todowrite` overrides; old transcripts then inherit `other` and can change presentation with no UI path to restore the setting. I reproduced `{ apply_patch: hidden, todowrite: expanded, read: collapsed, other: collapsed }` becoming collapsed for both hidden tools after changing `read`. Please preserve hidden registered keys/effective values during the custom-mode transition and add a regression test.

packages/ui/src/components/tool-call/tool-presentation.ts
pascalandr, line 104: **[P1] Apply the alias to every task-specific consumer, not only renderer lookup.** `resolveToolRenderer("subagent")` now selects `taskRenderer`, but `tool-call.tsx` and task-summary title logic still branch on the raw name being exactly `"task"`. A native V2 `subagent` call can consequently render a mixed header such as `subagent Task[explore] Inspect it`, let a generic state title override the task title, or describe nested summaries generically. Please expose/use the canonical registry tool for semantic comparisons and cover a V2 `subagent` title/summary path.

pascalandr (changes requested):
Focused gatekeeper pass on the final two-commit diff. The original `patch`, shell-fixture, and CI-wiring issues are fixed, and the transient timeline failure is not attributable to this PR. Two alias/migration regressions remain: hiding legacy rows drops their persisted overrides on the next visible edit, and `subagent` does not receive all of the `task` title semantics. The known system-message timeouts are ignored per project-owner direction.

markerikson:
Both addressed in `891dcdbb` (rebased on current dev, which now includes #730):

- **Hidden legacy overrides:** `transcriptVisibilityPatch` now builds the custom `tools` map from every registered entry (new `getRegisteredToolEntries()`), not just the configurable ones, so `apply_patch` / `todowrite` overrides survive edits to visible rows. Regression test in `transcript-visibility.test.ts` uses your exact repro (`apply_patch: hidden, todowrite: expanded, read: collapsed, other: collapsed`, then change `read`). That test file is now also in the plain CI list.
- **Alias semantics:** added `getCanonicalToolName()` to the registry. `tool-call.tsx` uses it for the task-vs-generic title branch, and the header label stripping (moved to a pure `getToolTitleDetail()` in `utils.ts`) accepts both the raw name and the canonical tool as prefixes. `task.tsx` uses it in `describeToolTitle` for nested summaries, and `describeTaskTitle` takes the raw tool name so a V2 call titles as `Subagent[explore] Inspect it` and the header shows `subagent [explore] Inspect it`. The title helpers moved to a JSX-free `renderers/task-title.ts` with tests for the V1 and V2 paths (in the force-exit CI list, since `utils.ts` pulls in i18n).

UI typecheck clean; 21/21 across the presentation, visibility, task-title, and edit/diff tests.

pascalandr (approved):
And 4/4 , thanks a lot @markerikson !

PR #735, opened 2026-09-20
fix(history): search edit diffs stored in tool metadata

markerikson:
## Problem

Session search (`/session-history/query`, `history-store.ts`) scans tool parts via their name, input, error and text content. Edit-style tools (`edit`, `patch`, `write`) keep their diff in metadata instead: `metadata.diff` in OpenCode 1.x transcripts and `metadata.files[].patch` in 2.x. A word that only appears in a diff is not searchable.

Split out of #730 per review: the client-side `search-text.ts` projection has no production callers since #723, so the search behavior belongs here.

## Change

- `findHit()` also scans string values from `state.metadata.diff` and `state.metadata.files[].patch`. Only those two known shapes are read; other metadata stays out of the scan.
- `history.test.ts`: new case inserting a 2.x edit whose token exists only in the patch and a 1.x edit with `metadata.diff`; both hit once with `kind: "tool"`, and neither hits with `includeTechnical: false`.

## Verification

- `node --import tsx --test packages/server/src/opencode/session-pruning/history.test.ts` → 6 pass (new test fails without the `history-store.ts` change).
- `npm run typecheck --workspace @neuralnomads/codenomad` → clean.
- Covered by the existing `packages/server/src/**/*.test.ts` CI glob.


pascalandr (approved):
Thanks @markerikson !

PR #744, opened 2026-09-22
fix(settings): restore inline OpenCode executable management

pascalandr:
## Summary
- Restore the executable selector as the first OpenCode settings card: path entry, browsing and saved binaries are immediately available.
- Show runtime status, updates and service actions inline; remove the separate management-modal entry point from settings.
- Keep only Version details and Troubleshooting collapsed below the runtime controls. Log levels remain the last card.
- Rename Recommended and tested version to Minimum recommended version in all ten locales.
- Reuse the controls in startup recovery and document the intended layout.

## Validation
- All 12 OpenCode setup browser scenarios validated (full run plus focused rerun of the adjusted layout assertion).
- UI typecheck and diff checks passed.
- Real-component screenshots inspected at desktop and French narrow widths.

## Maintenance note
The existing locale settings.ts files touched by the two label changes are approximately 620–635 lines each.

Gatekeeper review will be posted separately against the published commit.


pascalandr (commented):
## Gatekeeper review — PASS
Reviewed commit: a4efa86f6614f101024ad871f399391affbc6b49, against base 7fc21f8c.

An independent read-only gatekeeper review found no actionable regressions. The review loop therefore ends with no remaining findings on this head.

### Reviewed
- Executable path entry, browsing and saved selection are directly available in the first settings card.
- Runtime state and install/update/connect/restart actions are inline through the shared setup panel.
- Version details and Troubleshooting are the only runtime disclosures, collapsed by default after actions and feedback.
- Log level remains the final settings card.
- Minimum recommended version and Version details labels cover all ten registered locales.
- Checked caller/store integration: settings and native preferences retain executable-selection persistence, status invalidation, shared busy guards and explicit service mutations.
- Browser regression coverage includes inline binary changes, disclosure defaults, card order, delayed checks/reloads, error recovery, optional updates and French narrow-width layout.

### Evidence and limits
Implementation validation: 12 browser scenarios passed across the full run and focused assertion rerun; UI typecheck and diff checks passed; desktop/narrow captures inspected. The separately launched preview of this exact implementation loaded without browser page errors and showed the expected French layout with simulated data.

The independent gatekeeper pass was code review and did not rerun those checks. The deterministic browser fixture exercises the real settings components rather than the complete settings shell; native file dialogs, full-shell remount navigation and every locale's rendered layout are not covered by this fixture. No claim of green full CI is made here.

No merge performed: this request is to publish the PR, iterate gatekeeper review until clean, and show the preview.

PR #749, opened 2026-09-22
fix(auth): recover browser sessions after server restart

### LINKED ISSUE #509: better message for try open project when CodeNomad is restart but page didn't F5
better message for try open project when CodeNomad is restart but page didn't F5

- start CodeNomad 
- open CodeNomad page
- close or restart CodeNomad 
- try open project

Image

pascalandr:
Fixes #509.

I added an in-page login form when CodeNomad confirms that the browser session is no longer valid. API failures and event-stream disconnects trigger a bounded auth-status check; upstream OpenCode 401s and network outages do not trigger a misleading login prompt. Successful login reconnects events without reloading the page or replaying failed actions.

I also preserve live workspace drafts before restart reconciliation removes old backend workspace IDs. Reopening the project restores that state, including on a fresh page with no startup snapshot. Generic alerts wait behind authentication recovery without losing their pending state.

### Validation

- Four Chromium scenarios using real CodeNomad auth routes and replaced in-memory session managers: restart/SSE recovery, wrong credentials, draft and attachment restoration after workspace removal/reopening, late alerts, upstream 401s, outages, another-tab login, narrow screens and no mutation replay.
- 47 targeted unit regressions for auth races, SDK transport, event recovery, drafts, snapshot preservation and instance runtime authority.
- UI typecheck, production UI build and diff checks pass; rendered narrow-screen capture inspected.

### Review

The first independent gatekeeper review found two P1 issues: restart reconciliation could discard fresh-page drafts, and a late alert could block the login form. I fixed both and added regressions. The second pass reported zero actionable findings and independently passed the four browser scenarios and thirteen focused unit tests.


pascalandr:
I checked the failed CI jobs against the pre-PR code. Both system-message search failures reproduce on the exact base commit, and the same failures appear in the earlier run https://github.com/NeuralNomadsAI/CodeNomad/actions/runs/35753233159. That run also has the same native migration assertion failure: OpenCode 2.0.14 adds `method: "key"` to credential metadata.

All four new auth-recovery browser scenarios passed in the full CI run. The minimum/latest runtime contracts and both Tauri test jobs passed. I reran the Windows pruning job, which exited without an assertion or useful error in its log.

pascalandr:
I completed the gatekeeper loop: both initial P1 findings were fixed, and subsequent reviews found no remaining actionable code issues.

The Windows pruning rerun still exits without a diagnostic at UI startup. I ran the complete isolated native/UI fixture successfully on Windows with OpenCode 2.0.14, then repeated it with Node 24.20.0, the exact CI version. Both passed busy-state handling, individual/group/response/session cleanup, persistence, reload and plugin lifecycle checks. The CI-only exit remains undiagnosed; I am not treating CI as fully green.

I am admin-merging the reviewed c64a06dc head with these limitations recorded. The other red checks are the pre-existing search and migration failures documented above.

PR #751, opened 2026-09-22
fix(proxy): unblock message sends on session instruction sync

### LINKED ISSUE #750: [Bug]: Every message send fails with UnexpectedStatus — proxied DELETE …/instructions/entries/codenomad.voice-mode returns 500
## Issue Summary
Every message send is aborted with the dialog "Échec de l'envoi / Impossible d'envoyer le message / UnexpectedStatus". The prompt itself reaches OpenCode and the agent runs (full transcript server-side), but the send pipeline fails on a secondary call: `DELETE /workspaces/{id}/instance/api/experimental/session/{sessionID}/instructions/entries/codenomad.voice-mode` returns **500** through the CodeNomad proxy, while the same call directly against OpenCode returns **204**. The generated client only declares `[400, 401, 404]` for this operation, so the 500 surfaces as `ClientError("UnexpectedStatus")` and aborts the whole send.

## Steps to Reproduce
1. Open any workspace with any session (tried on old and new sessions).
2. Voice mode OFF (default).
3. Type any message and hit send.
4. Observe the error dialog "Échec de l'envoi … UnexpectedStatus"; DevTools Network shows `DELETE …/instructions/entries/codenomad.voice-mode → 500 (Internal Server Error)`.
5. Direct call bypassing the proxy succeeds:
   - `GET /api/experimental/session/{id}/instructions/entries` → `200 {"data":[]}`
   - `DELETE …/entries/codenomad.voice-mode` → `204` (existing or missing entry, both fine)

## Expected Behavior
The voice-mode instruction sync is a non-critical side call: its failure (especially a proxy/contract error like this) should not abort the user's message send. And the proxied DELETE should succeed (204) like the direct call.

## Logs & Screenshots
DevTools console (send flow):
```
DELETE http://127.0.0.1:9899/workspaces/02677658-…/instance/api/experimental/session/ses_f55f8d2f…/instructions/entries/codenomad.voice-mode 500 (Internal Server Error)
    at sdk-manager.ts:51
    at service-5mmhmmk7.js:42 / :56 / :492
    at session-instructions.ts:46
    at syncVoiceModeInstruction (session-instructions.ts:54)
    at session-actions.ts:338 → admitSessionAction → sendMessage (session-actions.ts:332)
```

## Additional Context
Root cause (verified by replaying CodeNomad's own runtime transport in Node):

- Frontend `syncVoiceModeInstruction` (`kj` in the UI bundle) calls `session.instructions.entry.remove({sessionID, key})` with **no location scope** whenever voice mode is off — i.e. on every send.
- The browser request carries the `x-codenomad-location` workspace header, which the proxy forwards into its runtime transport (`proxyWorkspaceRequest`, `dist/server/http-server.js` → `forwardRuntimeRequest`).
- The transport's `applyLocationContext` (`dist/opencode/compatibility/location.js:77-79`) requires a `location[directory]` query param for this route; it's absent, so it throws `Error("Location context does not match request")`.
- The throw propagates out of `forwardRuntimeRequest` → Fastify default handler → **500**. OpenCode never sees the request (no 500 in its log; direct calls are 204).
- Reproduced deterministically: same transport + same path **without** the header → forwarded fine; **with** the header → throws. Opencode 2.0.14 is innocent (full PUT/DELETE cycle verified 204).

Second bug in the same area (seen while capturing): `GET …/instructions/entries` is **missing** from the proxy allowlist (`isAllowedInstanceApiRoute`, `http-server.js:1060` — PUT/DELETE entries are listed at lines 1098-1099, no GET) → `403 {"error":"OpenCode route is not available through a workspace"}`.

Suggested fixes:
1. Proxy: scope/exempt the experimental instructions routes in `applyLocationContext` (the session was already ownership-checked via `session.get`), and add the missing `GET …/instructions/entries` to the allowlist.
2. Frontend (or proxy): send location scope with instructions entry calls.
3. `sendMessage`: don't hard-fail the whole send on voice-instruction sync errors.
4. UX: include the HTTP status (and URL) in the send-failure dialog instead of the bare `UnexpectedStatus` reason — it currently hides the real cause.

Environment: CodeNomad `0.20.0-dev-20260922-5650ed9d` (Tauri, Windows 11), OpenCode `2.0.14` (managed `serve --service` daemon and npm `opencode`/`opencode2` agree).

pascalandr:
Fixes #750.

I reproduced the failure against an isolated OpenCode 2.0.14 daemon: a proxied DELETE of the voice-mode instruction entry returns 500 when a location header is present, while the direct call returns 204. The runtime transport required a location query slot that session routes do not have, even though the proxy already authorizes them through session ownership. The generated instructions list route was also missing from the proxy allowlist (403).

Session routes (including single-identity reads/deletes, with normalized trailing slashes and the same active/import exclusions as the proxy) now pass through the transport without the location-slot assertion. The list route is allowlisted, voice-mode sync is best-effort so it can no longer abort a send, and bare UnexpectedStatus errors now include the HTTP status when no deeper detail exists.

### Validation

- Isolated OpenCode 2.0.14 fixture: proxied PUT/DELETE with location header return 204 (were 500), list returns 200 (was 403), native entries empty after proxied DELETE.
- 49 server tests (transport exemption incl. strict-route preservation, proxy allow/deny incl. foreign-session 403) and 53 UI tests (voice-failure continuation for prompt/command/shell, placement-failure propagation, error-message priority) pass.
- Server, UI and desktop-client typechecks pass.

### Review

Two independent gatekeeper rounds. Round 1 found two minors (trailing-slash dependence, status masking nested detail); I fixed both and extended coverage. Round 2 reported zero findings.


pascalandr:
CI triage for run https://github.com/NeuralNomadsAI/CodeNomad/actions/runs/35783167377: green are authorize, runtime contracts (minimum and latest), session-pruning native on all three OS including Windows, and both Tauri jobs. The remaining reds are pre-existing and unrelated to this change: the history-migration assertion fails on the credential metadata field added by OpenCode 2.0.14 (same failure on the earlier unrelated run https://github.com/NeuralNomadsAI/CodeNomad/actions/runs/35753233159), and the browser job passes 173 of 176 with only the two system-message search timeouts that I reproduced on the pre-PR base commit. The comment job fails only as a cascade of the skipped build.

I am admin-merging the reviewed 0d395ce3 head with these limitations recorded.

